PR #1, opened 2017-03-13
Changed strings used for the welcome msg

Zaab1t:


porcupine/editor.py
```diff
@@ -68,7 +68,7 @@ def create_welcome_msg(frame):
     innerframe.place(relx=0.5, rely=0.5, anchor='center')  # float in center
 
     titlelabel = tk.Label(innerframe, font='TkDefaultFont 16',
-                          text="Welcome to Porcupine!")
+                          text="<class 'Porcupine'>")
```
Akuli: Is this a copy/pasta or something intended?

Zaab1t: Intended. Just thought it was a fun touch.

PR #2, opened 2017-03-16
Added two new themes; HitGub and Zaab

Zaab1t:


Akuli:
Really nice! Thank You!

Akuli:
After working with the Zaab theme for a while I'm starting to love it even more. Absolutely awesome work!

PR #3, opened 2017-05-29
Happy birthday, dude!

Zaab1t:


Akuli:
Thanks, I didn't notice this earlier :)

This stuff looks good. The selection thing is nice, but I'm not sure about the Share thing. People guess what it does, but on the other hand, people talk about pastebinning code on the IRC channel. How about just "Pastebin"?

Zaab1t:
Share > Pastebin > Pastebin to. IMO. But it's your call.

Akuli:
Actually I like this too. I'll update the wiki :)

PR #6, opened 2017-06-07
Allow opening STDIN as a file again.

PurpleMyst:


Akuli:
Do we really need a temporary file for this? As far as I know there are no size limitations for what you can ipc with multiprocessing (just must be picklable), so you could just read stdin and send the content.

Akuli:
This is great! Thanks :)

PR #8, opened 2017-08-10
Make the autocompleter plugin more extendible.

PurpleMyst:
A Jedi-powered auto-completer for Python is already provided.
The rest of the info can be found in the commit messages.


Akuli:
Awesome! :D

PR #10, opened 2017-08-10
Modify find/replace behavior and add IRC client.

PurpleMyst:


Akuli:
I've thought about the find/replace thing a bit more, and I think it should work like this:

- The "find" button finds the next occurance of the search string, starting at the current cursor position. If there are no more occurences, searching is continued from the beginning of the file (let's call this *cycling* from now on), but a "Continued searching from beginning" message or similar is displayed in the status thing at right.
- The "replace" button is usually grayed out (`button["state"] = "disabled"`), but it's enabled (`button["state"] = "normal"`) when the find button has just been clicked. The replace button also cycles and displays a warning about it.
- The "replace and find" button is also grayed out when replace is grayed out, and clicking it is equivalent to first clicking replace and then clicking find.
- The "replace all" button does *not* cycle, otherwise porcupine crashes if the user replaces e.g. `x` with `x_` with the full words thing turned off. It simply replaces everything in the whole file, and for simplicity (to both implement and use) does not keep track of whether other buttons have been used to replace some occurances.

What do you think?

Akuli:
Can you close this pull request and make separate pull requests about the find dialog and IRC client? I would really like to have a less fucked up find dialog, but the IRC client is not really in a hurry like that at all.

PR #15, opened 2017-09-07
Add find.py changes

PurpleMyst:


Akuli:
Can you fix the ttk stuff? I want the find area to use ttk widgets.

Akuli:
The find area needs to be pretty much rewritten anyway, so let me close this old thing that is definitely not a complete rewrite...

PR #16, opened 2017-10-26
changed find.py to ttk

Zaab1t:
as you requested in #15. I don't think it's acting as intended though. I might look into it this weekend :)

### DIFF
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -1,6 +1,7 @@
 """Find/replace widget."""
 import re
 import tkinter as tk
+from tkinter import ttk
 import weakref
 
 import porcupine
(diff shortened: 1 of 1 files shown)


Akuli:
`__Myst__` has written a couple versions of the `find.py`, and they have been "working" but I haven't been very happy with them. Currently it has at least these issues:

- Pressing any key while the main area is focused moves the focus to the find area.
- Try find/replacing something in a long file and skipping some matches. After every replace, clicking the find button finds the first match in the whole file instead of next after the replace.
- Copy/pasting is very broken, but that's my fault. It's broken in all other entries as well because it's a global binding (and it shouldn't be).

Don't feel bad if you rewrite the whole thing. After all I'm really happy to have someone fixing the find area for me. It has been annoying for several months now, and I have used Porcupine as my main editor most of that time.

Akuli:
Are you still interested in working on this? The copy/pasting issue is fixed in new porcupines.

Akuli:
Alright, over a year of find area bugginess :D I remember using porcupine with the buggy find area at work last summer.

Zaab1t:
Can I close this issue? find needs a complete rework imo

Akuli:
yes

Akuli:
the complete rework is done now, i rewrote the whole thing

PR #225, opened 2021-01-30
fix windows copy/paste error

### LINKED ISSUE #215: Using ctrl+v causes Porcupine to generate error message
Using ctrl+v causes Porcupine to generate following error message:

  File "C:\Program Files\Python39\lib\tkinter\__init__.py", line 1884, in __call__
    return self.func(*args)
  File "C:\Users\Martin\AppData\Roaming\Python\Python39\site-packages\porcupine\textwidget.py", line 232, in _change_event_from_command
    args = [widget.index(arg) for arg in args_tuple]
  File "C:\Users\Martin\AppData\Roaming\Python\Python39\site-packages\porcupine\textwidget.py", line 232, in <listcomp>
    args = [widget.index(arg) for arg in args_tuple]
  File "C:\Program Files\Python39\lib\tkinter\__init__.py", line 3734, in index
    return str(self.tk.call(self._w, 'index', index))
_tkinter.TclError: text doesn't contain any characters tagged with "sel"

gxDNNa7h9q

Akuli:
fixes #215

### DIFF
--- a/porcupine/textwidget.py
+++ b/porcupine/textwidget.py
@@ -1,6 +1,7 @@
 import contextlib
 import dataclasses
 import functools
+import re
 import tkinter
 import weakref
 from typing import TYPE_CHECKING, Any, Callable, Iterator, List, Optional, Tuple, overload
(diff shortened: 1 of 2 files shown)


Akuli:
Unfortunately I can't figure out how to make a unit test to check if this works, but I tested it on my windows computer

PR #275, opened 2021-02-04
run tests on macos

### LINKED ISSUE #265: run tests on mac osx
same as #197 but for mac: ci tests should run on mac too

Akuli:
Fixes #265


Akuli:
Mac tests are still slower than windows tests because https://github.com/actions/setup-python/issues/58

Issue #84, opened 2020-08-11
use f-strings everywhere


Akuli:
currently porcupine contains some %s formatting, and all of my new code uses f-strings

Akuli:
i use f-strings because they are nice, not because there's something wrong with % formatting, so this doesn't actually solve a problem

i will keep using f-strings for new code, but replacing all % formatting with f-strings "just because" is just dumb imo

PR #289, opened 2021-02-07
Fstrings

Akuli:
I ended up doing #84 anyway, although I used some tools to help me with it.


Akuli:
Don't know why tests of 709d743 errored. Should be fixed now.

PR #317, opened 2021-02-10
handle macos smooth scrolling

### LINKED ISSUE #299: Scrolling through tabs too sensitive on mac (trackpad)
If you slide very slowly with two fingers, you can go to specific tabs, but slowing the sensitivity down would make it easier to control it

Akuli:
Fixes #299


Akuli:
Note to self: ask ThePhilgrim to try this and adjust sensitivity before merge (both #299 and font size changing with ctrl+wheel)

PR #324, opened 2021-02-10
run macos tests with the correct python

Akuli:
Both 3.8 and 3.9 seem to be installed by default. Make sure that the correct one is used.


Akuli:
Don't know why windows CI froze, likely will never happen again (or at least will be too difficult to reproduce)

All tests passed, so seems like the cleanup code froze

PR #360, opened 2021-02-16
Typing update

Akuli:
Update mypy and typeshed


Akuli:
The new mypy is ridiculously slow, even slower than running tests on Windows.

Akuli:
I run mypy 9 times to go through each platform+version combination.

Old mypy: 5sec*9 = 45sec
New mypy: 20sec*9 = 3min

https://github.com/python/typeshed/issues/4580#issuecomment-701499475

Akuli:
fixed with https://github.com/python/mypy/issues/10116

PR #371, opened 2021-02-21
Use gui for finding plugin directory

### LINKED ISSUE #359: --print-plugindir is difficult to run when porcupine is installed with exe installer
When installed this way, no porcupine stuff ends up in PATH

Akuli:
Fixes #359

### DIFF
--- a/docs/plugin-intro.rst
+++ b/docs/plugin-intro.rst
@@ -96,7 +96,7 @@ implemented as plugins. They come with Porcupine (see `Where are the plugins?`_
 above), but you can also :source:`read them on GitHub <porcupine/plugins>`.
 
 This tutorial is all about writing *more* plugins, but we'll start by running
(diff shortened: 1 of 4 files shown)


docs/plugin-intro.rst
```diff
@@ -96,17 +96,16 @@ implemented as plugins. They come with Porcupine (see `Where are the plugins?`_
 above), but you can also :source:`read them on GitHub <porcupine/plugins>`.
 
 This tutorial is all about writing *more* plugins, but we'll start by running
-Porcupine with *no* plugins at all just to get an idea of how much can be done
-with plugins. Open a command prompt or terminal and run this command on it::
+Porcupine with *no* plugins at all just to get an idea of how much can be done with plugins.
```
Akuli, line 99:
```suggestion
Porcupine with *no* plugins at all, just to get an idea of how much can be done with plugins.
```

PR #447, opened 2021-06-04
Urls

### LINKED ISSUE #443: url before dot
`Visit https://github.com.` gets highlighted as if `.` was part of URL, but it isn't. I don't remember the last time I saw an URL that ends with `.`.

Akuli:
Fixes #443

### DIFF
--- a/tests/test_urls_plugin.py
+++ b/tests/test_urls_plugin.py
@@ -22,11 +22,13 @@ def test_find_urls_basic():
          <URL>
         ("URL")bla
         "(URL)" :)
- Bla bla (URL, bla)
+ Bla bla  URL.
+ Bla bla  URL, foo and bar.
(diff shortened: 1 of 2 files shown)


tests/test_urls_plugin.py
```diff
@@ -24,20 +22,28 @@ def test_find_urls_basic():
          <URL>
         ("URL")bla
         "(URL)" :)
- Bla bla (URL, bla)
+ Bla bla  URL.
+ Bla bla  URL, foo and bar.
+ Bla bla (URL) bla.
+ Bla bla (URL).
+ Bla bla (URL.)
+ Bla bla (URL, bla).
```
Akuli, line 30: Perhaps this should no longer be named `test_find_urls_basic`.

PR #449, opened 2021-06-04
support slashes in filetypes conf

### LINKED ISSUE #440: mako-templates
I have a project with files like `mako-templates/hello.html`. I tried adding this to filetypes.toml:

```toml
['Mako HTML']
filename_patterns = ["mako-templates/*.html"]
pygments_lexer = 'pygments.lexers.MakoHtmlLexer'
```

This doesn't work. Even `["*.html"]` doesn't work, without this small modification to Porcupine:

```diff
diff --git a/porcupine/plugins/filetypes.py b/porcupine/plugins/filetypes.py
index 9de5277..9c1c5d8 100644
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -77,8 +77,8 @@ def get_filetype_from_matches(
     if not matches:
         return None

-    [result, *rest] = matches.values()
+    [*rest, result] = matches.values()
     if rest:
         names = ', '.join(matches.keys())
         log.warning(f"multiple file types match {they_match_what}: {names}")
     return result
```

Akuli:
Fixes #440 

- Allow using forward-slashes in `filename_patterns`
- When multiple matches are found, use the last matching pattern instead of the first one, so that the user's custom config file is used rather than defaults

### DIFF
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -10,7 +10,7 @@
 #    filename_patterns
 #        List of strings with * used as wildcard, such as "*.py" to match files
 #        named "something.py". You can also specify "foo/*.py" to only match
-#        Python files inside a folder named "foo", but you must use forward
+#        Python files inside a folder named foo, but you must use forward
 #        slashes for this, even on Windows.
 #
 #    shebang_regex
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -9,7 +9,9 @@
 #
 #    filename_patterns
 #        List of strings with * used as wildcard, such as "*.py" to match files
-#        named "something.py".
+#        named "something.py". You can also specify "foo/*.py" to only match
+#        Python files inside a folder named "foo", but you must use forward
+#        slashes for this, even on Windows.
 #
 #    shebang_regex
 #        Regular expression checked against the first line of a file when no
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -106,7 +106,7 @@ def guess_filetype_from_path(filepath: pathlib.Path) -> Optional[FileType]:
             fnmatch.fnmatch(filepath.as_posix(), '*/' + pat)
             for pat in filetype['filename_patterns']
         )
-    }, f"path {filepath}")
+    }, str(filepath))
 
 
 def guess_filetype_from_shebang(content_start: str) -> Optional[FileType]:
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -70,7 +70,7 @@ def load_filetypes() -> None:
         filetype.setdefault('langserver', None)
 
 
-def set_filedialog_kwargs():
+def set_filedialog_kwargs() -> None:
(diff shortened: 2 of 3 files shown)


tests/test_filetypes_plugin.py
```diff
@@ -42,3 +65,28 @@ def test_unknown_filetype(filetab, tmp_path):
     filetab.save()
     lexer_class_name = filetypes.get_filetype_for_tab(filetab)['pygments_lexer']
     assert lexer_class_name.endswith('.TextLexer')
+
+
+def test_slash_in_filename_patterns(custom_filetypes, caplog):
+    assert filetypes.guess_filetype_from_path(
+        pathlib.Path.home() / "foo" / "bar.html"
```
Akuli, line 72: Use `tmp_path` fixture

porcupine/default_filetypes.toml
```diff
@@ -9,7 +9,9 @@
 #
 #    filename_patterns
 #        List of strings with * used as wildcard, such as "*.py" to match files
-#        named "something.py".
+#        named "something.py". You can also specify "foo/*.py" to only match
+#        Python files inside a folder named "foo", but you must use forward
```
Akuli, line 13:
```suggestion
#        Python files inside a folder named foo, but you must use forward
```

porcupine/plugins/filetypes.py
```diff
@@ -70,29 +70,43 @@ def load_filetypes() -> None:
...
     return get_filetype_from_matches({
         name: filetype
         for name, filetype in filetypes.items()
         if any(
-            fnmatch.fnmatch(filepath.name, pat)
+            fnmatch.fnmatch(filepath.as_posix(), '*/' + pat)
             for pat in filetype['filename_patterns']
         )
-    }, f"filename {filepath.name!r}")
+    }, f"path {filepath}")
```
Akuli, line 109: It is an absolute path, so you will know it is a path:

```suggestion
    }, str(filepath))
```

tests/test_filetypes_plugin.py
```diff
@@ -42,3 +65,28 @@ def test_unknown_filetype(filetab, tmp_path):
...
+        tmp_path / "foobar-mako-templates" / "bar.html"
+    )['pygments_lexer'] == 'pygments.lexers.HtmlLexer'
+
+    with caplog.at_level(logging.WARNING):
+        assert filetypes.guess_filetype_from_path(
+            tmp_path / "mako-templates" / "bar.html"
+        )['pygments_lexer'] == 'pygments.lexers.MakoHtmlLexer'
+
+    assert len(caplog.records) == 1
+    assert "2 file types match path" in caplog.records[0].message
```
Akuli, line 85:
```suggestion
    assert "2 file types match" in caplog.records[0].message
```

tests/test_filetypes_plugin.py
```diff
@@ -42,3 +65,28 @@ def test_unknown_filetype(filetab, tmp_path):
...
+    assert str(tmp_path) in caplog.records[0].message
+    assert "HTML, Mako template" in caplog.records[0].message
+
+    # Filedialog doesn't support slashes in patterns
```
Akuli, line 89:
```suggestion
    # filedialog doesn't support slashes in patterns
```

PR #455, opened 2021-06-05
try to make tests less fragile

### LINKED ISSUE #379: sometimes tests freeze weirdly
hard to reproduce, has happened on my system and in ci, ctrl+c traceback says something about threads (don't have the traceback anymore)

Akuli:
I think will fix #379, will reopen it if not

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,7 @@ def porcusession(monkeypatch_dirs):
 # utils.run_in_thread() can make tests fragile
 @pytest.fixture
 def dont_run_in_thread(monkeypatch):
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -97,6 +98,19 @@ def porcusession(monkeypatch_dirs):
     porcupine.quit()
 
 
+# utils.run_in_thread() can make tests fragile
+@pytest.fixture
+def dont_run_in_thread(monkeypatch):
+    def func(blocking_function, done_callback, check_interval_ms=69):
```
Akuli, line 104:
```suggestion
    def func(blocking_function, done_callback, check_interval_ms=69, daemon=True):
```

PR #471, opened 2021-06-06
Highlight previous search in find entry when re opening it

### LINKED ISSUE #465: ctrl+f select all
When nothing is selected in main editing widget and I press ctrl+f, find entry should have everything selected so that it's easy to type new thing to find

ThePhilgrim:
Fixes #465

### DIFF
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -145,6 +145,12 @@ class Finder(ttk.Frame):
         try:
             selected_text: Optional[str] = self._textwidget.get("sel.first", "sel.last")
         except tkinter.TclError:
+            selected_text = None
+
+        self.pack(fill="x")
+
+        if selected_text is None or "\n" in selected_text:
+            self.find_entry.focus_set()
             self.find_entry.selection_range(0, "end")  # when ctrl + f without text selected
         else:
             self.find_entry.delete(0, "end")
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -151,7 +151,8 @@ class Finder(ttk.Frame):
 
         if selected_text is None or "\n" in selected_text:
             self.find_entry.focus_set()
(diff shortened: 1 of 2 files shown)


porcupine/plugins/find.py
```diff
@@ -141,6 +141,7 @@ def show(self) -> None:
         try:
             selected_text = self._textwidget.get('sel.first', 'sel.last')
         except tkinter.TclError:
+            self.find_entry.selection_range(0, 'end')
             pass   # nothing is selected, leave find entry content as is
```
Akuli, line 145: A few things:
- [ ] `pass` can be deleted. It does nothing, and when there was nothing in the `except:` part, it was needed to avoid syntax error.
- [ ] Even though we are deleting the `pass`, I would like to have a comment to mention nothing is selected. It's not obvious by looking at the code.
- [ ] This PR is based on an old version of main branch, because it has single quotes. Latest main has black and double quotes everywhere. Please merge the latest main into it.

porcupine/plugins/find.py
```diff
@@ -145,12 +145,7 @@ def show(self) -> None:
         try:
             selected_text: Optional[str] = self._textwidget.get("sel.first", "sel.last")
         except tkinter.TclError:
-            selected_text = None
-
-        self.pack(fill="x")
-
-        if selected_text is None or "\n" in selected_text:
-            self.find_entry.focus_set()
```
Akuli, line 153: I would rather not delete all this :) Can you add it back?

Akuli (changes requested):

porcupine/plugins/find.py
ThePhilgrim: WTF?

ThePhilgrim: .... git diff --staged .... I never learn .

Akuli (approved):
This is awesome! Thanks!!!

PR #476, opened 2021-06-06
Undo ctrl + f through change_batch

### LINKED ISSUE #431: undoing ctrl+f
doesn't seem to use `edit_separator()` or something similar

ThePhilgrim:
Fixes #431

### DIFF
--- a/tests/test_find_plugin.py
+++ b/tests/test_find_plugin.py
@@ -385,43 +385,13 @@ def test_highlight_text(filetab_and_finder):
     assert finder.find_entry.selection_present()
 
 
-def test_find_undo(filetab_and_finder):
+def test_undo_replace_all(filetab_and_finder):
     filetab, finder = filetab_and_finder
     filetab.textwidget.insert("end", "foo bar foo")
-    filetab.textwidget.mark_set("insert", "1.0")
-    filetab.textwidget.tag_add("sel", "1.0", "1.3")  # Select foo so it goes to find entry
     finder.find_entry.insert("end", "foo")
-    finder.show()
 
     finder.replace_entry.insert("end", "baz")
-    finder.replace_this_button.invoke()
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "baz bar foo"
-    filetab.textwidget.edit_undo()
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo bar foo"
-
-    finder.find_entry.delete(0, "end")  # Delete & insert because of issue #477
-    finder.find_entry.insert("end", "foo")
-
     finder.replace_all_button.invoke()
     assert filetab.textwidget.get("1.0", "end - 1 char") == "baz bar baz"
     filetab.textwidget.edit_undo()
     assert filetab.textwidget.get("1.0", "end - 1 char") == "foo bar foo"
-
-
-@pytest.mark.xfail
-def test_replace_undo_replace(filetab_and_finder):
-    filetab, finder = filetab_and_finder
-    filetab.textwidget.insert("end", "foo")
-    filetab.textwidget.mark_set("insert", "1.0")
-    filetab.textwidget.tag_add("sel", "1.0", "1.3")  # Select foo so it goes to find entry
-    finder.find_entry.insert("end", "foo")
-    finder.show()
-
-    finder.replace_entry.insert("end", "bar")
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo"
-    finder.replace_this_button.invoke()
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "bar"
-    filetab.textwidget.edit_undo()
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo"
-    finder.replace_this_button.invoke()
-    assert filetab.textwidget.get("1.0", "end - 1 char") == "bar"
--- a/tests/test_find_plugin.py
+++ b/tests/test_find_plugin.py
@@ -408,3 +408,16 @@ def test_replace_this_match(filetab_and_finder):
     assert filetab.textwidget.get("1.0", "end - 1 char") == "foo lol baz"
     filetab.textwidget.edit_undo()
     assert filetab.textwidget.get("1.0", "end - 1 char") == "foo bar baz"
+
+
+@pytest.mark.xfail
+def replace_this_greyed_out(filetab_and_finder):
+    filetab, finder = filetab_and_finder
+    filetab.textwidget.insert("end", "foo bar foo")
+    filetab.textwidget.mark_set("insert", "1.0")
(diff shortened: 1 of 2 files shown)


tests/test_find_plugin.py
```diff
@@ -383,3 +383,18 @@ def test_highlight_text(filetab_and_finder):
...
     assert finder.find_entry.selection_present()
+
+
+def test_find_undo(filetab_and_finder):
+    filetab, finder = filetab_and_finder
+    finder.hide()
+    filetab.textwidget.insert("end", "foo bar foo")
+    filetab.textwidget.mark_set("insert", "1.0")
+    filetab.textwidget.tag_add("sel", "1.0", "1.3")
+    finder.show()
```
Akuli, line 394: Nit: There's no need to hide it. You can just leave it shown for the whole time.

tests/test_find_plugin.py
```diff
@@ -383,3 +383,18 @@ def test_highlight_text(filetab_and_finder):
...
+
+    finder.replace_entry.insert("end", "baz")
+    finder.replace_all_button.invoke()
```
Akuli, line 397: Can you also add a test for "Replace this" button?

Akuli (commented):
Looks good overall! A couple small nits.

tests/test_find_plugin.py
```diff
@@ -383,3 +383,18 @@ def test_highlight_text(filetab_and_finder):
...
     assert finder.find_entry.get() == "foo"
     assert finder.find_entry.selection_present()
+
+
+def test_find_undo(filetab_and_finder):
+    filetab, finder = filetab_and_finder
+    finder.hide()
+    filetab.textwidget.insert("end", "foo bar foo")
+    filetab.textwidget.mark_set("insert", "1.0")
+    filetab.textwidget.tag_add("sel", "1.0", "1.3")
```
Akuli, line 393:
```suggestion
    filetab.textwidget.tag_add("sel", "1.0", "1.3")  # Select foo so it goes to find entry
```

Akuli (changes requested):

tests/test_find_plugin.py
Akuli: ...or add to end of this test, whatever feels better :)

ThePhilgrim:
**Found bug that needs addressing!**

If you find text so it's highlighted, and you replace it, then ctrl + f, the text is no longer highlighted. This can become problematic if you want to replace something, then ctrl + f cause you realize you want to replace all. Then all wont be replaced.

I spend time trying to fix it, but I can't figure out how I would do it. My thought was to add a trace or Event to tkinter's edit_undo(). If the event runs, and find window is open -> run highlight_text(). I can't find how to trace the undo event.. Seems impossible.


https://imgur.com/bk01kf2

Akuli:
I think this problem only happens when you undo a "Replace this" operation, as there is no other way to get a match that isn't highlighted. And if you undo "Replace this", it means you didn't want that match to get replaced, so it should behave just like clicking "Next match". The difference is that the "Next match" button leaves it highlighted, so that "Replace all" will find it.

I can see a few options:
- Trace the undo event. When undoing a "Replace this" operation, it would re-highlight the match. I can see one way to trace an undo event, but I think it would be difficult to figure out whether it was a "Replace this" undo or some other undo. It's important to get that right, because we really don't want to touch other unrelated undos. I see this as a limitation in Tk's undo mechanism, and we can revisit this if I ever make custom undo logic.
- Ignore highlights in "Replace all". This becomes a problem if you first find something, then add more matches by editing the text, and then "Replace all", as it will also replace the new matches **without ever highlighting them**. That's surprising, and I don't want that.
- Do nothing. This seems fine to me.

I would prefer this to be fixed in a separate PR. You can still leave the failing test in and mark it with [pytest.mark.xfail](https://docs.pytest.org/en/reorganize-docs/new-docs/user/xfail.html). I would prefer two separate tests, one succeeding test for the functionality that this PR adds (that you had earlier) and another xfail test for the new bug.

ThePhilgrim:
Actually, I can see this happening though. Not often. But I sometimes replace, change my mind, and then change my mind again (redo the undo)

If it possible to trace the undo event (I found no way to do it), I would on the undo event, check if finder window is open, and if so, run the highlight text method. That means that:

1) if finder is not open, nothing happens
2) if finder is open, but no text is in find entry, nothing will be highlighted,
3) if finder is open and find entry contains something, it will highlight that text again. Which is the expected behaviour.

So if there is a way to get a trace undo, it should be a simple fix.

ThePhilgrim:
> I would prefer this to be fixed in a separate PR. You can still leave the failing test in and mark it with [pytest.mark.xfail](https://docs.pytest.org/en/reorganize-docs/new-docs/user/xfail.html). I would prefer two separate tests, one succeeding test for the functionality that this PR adds (that you had earlier) and another xfail test for the new bug.

Okay, will split them up tonight after work

Akuli:
There is an event named `<<Undo>>` which runs on Ctrl+Z press, but not when the `.edit_undo()` method is called. It seems good enough for me.

To use it in the find plugin, you would write:

```
def __init__(self, whatever):
    ...
    self._textwidget.bind("<<Undo>>", self._handle_undo, add=True)
    ...

def _handle_undo(self, event: object) -> None:
    ...
```

As you would expect, `event` is something that tkinter passes the the callback function, but we don't need in this case.

You will need to change the test to use `filetab.textwidget.event_generate("<<Undo>>")` instead of `filetab.textwidget.edit_undo()` to better simulate the user pressing Ctrl+Z, as the `<<Undo>>` event doesn't catch `.edit_undo()`. If the undo event doesn't seem to work in the test, you can add `filetab.update()` before it.

Alternatively, we can add more code under the big warning triangle comment in porcupine/textwidget.py. That magic is powerful enough to catch all undos, including those created with the `.edit_undo()` method.

Akuli:
Turns out there's one more gotcha with the `<<Undo>>` event: it runs before the undoing actually happens, and if the undo changes `bar` to `foo`, the bind callback will see `bar`. To work around this, you can use `after_idle`, which runs a function later, when the event loop is done with processing the `<<Undo>>`:

```python
def _handle_undo(self, event: object) -> None:
    self.after_idle(self.highlight_all_matches)
```

tests/test_find_plugin.py
```diff
@@ -383,3 +383,41 @@ def test_highlight_text(filetab_and_finder):
...
+
+    finder.next_button.invoke()
+    finder.replace_this_button.invoke()
+    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo lol baz"
+    filetab.textwidget.edit_undo()
+    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo bar baz"
+
+
+@pytest.mark.xfail
+def replace_this_greyed_out(filetab_and_finder):
```
Akuli, line 414:
```suggestion
def test_replace_this_greyed_out(filetab_and_finder):
```

Akuli (approved):

PR #481, opened 2021-06-07
Fix highlight text when undo replace

### LINKED ISSUE #477: Undoing find/replace will not highlight the text in find_entry again.
If you find text so it's highlighted, and you replace it, then ctrl + f, the text is no longer highlighted. This can become problematic if you want to replace something, then ctrl + f cause you realize you want to replace all. Then all wont be replaced.

I spend time trying to fix it, but I can't figure out how I would do it. My thought was to add a trace or Event to tkinter's edit_undo(). If the event runs, and find window is open -> run highlight_text(). I can't find how to trace the undo event.. Seems impossible.

https://imgur.com/bk01kf2

Trying to solve in #476

ThePhilgrim:
Fixes #477

### DIFF
--- a/tests/test_find_plugin.py
+++ b/tests/test_find_plugin.py
@@ -385,6 +385,22 @@ def test_highlight_text(filetab_and_finder):
     assert finder.find_entry.selection_present()
 
 
+def test_highight_on_undo(filetab_and_finder):
+    filetab, finder = filetab_and_finder
+    filetab.textwidget.insert("end", "foo")
+    finder.find_entry.insert("end", "foo")
+
+    finder.replace_entry.insert("end", "baz")
+    finder.replace_all_button.invoke()
+    assert filetab.textwidget.get("1.0", "end - 1 char") == "baz"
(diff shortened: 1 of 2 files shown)


tests/test_find_plugin.py
```diff
@@ -385,6 +385,22 @@ def test_highlight_text(filetab_and_finder):
...
+    finder.find_entry.insert("end", "foo")
+
+    finder.replace_entry.insert("end", "baz")
+    finder.replace_all_button.invoke()
+    assert filetab.textwidget.get("1.0", "end - 1 char") == "baz"
+    filetab.update()
+    filetab.textwidget.event_generate("<<Undo>>")
+    assert filetab.textwidget.get("1.0", "end - 1 char") == "foo"
+    filetab.update()
+    print(filetab.textwidget.tag_ranges("find_highlight"))
```
Akuli, line 400:
```suggestion
```

Akuli:
Thanks!

PR #483, opened 2021-06-07
Fix rendering issue when opening ctrl f on mac

ThePhilgrim:
BEFORE:
https://imgur.com/p1l5snQ

AFTER:
https://imgur.com/4zdz6ww

### DIFF
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -163,6 +163,7 @@ class Finder(ttk.Frame):
             self.find_entry.focus_set()
 
         self.highlight_all_matches()
+        self.update_idletasks()
 
     def hide(self, junk: object = None) -> None:
         self._textwidget.tag_remove("find_highlight", "1.0", "end")
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -163,7 +163,10 @@ class Finder(ttk.Frame):
             self.find_entry.focus_set()
 
         self.highlight_all_matches()
-        self.update_idletasks()
+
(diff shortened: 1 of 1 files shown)


ThePhilgrim:
It seems to be a mac-specific issue (https://stackoverflow.com/questions/55366795/does-anyone-know-why-my-tkinter-buttons-arent-rendering#comment97457465_55366795)

But I will look into it in accordance with your instructions.

ThePhilgrim:
The issue seems to be introduced in this commit

https://github.com/Akuli/porcupine/commit/485ae79eb7ea9a3573e8ce838459ec3a518bf128

Akuli:
That commit adds a new plugin that was then called overview, and is now called minimap. What happens if you disable the minimap plugin in the plugin manager on the latest porcupine?

ThePhilgrim:
The problem seems to be in the plugins _minimap_ and _highlight_ . Disabling either of them still leaves the problem there. If you disable both, the issue disappears.

porcupine/plugins/find.py
```diff
@@ -163,6 +163,7 @@ def show(self) -> None:
             self.find_entry.focus_set()
 
         self.highlight_all_matches()
+        self.update_idletasks()
```
Akuli, line 166:
```suggestion

        # weird hack to prevent rendering issue on mac
        # https://stackoverflow.com/questions/55366795/does-anyone-know-why-my-tkinter-buttons-arent-rendering
        self.update_idletasks()  
```

Akuli (changes requested):

Akuli (approved):

Issue #486, opened 2021-06-08
requests in requirements-dev.txt


Akuli:
should be in requirements.txt

danger-ahead:
I can do this. Please assign.

PR #487, opened 2021-06-08
updated requirements.txt

danger-ahead:
fixed issue #486


Akuli:
Thanks for contributing! I don't know who you are, but I love contributors  in general :)

Can you also delete it from requirements-dev.txt? It's fine to assume that if someone installs requirements-dev.txt, then they also install requirements.txt.

danger-ahead:
@Akuli done. Have a look.

PR #490, opened 2021-06-09
Create anchors plugin

ThePhilgrim:


porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,58 @@
...
+from porcupine.plugins import linenumbers
+
+# Dependent on code from linenumbers.py
+setup_after = ["linenumbers"]
+
+
+class AnchorManager:
+    def __init__(self, tab_textwidget: tkinter.Text, linenumber_instance: tkinter.Canvas) -> None:
+        self.tab_textwidget = tab_textwidget
+        self.linenumber_instance = linenumber_instance
```
Akuli, line 13: I would just name it `linenumbers`. Everything is an instance in Python, so the extra `_instance` adds no information.

If you want to avoid conflicting it with the global `linenumbers` variable introduced in `from porcupine.plugins import linenumbers`, you can change the import to `from porcupine.plugins.linenumbers import LineNumbers`. This creates a global variable named `LineNumbers`, which won't conflict with any lowercase variable names.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.anchor_symbol = "¶"
+
+        linenumber_instance.bind("<<Updated>>", self.update_linenumbers, add=True)
+
+    def toggle_on_off(self, event):
+        [cursor_index] = [self.tab_textwidget.index("insert linestart")]
```
Akuli, line 20: While `[foo] = [bar]` sets `foo` to `bar`, it is a rather complicated way to do so: it first creates a list containing `bar`, and then immediately unpacks the list, making `foo` be the only element.

```suggestion
        cursor_index = self.tab_textwidget.index("insert linestart")
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,58 @@
...
+        for anchor in self.tab_textwidget.mark_names():
+            if anchor.startswith("anchor_") and self.tab_textwidget.index(anchor) == cursor_index:
```
Akuli, line 22: Anchor marks can be in the middle of the line. Suppose the text widget contains two lines, and there's an anchor mark in the middle of the second line. I'll denote an anchor mark with `|` even though it's actually invisible.

```
hello world
|blah blah
```

If I now go to end of first line and press delete, or beginning of second line and press backspace, it turns into this:

```
hello world|blah blah
```

There are a couple ways to handle this. Choose your favorite:
1. Before you do anything with anchor marks, loop through them all and move them to the beginning of the line if they aren't already. Something like `textwidget.mark_set(mark, f"{mark} linestart")`.
2. Change this check so that it checks whether the mark is on the same line. Something like this:
    ```
    if textwidget.compare("insert linestart", "<=", anchor) and textwidget.compare(anchor, "<=", "insert lineend"):
    ```
    Or like this:
    ```
    if textwidget.index("insert").split(".")[0] == textwidget.index(anchor).split(".")[0]:
    ```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def update_linenumbers(self, event):
+        """Re-draws the anchor points every time the linenumber instance updates (scroll, insertion/deletion of text)"""
+        anchor_list = [
+            anchor for anchor in self.tab_textwidget.mark_names() if anchor.startswith("anchor_")
+        ]
+
+        for anchorpoint in anchor_list:
+            row_tag = "line_" + self.tab_textwidget.index(anchorpoint).split(".")[0]
+            [row_id] = self.linenumber_instance.find_withtag(row_tag)
```
Akuli, line 38: In a long file, the linenumbers plugin doesn't draw invisible lines. If you scroll so that an anchor mark is not visible, `find_withtag` will return an empty tuple and this code will error. Change it to handle `find_withtag` returning an empty tuple.

Akuli (commented):
As the title of the PR indicates, the code isn't done, but I was able to point out a few things anyway :)

Akuli:
This PR conflicted with my master, so I merged my master into your branch. (By default, I can modify your pull requests to my repos.) You will likely need to `git pull` on `anchor_plugin` branch before you can push.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+    def toggle_on_off(self, event: tkinter.Event[tkinter.Misc]) -> None:
+
+        self.prevent_duplicate_anchors()
+
+        cursor_index = self._get_cursor_index()
+        for anchor in self.tab_textwidget.mark_names():
+            if (
+                anchor.startswith(self.custom_anchor_prefix)
+                and self.tab_textwidget.index(anchor) == cursor_index
+            ):
```
Akuli, line 45: Nit: You can simplify this slightly by using `_get_anchors()`.

```suggestion
        for anchor in self._get_anchors():
            if self.tab_textwidget.index(anchor) == cursor_index:
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+                ]
+            )
+        )
+
+        next_anchor_row = 0
+        for anchor_row in anchor_rows:
+            if anchor_row > int(cursor_row):
+                next_anchor_row = anchor_row
+            else:
+                break
```
Akuli, line 72: This logic is a bit difficult to follow. How about the following:

```suggestion
        rows_after_cursor = [n for n in anchor_rows if n > int(cursor_row)]
        if rows_after_cursor:
            next_anchor_row = min(rows_after_cursor)
        else:
            next_anchor_row = 0
```

This leads to other simplifications:
- Because `min()` does not care about the order of the list given to it, you no longer need `sorted()` or `reversed()`.
- You can `return` instead of `next_anchor_row = 0`.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+        if not next_anchor_row:
+            return
+        else:
+            self.tab_textwidget.mark_set("insert", f"{str(next_anchor_row)}.0")
+
+        # If cursor is below last row
+        if int(self._get_cursor_index().split(".")[0]) > int(
+            self.tab_textwidget.index(f"@0,{self.tab_textwidget.winfo_height()}").split(".")[0]
+        ):
+            self.tab_textwidget.see("insert")
```
Akuli, line 83: It's fine to call `.see("insert")` unconditionally. It will do nothing if the cursor is already visible.

```suggestion
        self.tab_textwidget.see("insert")
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # text.see("insert")
+
+        # first_line = int(self._textwidget.index("@0,0").split(".")[0])
+        # last_line = int(
+        #     self._textwidget.index(f"@0,{self._textwidget.winfo_height()}").split(".")[0]
+        # )
+
+        # TODO: If user jumps to the last anchor and then to next, it should jump to the first one in the file.
```
Akuli, line 85: This is obviously a matter of opinion, but I would prefer it to do nothing. If I press the key binding to jump forward, I really wouldn't expect to ever jump backwards to the beginning.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+        ):
+            self.tab_textwidget.see("insert")
+
+        # text.see("insert")
+
+        # first_line = int(self._textwidget.index("@0,0").split(".")[0])
+        # last_line = int(
+        #     self._textwidget.index(f"@0,{self._textwidget.winfo_height()}").split(".")[0]
+        # )
+
```
Akuli:
```suggestion
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+        # If cursor is above first row
+        if int(self._get_cursor_index().split(".")[0]) < int(
+            self.tab_textwidget.index("@0,0").split(".")[0]
+        ):
+            self.tab_textwidget.see("insert")
+
+        # TODO: Current Bugs:
+        # 1) Jump to previous will delete code on current row if no more anchors above.
+        # 2) Can't commit/push because of git merge weirdness
```
Akuli: Looks like `2)` got resolved. Is `1)` still a problem? I don't see any code that should change the contents of the text widget.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+        anchor_list = self._get_anchors()
+
+        for anchorpoint in anchor_list:
+            row_tag = "line_" + self.tab_textwidget.index(anchorpoint).split(".")[0]
+            try:
+                [row_id] = self.linenumbers.find_withtag(row_tag)
+                row_text = self.linenumbers.itemcget(row_id, "text")
+                self.linenumbers.itemconfigure(row_id, text=row_text + " " + self.anchor_symbol)
+            except ValueError:
+                pass
```
Akuli, line 140: Look up `try,except,else` and consider using it here. If `itemcget()` or `itemconfigure()` raises a `ValueError` for whatever reason, you don't want it to get silently ignored.

I would also appreciate a comment like `# line not visible`.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,173 @@
...
+
+def on_new_tab(tab: tabs.Tab) -> None:
+    if isinstance(tab, tabs.FileTab):
+        [linenumbers] = [
+            child for child in tab.left_frame.winfo_children() if isinstance(child, LineNumbers)
+        ]
+        anchor = AnchorManager(tab.textwidget, linenumbers)
+        tab.textwidget.bind("<Control-g>", anchor.toggle_on_off, add=True)
+        tab.textwidget.bind("<Control-k>", anchor.jump_to_previous, add=True)
+        tab.textwidget.bind("<Control-l>", anchor.jump_to_next, add=True)
```
Akuli, line 163: We should make these configurable in `default_keybindings.tcl`. See other plugins for examples.

Akuli (commented):
Looks good! You pushed while I was reviewing, so I guess some of this is already fixed :)

ThePhilgrim:
Need help with mypy in this file.

porcupine/plugins/anchors.py
ThePhilgrim: Ah this makes sense! I need to learn to use min / max more!

Akuli:
Mypy fixd

porcupine/plugins/anchors.py
ThePhilgrim: But I was thinking that if two anchors are visible and you jump, .see will make the "camera" jump.. Whereas I'd like it to not move if both anchors are visible.

porcupine/plugins/anchors.py
ThePhilgrim: If you ctrl + f, and you "find next" all the way to the end of the file, and then "find next" again.. doesn't it jump back up?

porcupine/plugins/anchors.py
ThePhilgrim: These are still subject for change. Will add todo comment

porcupine/plugins/anchors.py
Akuli: It does, but to me it's fundamentally different from anchors. With Ctrl+F, you typically want to walk through the whole file, making sure you didn't miss anything. With anchors, you just want to jump to a different place.

But it makes sense if you are using anchors as basically a TODO list, for example: you mark many places, and then go through every anchor until there's no more anchors left. Is this something you want to do? Maybe we could add a checkbox to the settings dialog, and then argue about whether it should be checked by default?

porcupine/plugins/anchors.py
Akuli: It shouldn't jump. From manual page:

> If index is already visible then              the command does nothing. If index is a short  distance  out  of              view,  the  command  adjusts  the view just enough to make index              visible at the edge of the window.  If index is far out of view,              then the command centers index in the window.

porcupine/plugins/anchors.py
ThePhilgrim: Haha, that sounds good.

porcupine/plugins/anchors.py
ThePhilgrim: Comment 2 has also been deleted.

Yes, comment 1 is a bug. If you jump to the first anchor with "jump_to_previous", and then continue clicking ctrl+k, code from that row is deleted.

porcupine/plugins/anchors.py
Akuli: https://akuli.github.io/porcupine/settings.html#the-settings-dialog

There's `add_entry`, `add_combobox`, `add_spinbox` and `add_label`, but no `add_checkbutton`. I'll create an `add_checkbutton` method for you.

ThePhilgrim:
mypy --strict still throwing errors . Not sure if we care about these?

Akuli:
It passes on GitHub actions. Maybe you have an old mypy? `pip install -r requirements-dev.txt`

ThePhilgrim:
> It passes on GitHub actions. Maybe you have an old mypy? `pip install -r requirements-dev.txt`

https://dpaste.com/7VHVU7FKK

Akuli:
Adder has the same mypy problem. I don't have it. Maybe deleting `.mypy_cache` in porcupine folder helps? If not, I don't know what to try next.

ThePhilgrim:
Did not help. Strange.

Akuli:
Now there's an `add_checkbutton()` method in `porcupine/settings.py` on my master branch. There's a usage example in its docstring.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,164 @@
...
+            self.custom_anchor_prefix + str(time.time()), "insert linestart"
+        )
+
+        self.linenumbers.do_update()
+
+    def jump_to_next(self, event: tkinter.Event[tkinter.Misc]) -> None:
+        cursor_row = self._get_cursor_index().split(".")[0]
+        anchor_list = self._get_anchors()
+        anchor_rows = reversed(
+            sorted(
```
Akuli, line 57: Reversed and sorted are no longer needed.

porcupine/plugins/anchors.py
Akuli: Somehow the suggestion didn't click applied. Maybe you accidentally clicked "Resolve conversation" instead of "Commit suggestion"?

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,164 @@
...
+        rows_after_cursor = [n for n in anchor_rows if n > int(cursor_row)]
+        if rows_after_cursor:
+            next_anchor_row = min(rows_after_cursor)
+        else:
+            next_anchor_row = 0
+
+        if not next_anchor_row:
+            return
+        else:
+            self.tab_textwidget.mark_set("insert", f"{str(next_anchor_row)}.0")
```
Akuli, line 74:
```suggestion
        if not rows_after_cursor:
            return

        next_anchor_row = min(rows_after_cursor)
        self.tab_textwidget.mark_set("insert", f"{str(next_anchor_row)}.0")
```

porcupine/plugins/anchors.py
Akuli: This is the default behavior for pressing Ctrl+K. It happens even without the plugin. To fix it, change your methods to `return "break"`. This tells Tk that you have handled the event and nothing else should do anything about it.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,150 @@
...
+def on_new_tab(tab: tabs.Tab) -> None:
+    if isinstance(tab, tabs.FileTab):
+        [linenumbers] = [
+            child for child in tab.left_frame.winfo_children() if isinstance(child, LineNumbers)
+        ]
+        anchor = AnchorManager(tab.textwidget, linenumbers)
+        tab.textwidget.bind("<Control-g>", anchor.toggle_on_off, add=True)
+        tab.textwidget.bind("<Control-k>", anchor.jump_to_previous, add=True)
+        tab.textwidget.bind("<Control-l>", anchor.jump_to_next, add=True)
+        # TODO: When keybindings decided, add to default_keybindings.tcl
```
Akuli, line 146: Deciding key bindings seems to be more difficult than I thought. Ideally "jump to previous" would involve the arrow up somehow (and similarly for "jump to next" and arrow down). But:
- Up arrow by itself moves cursor up.
- Alt+Up scrolls while keeping the cursor in the center of the view
- Ctrl+Up jumps to the previous whitespace-separated section.
- Ctrl+Alt+Up doesn't do anything, but on my system, it's not possible to bind to it because Ctrl+Alt+(any arrow key) has a special meaning in general. For example, Ctrl+Alt+Right switches to the next workspace.

We get more possible key combinations by holding down shift, but it doesn't really help because it should move the cursor in the same way as without shift, but selecting as it goes. As a weird exception, Alt+Shift+Up is same as Alt+Up.

I see a couple options:
- Delete the Alt+Up binding, and use it for this instead. When I added it, I thought i would use it all the time, but I didn't end up using it very much.
- Use Alt+Shift+Up. This has the weird side effect that holding down Shift doesn't do what you would expect, but I don't think highlighting text while scrolling with cursor fixed in middle of screen would be very useful anyway.

Akuli: Forgot to mention, in case it isn't obvious yet: on Mac, you need to imagine it says "Command" whenever I wrote "Ctrl". In default_keybindings.tcl, we can use the `$contmand` variable.

ThePhilgrim: I just changed the commands to ctrl+e / r for jumping to prevent the deletion of code. However, these keybindings are only meant as temporary for developing the plugin. No thought have been put into these key bindings yet.

I don't know how you reposition your cursor per default (keyboard / mouse), for me it differs. But I think in a situation where I'm on line 5, and then jump to line 500 with an anchor, I might use the mouse to reposition the cursor to where I actually need it (not on linestart). Because of this I would keep my right hand on the mouse when using the anchors, and therefore a keybinding that could be used with one hand would make sense.

Is this a bad way to think about this?

Akuli: I wouldn't use the mouse. I would hold down control and press arrow keys :)

I think what matters the most is whether the key bindings are easy to learn. Even though "E=previous R=next" is convenient if you use a mouse, it's a bit of a gotcha compared to something that is associated with the "Up" and "Down" buttons in the obvious way. Once we use  `default_keybindings.tcl`, you can configure it however you want if you don't like the default.

Akuli:
Todo:
- [ ] Clean up `jump_to_previous()` so that the code is similar to `jump_to_next()`
- [ ] Create the checkbox that decides what to do once you reach the end/beginning of file
- [ ] Figure out what `bind_specific()` and `jump_to_specific()` are supposed to be?
- [ ] Figure out whether `self.tab_textwidget.see("insert")` can be made unconditional

ThePhilgrim:
Checkbox 1 & 4 done

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,133 @@
+from __future__ import annotations
```
Akuli, line 1: Please add a docstring to the beginning of the file. It will show up in plugin manager.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,133 @@
...
+        cursor_row = self._get_cursor_index().split(".")[0]
+        anchor_list = self._get_anchors()
+        anchor_rows = [
+            int(self.tab_textwidget.index(anchorpoint).split(".")[0]) for anchorpoint in anchor_list
+        ]
+
+        rows_after_cursor = [n for n in anchor_rows if n > int(cursor_row)]
+        if rows_after_cursor:
+            next_anchor_row = min(rows_after_cursor)
+            self.tab_textwidget.mark_set("insert", f"{str(next_anchor_row)}.0")
```
Akuli, line 63: Nit: f-strings automatically `str()` everything.

```suggestion
            self.tab_textwidget.mark_set("insert", f"{next_anchor_row}.0")
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,133 @@
...
+        cursor_row = self._get_cursor_index().split(".")[0]
+        anchor_list = self._get_anchors()
+        anchor_rows = [
+            int(self.tab_textwidget.index(anchorpoint).split(".")[0]) for anchorpoint in anchor_list
+        ]
+
+        rows_before_cursor = [n for n in anchor_rows if n < int(cursor_row)]
+        if rows_before_cursor:
+            previous_anchor_row = max(rows_before_cursor)
+            self.tab_textwidget.mark_set("insert", f"{str(previous_anchor_row)}.0")
```
Akuli, line 78: Same here

```suggestion
            self.tab_textwidget.mark_set("insert", f"{previous_anchor_row}.0")
```

Akuli (commented):
Last little nits

ThePhilgrim:
Check 3 "done".

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,143 @@
...
+        anchor_rows = [
+            int(self.tab_textwidget.index(anchorpoint).split(".")[0]) for anchorpoint in anchor_list
+        ]
+
+        rows_after_cursor = [n for n in anchor_rows if n > int(cursor_row)]
+        if rows_after_cursor:
+            next_anchor_row = min(rows_after_cursor)
+            self.tab_textwidget.mark_set("insert", f"{next_anchor_row}.0")
+            self.tab_textwidget.see("insert")
+        elif not rows_after_cursor and len(anchor_list) >= 2 and settings.get("anchors", bool):
```
Akuli, line 69: Nit: `rows_after_cursor` will always be False when this condition is considered, because if it isn't, then we choose the first part.

```suggestion
        elif len(anchor_list) >= 2 and settings.get("anchors", bool):
```

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,143 @@
+"""
+Allow for easy navigation in your files by toggling on/off anchorpoints,
+and jumpbing back and forth between them
```
Akuli, line 3:
```suggestion
and jumping back and forth between them.
```

porcupine/plugins/anchors.py
ThePhilgrim: Genius.

porcupine/plugins/anchors.py
ThePhilgrim: Woopsie.

porcupine/plugins/anchors.py
```diff
@@ -0,0 +1,143 @@
...
+        anchor_rows = [
+            int(self.tab_textwidget.index(anchorpoint).split(".")[0]) for anchorpoint in anchor_list
+        ]
+
+        rows_before_cursor = [n for n in anchor_rows if n < int(cursor_row)]
+        if rows_before_cursor:
+            previous_anchor_row = max(rows_before_cursor)
+            self.tab_textwidget.mark_set("insert", f"{previous_anchor_row}.0")
+            self.tab_textwidget.see("insert")
+        elif not rows_before_cursor and len(anchor_list) >= 2 and settings.get("anchors", bool):
```
Akuli, line 88: Forgot this one
```suggestion
        elif len(anchor_list) >= 2 and settings.get("anchors", bool):
```

Akuli (approved):
I'll make separate PR to add these to menu bar

PR #492, opened 2021-06-10
replace <<UpdateLineNumbers>> with accessing LineNumbers instance directly

### LINKED ISSUE #491: delete <<UpdateLineNumbers>> event
#490 changes `class LineNumbers` to expose its `do_update()` method, and it contains code that accesses the `LineNumbers` instance in a different plugin. For consistency, other plugins should do that too, and then we can delete `<<UpdateLineNumbers>>` because it is no longer used anywhere.

Akuli:
Fixes #491

### DIFF
--- a/porcupine/plugins/fold.py
+++ b/porcupine/plugins/fold.py
@@ -5,8 +5,6 @@ from typing import Optional
 from porcupine import get_tab_manager, menubar, tabs, utils
 from porcupine.plugins.linenumbers import LineNumbers
 
-setup_after = ["linenumbers"]
-
 
 def get_indent(tab: tabs.FileTab, lineno: int) -> Optional[int]:
     line = tab.textwidget.get(f"{lineno}.0", f"{lineno}.0 lineend")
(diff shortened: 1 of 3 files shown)


porcupine/plugins/fold.py
```diff
@@ -3,6 +3,9 @@
 from typing import Optional
 
 from porcupine import get_tab_manager, menubar, tabs, utils
+from porcupine.plugins.linenumbers import LineNumbers
+
+setup_after = ["linenumbers"]
```
Akuli, line 8: This missing from mergeconflict plugin

porcupine/plugins/mergeconflict.py
```diff
@@ -151,6 +152,9 @@ def setup_displayers(tab: tabs.FileTab) -> None:
 
 def on_new_tab(tab: tabs.Tab) -> None:
     if isinstance(tab, tabs.FileTab):
+        [linenums] = [
+            child for child in tab.left_frame.winfo_children() if isinstance(child, LineNumbers)
+        ]
```
Akuli, line 157: With a loop instead of this, fold plugin could be made to work when linenumbers plugin is disabled. Also applies to fold plugin.

porcupine/plugins/fold.py
Akuli, line 8: This is not needed, because LineNumbers stuff runs when folding.

PR #497, opened 2021-06-11
This fixes #485

### LINKED ISSUE #485: multiple plugin manager windows can be open at the same time
If a plugin manager is already open, then opening a new one shouldn't be possible. @ThePhilgrim's [minesweeper project](https://github.com/ThePhilgrim/minesweeper) has a high scores window that does this correctly, and Porcupine could use similar code.

adder46:
fixes #485

### DIFF
--- a/porcupine/plugins/pluginmanager.py
+++ b/porcupine/plugins/pluginmanager.py
@@ -6,7 +6,7 @@ import re
 import tkinter
 from functools import partial
 from tkinter import ttk
-from typing import List
+from typing import List, Optional
 
 from porcupine import get_main_window, menubar, pluginloader, settings, textwidget
 
--- a/porcupine/plugins/pluginmanager.py
+++ b/porcupine/plugins/pluginmanager.py
@@ -12,6 +12,7 @@ from porcupine import get_main_window, menubar, pluginloader, settings, textwidg
 
 log = logging.getLogger(__name__)
 
+dialog = None
 
 def get_docstring(module_name: str) -> str:
     try:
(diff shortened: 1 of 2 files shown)


porcupine/plugins/pluginmanager.py
```diff
@@ -12,6 +12,8 @@
 
 log = logging.getLogger(__name__)
 
+dialog = None
```
Akuli, line 15: You need to tell mypy that this is not always None, even though the initial assignment sets it to None:

```suggestion
dialog: Optional[tkinter.Toplevel] = None
```

This hides other bugs that we'll find with mypy once this suggestion is applied. I created https://github.com/python/mypy/issues/10634.

Akuli (changes requested):

tests/test_pluginmanager.py
```diff
@@ -3,14 +3,12 @@
...
-from porcupine.plugins.pluginmanager import show_dialog
+from porcupine.plugins.pluginmanager import create_dialog
 
 
 @pytest.fixture
 def dialog_content(mocker):
-    mocker.patch("tkinter.Toplevel.wait_window", autospec=True)
-    content = show_dialog()
-    tkinter.Toplevel.wait_window.assert_called_once()
+    content = create_dialog()
```
Akuli, line 11: It returns a tuple:

```suggestion
    dialog, content = create_dialog()
```

tests/test_pluginmanager.py
```diff
@@ -3,14 +3,12 @@
...
     yield content
     tkinter.Toplevel.wait_window.call_args[0][0].destroy()
```
Akuli, line 13: With my other suggestion, you can now access the dialog much more easily:

```suggestion
    dialog.destroy()
```

Akuli:
Thanks!

PR #499, opened 2021-06-12
Fix plugin setup order stuff

### LINKED ISSUE #493: plugin manager does not respect setup_before setup_after
Even if plugin A wants to be set up before plugin B, it's possible to disable A, then restart to actually get it disabled, and then enable A. This will run its setup, even though B's setup ran earlier when Porcupine started.

A couple alternatives for what should happen when enabling A with B already enabled:
- Ask user "Are you really want to enable", but let the user enable it anyway
- Require restart to enable (just like with disabling)

Akuli:
Fixes #493. Code was already there, but it was broken in a couple ways.

Todo:
- [x] add test

### DIFF
--- a/tests/test_pluginloader.py
+++ b/tests/test_pluginloader.py
@@ -25,6 +25,7 @@ def test_setup_order_bugs(monkeypatch):
         # can setup rstrip when autoindent is already active
         monkey.setattr(rstrip, "status", pluginloader.Status.DISABLED_BY_SETTINGS)
         assert pluginloader.can_setup_while_running(rstrip)
+
     with monkeypatch.context() as monkey:
         # but not the other way, autoindent must go first
         monkey.setattr(autoindent, "status", pluginloader.Status.DISABLED_BY_SETTINGS)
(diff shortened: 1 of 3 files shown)


tests/test_pluginloader.py
```diff
@@ -12,3 +12,20 @@ def test_all_plugins_loaded_successfully():
...
+    [rstrip] = [i for i in pluginloader.plugin_infos if i.name == "rstrip"]
+
+    assert autoindent.status == pluginloader.Status.ACTIVE
+    assert rstrip.status == pluginloader.Status.ACTIVE
+
+    with monkeypatch.context() as monkey:
+        # can setup rstrip when autoindent is already active
+        monkey.setattr(rstrip, "status", pluginloader.Status.DISABLED_BY_SETTINGS)
+        assert pluginloader.can_setup_while_running(rstrip)
+    with monkeypatch.context() as monkey:
```
Akuli, line 28: nit: blank line
```suggestion

    with monkeypatch.context() as monkey:
```

PR #502, opened 2021-06-12
update plugin dir description

### LINKED ISSUE #489: plugin api docs: document two places where to put plugin
When developing Porcupine plugins, it's usually best to put them to `cloned_repo_dir/porcupine/plugins/`, and treat the other possible folder as a place for already working plugins. The docs should say it more clearly.

```
<ThePhilgrim> it says in API to put it where that command tells you to put it
<Akuli> ThePhilgrim, then you can't mypy it
<Akuli> i should be more clear about it in the docs :)
<ThePhilgrim> porcupine folder: /Users/Writing/documents/Development/Programming/Python/others_projects/porcupine
<ThePhilgrim> anchors:
<ThePhilgrim> ./Users/Writing/Library/Application Support/Porcupine/plugins
<Akuli> put it to the same place with linenumbers.py
<ThePhilgrim> and leave it there or do I need to put it back later?
<Akuli> leave it there
<Akuli> porcupine looks for plugins in two places
<Akuli> if you have it in either place, it will work
<ThePhilgrim> ahhh
<ThePhilgrim> Yeah, API docs need to say that xD
```

Akuli:
Fixes #489. Needs @ThePhilgrim's review.

### DIFF
--- a/docs/plugin-intro.rst
+++ b/docs/plugin-intro.rst
@@ -42,7 +42,7 @@ The plugins that come with Porcupine are in a different ``plugins`` folder::
 
 In the rest of this tutorial, **the plugin directory** means the place where you
 will put your plugin. It can be either of the two ``plugins`` folders.
-If your plugin will be included with Porcupine, then you shouldn't use your
+If you want your plugin to come with Porcupine, then you shouldn't use your
 user-specific ``plugins`` folder.
 
 
--- a/docs/plugin-intro.rst
+++ b/docs/plugin-intro.rst
@@ -20,7 +20,7 @@ Where are the plugins?
 ----------------------
 
 Porcupine searches for plugins in two places. The plugins that come with
-Porcupine are installed along with it, but Porcupine also creates a user-wide
+Porcupine are installed along with it, but Porcupine also creates a user-specific
 ``plugins`` folder in its config directory. You can check where it is using
 Python::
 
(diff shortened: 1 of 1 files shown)


docs/plugin-intro.rst
```diff
@@ -19,26 +19,31 @@ Porcupine's API documentation assume
...
+Your plugin directory is empty by default.
+The plugins that come with Porcupine are in a different ``plugins`` folder::
 
     >>> print(porcupine.plugins.__path__[1])
-    /home/akuli/.local/lib/python3.7/site-packages/porcupine/plugins
+    /home/akuli/porcu/porcupine/plugins
+
+In the rest of this tutorial, **the plugin directory** means the place where you
+will put your plugin. It can be either of the two ``plugins`` folders.
+If your plugin will be included with Porcupine, then you shouldn't use your
```
ThePhilgrim, line 45: > If your plugin will be included with Porcupine

Maybe something like "if you want to publish your plugin so others can use it"

ThePhilgrim, line 45:
```suggestion
If your plugin will be included with Porcupine, then you should not use your
```

ThePhilgrim (changes requested):

docs/plugin-intro.rst
Akuli: It's possible to publish a Porcupine plugin without including it in Porcupine, because it's just a Python file. It's fine to develop plugins like that with your user-specific plugin directory.

ThePhilgrim: Ah right, so maybe "if you want your plugin to come with Porcupine"? "be a stock plugin"? – something similar.. "be included with porcupine" feels slightly unclear.

Akuli: I like "If you want your plugin to come with Porcupine". Thanks!

ThePhilgrim (approved):

PR #505, opened 2021-06-12
create settings.add_checkbutton()

Akuli:
For #490

### DIFF
--- a/porcupine/settings.py
+++ b/porcupine/settings.py
@@ -595,7 +595,7 @@ def add_checkbutton(option_name: str, **checkbutton_kwargs: Any) -> ttk.Checkbut
                 print("Foobar disabled")
 
         def setup() -> None:
-            settings.add_option("foobar", False, bool)  # False is default value
+            settings.add_option("foobar", False)  # False is default value
             settings.add_checkbutton("foobar", text="Enable foobar")
 
     Currently it is not possible to display a |triangle| next to the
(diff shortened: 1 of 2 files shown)


porcupine/settings.py
```diff
@@ -578,6 +578,47 @@ def add_entry(
...
+        from porcupine import settings
+
+        def do_something() -> None:
+            if settings.get("foobar", bool):
+                print("Foobar enabled")
+            else:
+                print("Foobar disabled")
+
+        def setup() -> None:
+            settings.add_option("foobar", False, bool)  # False is default value
```
Akuli, line 596: It uses `type(False)` if `bool` is not given:

```suggestion
            settings.add_option("foobar", False)  # False is default value
```

porcupine/settings.py
```diff
@@ -578,6 +578,47 @@ def add_entry(
     return entry
 
 
+def add_checkbutton(option_name: str, **checkbutton_kwargs: Any) -> ttk.Entry:
+    """Add a :class:`tkinter.ttk.Checkbutton` to the setting dialog.
+
+    All ``**checkbutton_kwargs`` go to :class:`tkinter.ttk.Entry`.
```
Akuli, line 584:
```suggestion
    All ``**checkbutton_kwargs`` go to :class:`tkinter.ttk.Checkbutton`.
```

PR #511, opened 2021-06-12
Antiboilerplate

### LINKED ISSUE #510: The filetab-specific instance boilerplate
Many plugins want to make an instance of a class for every `FileTab`, and then run a method of the instance whenever an item in the menubar is clicked. This sounds simple, but it's currently about 20 lines of boilerplate code:
-  You need to create a `weakref.WeakKeyDictionary` with `FileTab`s as keys and instances of the class as values. Usual dicts aren't great because they would grow and grow and grow if you add and close tabs, making Python think that the old tabs are still needed because they are in the dict.
- You need to create a function to be used with `get_tab_manager().add_tab_callback`, which will check whether a newly added tab is a `FileTab`, and if it is, make an instance of your class and put it to the `WeakKeyDictionary`.
- Adding a menu item involves two steps: you need to use `menubar.get_menu(...).add_command` and `menubar.set_enabled_based_on_tab`.

Akuli:
Fixes #510

### DIFF
--- a/porcupine/menubar.py
+++ b/porcupine/menubar.py
@@ -10,7 +10,7 @@ import webbrowser
 from functools import partial
 from string import ascii_lowercase
 from tkinter import filedialog
-from typing import Callable, Iterator, Optional, Tuple
+from typing import Callable, Iterator, Optional, Tuple, Any
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -269,7 +269,7 @@ def _get_filetab() -> tabs.FileTab:
     return tab
 
 
-def add_filetab_command(path: str, func: Callable[[tabs.FileTab], None] | None = None) -> None:
+def add_filetab_command(path: str, func: Callable[[tabs.FileTab], Any] | None = None) -> None:
     """
     This is a convenience function that does several things:
 
@@ -314,7 +314,7 @@ def add_filetab_command(path: str, func: Callable[[tabs.FileTab], None] | None =
     if func is None:
         command = lambda: _get_filetab().event_generate(f"<<FiletabCommand:{path}>>")
     else:
-        command = lambda: func(_get_filetab())
+        command = lambda: func(_get_filetab())  # type: ignore
 
     menu_path, item_text = path.rsplit("/", maxsplit=1)
     get_menu(menu_path).add_command(label=item_text, command=command)
--- a/porcupine/menubar.py
+++ b/porcupine/menubar.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import functools
+from functools import partial
 import logging
 import pathlib
 import re
@@ -48,6 +48,7 @@ log = logging.getLogger(__name__)
 # before root.mainloop(), then it works, so that has to be done for every
 # text widget.
 def _generate_event(name: str, junk: object) -> Literal["break"]:
+    log.debug(f"Generating event: {name}")
     get_main_window().event_generate(name)
     return "break"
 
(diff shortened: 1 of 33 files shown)


Akuli:
Note to self: viewed in review tab = plugin tested that it still works

porcupine/plugins/comment_block.py
```diff
@@ -51,24 +51,15 @@ def comment_or_uncomment(
...
-def on_new_tab(tab: tabs.Tab) -> None:
-    if isinstance(tab, tabs.FileTab):
-        tab.textwidget.bind("<Key>", functools.partial(comment_or_uncomment, tab), add=True)
+def on_new_filetab(tab: tabs.FileTab) -> None:
+    tab.bind(
+        "<<FiletabCommand:Edit/Comment Block>>",
+        (lambda event: comment_or_uncomment(tab, None)),
+        add=True,
+    )
+    tab.textwidget.bind("<Key>", functools.partial(comment_or_uncomment, tab), add=True)
```
Akuli, line 60: The callbacks are a bit confusing. Maybe `comment_or_uncomment` should take the pressed key (event.char) as argument?

porcupine/plugins/langserver.py
```diff
@@ -727,16 +727,13 @@ def switch_langservers(
...
+    tab.settings.add_option("langserver", None, Optional[LangServerConfig])
 
-        tab.bind(
-            "<<TabSettingChanged:langserver>>", partial(switch_langservers, tab, False), add=True
-        )
-        tab.bind("<<PathChanged>>", partial(switch_langservers, tab, True), add=True)
-        switch_langservers(tab, False)
+    tab.bind("<<TabSettingChanged:langserver>>", partial(switch_langservers, tab, False), add=True)
+    tab.bind("<<PathChanged>>", partial(switch_langservers, tab, True), add=True)
+    switch_langservers(tab, False)
```
Akuli, line 735:
```suggestion
    switch_langservers(tab, called_because_path_changed=False)
```

porcupine/menubar.py
```diff
@@ -264,6 +263,40 @@ def update_enabledness(*junk: object) -> None:
...
 
+def _generate_filetab_event(path: str) -> None:
+    event = f"<<FiletabCommand:{path}>>"
+    log.debug(f"Generating event on selected filetab: {event}")
+    tab = get_tab_manager().select()
+    assert isinstance(tab, tabs.FileTab)
+    tab.event_generate(event)
+
+
+def add_filetab_command(path: str) -> None:
```
Akuli, line 274: The new boilerplate seems to be to call this, and then create `on_new_filetab()` function just to bind to event. Change this to take optional function that will be called with filetab as argument.

porcupine/plugins/anchors.py
```diff
@@ -124,40 +123,30 @@ def prevent_duplicate_anchors(self) -> None:
...
+        child for child in tab.left_frame.winfo_children() if isinstance(child, LineNumbers)
+    ]
+    manager = AnchorManager(tab.textwidget, linenumbers)
+    tab.bind(
+        "<<FiletabCommand:Edit/Anchors/Add or remove on this line>>",
+        manager.toggle_on_off,
+        add=True,
+    )
+    tab.bind("<<FiletabCommand:Edit/Anchors/Jump to previous>>", manager.jump_to_previous, add=True)
+    tab.bind("<<FiletabCommand:Edit/Anchors/Jump to next>>", manager.jump_to_next, add=True)
```
Akuli, line 137:
```suggestion
    # fmt: off
    tab.bind("<<FiletabCommand:Edit/Anchors/Add or remove on this line>>", manager.toggle_on_off, add=True)
    tab.bind("<<FiletabCommand:Edit/Anchors/Jump to previous>>", manager.jump_to_previous, add=True)
    tab.bind("<<FiletabCommand:Edit/Anchors/Jump to next>>", manager.jump_to_next, add=True)
    # fmt: on
```

PR #530, opened 2021-06-19
Directory tree: run git statuses in parallel

Akuli:
Refreshing the directory tree happens quite often, and it usually takes between 50ms and 100ms each time. Running git is the bottleneck of refreshing, and letting 2 gits run at the same time (if you have 2 cpu cores) should help. Python's GIL doesn't affect this, because gits run in subprocesses.

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -75,7 +75,6 @@ def run_git_status(project_root: Path) -> Dict[Path, str]:
             result[path] = "git_mergeconflict"
         else:
             log.warning(f"unknown git status line: {repr(line)}")
-    print("git status:", project_root, "-->", result)
     return result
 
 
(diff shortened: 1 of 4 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -73,9 +75,14 @@ def run_git_status(project_root: Path) -> Dict[Path, str]:
             result[path] = "git_mergeconflict"
         else:
             log.warning(f"unknown git status line: {repr(line)}")
+    print("git status:", project_root, "-->", result)
```
Akuli, line 78: Delete debug prints. This isn't the only one.

porcupine/plugins/directory_tree.py
```diff
@@ -283,35 +290,41 @@ def _hide_old_projects(self, junk: object = None) -> None:
...
-            return {path: run_git_status(path) for path in map(get_path, project_ids)}
+        project_ids = self.get_children("")
+        git_futures = {
+            path: _git_pool.submit(partial(run_git_status, path))
+            for path in map(get_path, project_ids)
+        }
+
+        def check_if_done() -> None:
+            print("check if done")
+            if not all(future.done() for future in git_futures.values()):
```
Akuli, line 306: Does this get stuck if a future errors?

Akuli: no

porcupine/plugins/directory_tree.py
```diff
@@ -283,35 +289,36 @@ def _hide_old_projects(self, junk: object = None) -> None:
...
                 for project_id in self.get_children(""):
                     self._update_tags_and_content(get_path(project_id), project_id)
                 self._update_selection_color()
                 log.info(f"refreshing done in {round((time.time()-start_time)*1000)}ms")
-                when_done()
-            elif success:
+            else:
                 log.info(
                     "projects added/removed while refreshing, assuming another fresh is coming soon"
                 )
```
Akuli, line 317: "if not babla: ... return" would make more sense

Akuli: no because `done_callback()` must run either way

PR #533, opened 2021-06-20
directory tree: show nested projects in a way that isn't confusing

### LINKED ISSUE #522: directory tree jumps with nested projects
With nested projects, the same file can be visible in two projects. When I have a file open and I browse the directory tree, it tries to focus the file I have open whenever I open a directory. This is great except that it always highlights the file in the first project containing it, not the project I'm browsing.

### LINKED ISSUE #528: directory tree: Nested projects suck
- Virtual envs: If the project `/home/akuli/porcu` uses `/home/akuli/porcu/env` as a virtual env, it's not visible when you view `/home/akuli/porcu` as a subproject of `/home/akuli`. In fact, if you right-click and say "Use this Python venv", it uses it as the venv for `/home/akuli`, not for `/home/akuli/porcu`. This could cause a lot of confusion.
- Jumping (#522)

When viewing a project that contains another project, it would be better if the other project would contain a dummy item showing a text like "(already open as separate project)" or something.

Akuli:
Fixes #522
Fixes #528

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -2,10 +2,12 @@ from __future__ import annotations
 
 import dataclasses
 import logging
+import os
 import subprocess
 import sys
 import time
 import tkinter
+from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 from pathlib import Path
 from tkinter import ttk
@@ -74,6 +76,10 @@ def run_git_status(project_root: Path) -> Dict[Path, str]:
         else:
             log.warning(f"unknown git status line: {repr(line)}")
     return result
+
+
+# Each git subprocess uses one cpu core
+_git_pool = ThreadPoolExecutor(max_workers=os.cpu_count())
 
 
 # For perf reasons, we want to avoid unnecessary Tcl calls when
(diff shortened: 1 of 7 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -232,38 +232,37 @@ def add_project(self, root_path: Path, *, refresh: bool = True) -> None:
...
-            if get_path(project_id) not in path.parents:
-                continue
-
-            path_to_root = list(_path_to_root_inclusive(path, get_path(project_id)))
-            root_to_path_excluding_root = path_to_root[::-1][1:]
+        matching_projects = [project_id for project_id in self.get_children() if get_path(project_id) in path.parents]
+        if not matching_projects:
+            # Happens when tab changes because a file was just opened. This
+            # will be called soon once the project has been added.
+            log.info(f"can't select '{path}', found no projects containing it")
```
Akuli, line 239:
```suggestion
            log.info(f"can't select '{path}' because there are no projects containing it")
```

porcupine/plugins/directory_tree.py
```diff
@@ -365,12 +364,19 @@ def _open_and_refresh_directory(self, dir_path: Path, dir_id: str) -> None:
         if self.contains_dummy(dir_id):
             self.delete(self.get_children(dir_id)[0])  # type: ignore[no-untyped-call]
 
+        project_ids = self.get_children("")
+        if dir_id not in project_ids and dir_path in map(get_path, project_ids):
+            for child in self.get_children(dir_id):
+                self.delete(child)  # type: ignore[no-untyped-call]
+            self._insert_dummy(dir_id, text="(open as a separate project)")
```
Akuli, line 371: Maybe `_insert_dummy` should include the loop?

PR #538, opened 2021-06-23
dir tree: show ~

### LINKED ISSUE #536: directory tree home folder
ss

Would be better to show `~/porcu` and `~`

Akuli:
Fixes #536

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -111,6 +111,14 @@ class FolderRefreshed(utils.EventDataclass):
     folder_id: str
 
 
+# TODO: show long paths more nicely?
+def _stringify_path(path: Path) -> str:
+    home = Path.home()
+    if path == home or home in path.parents:
+        return os.sep.join(["~"] + list(path.relative_to(home).parts))
+    return str(path)
+
+
 class DirectoryTree(ttk.Treeview):
     def __init__(self, master: tkinter.Misc) -> None:
         super().__init__(master, selectmode="browse", show="tree", style="DirectoryTree.Treeview")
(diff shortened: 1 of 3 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -216,15 +224,14 @@ def add_project(self, root_path: Path, *, refresh: bool = True) -> None:
...
-        text = str(root_path)
-        if Path.home() in root_path.parents:
-            text = text.replace(str(Path.home()), "~", 1)
-
         # Add project to beginning so it won't be hidden soon
         self._project_num_counter += 1
         project_item_id = self.insert(
-            "", 0, f"project:{self._project_num_counter}:{root_path}", text=text, open=False
+            "",
+            0,
```
Akuli, line 231: The comment should move after `0`

tests/test_directory_tree_plugin.py
```diff
@@ -3,7 +3,7 @@
 import sys
 from concurrent.futures import Future
 from functools import partial
-from pathlib import Path
+from pathlib import Path, PurePosixPath
```
Akuli, line 6:
```suggestion
from pathlib import Path
```

porcupine/plugins/directory_tree.py
```diff
@@ -216,15 +224,13 @@ def add_project(self, root_path: Path, *, refresh: bool = True) -> None:
...
-        if Path.home() in root_path.parents:
-            text = text.replace(str(Path.home()), "~", 1)
-
-        # Add project to beginning so it won't be hidden soon
         self._project_num_counter += 1
         project_item_id = self.insert(
-            "", 0, f"project:{self._project_num_counter}:{root_path}", text=text, open=False
+            "",
+            0,  # beginning, so it won't be hidden soon
+            f"project:{self._project_num_counter}:{root_path}",
```
Akuli, line 231: It's confusing how this is here but it ends up being the `project_item_id`, and there's also a loop with `project_item_id`

PR #540, opened 2021-06-25
rename textwidget to textutils

### LINKED ISSUE #534: porcupine.textwidget is a confusing name
When ThePhilgrim was getting started with Porcupine, he constantly confused `from porcupine import textwidget` with variables that were named `textwidget`. To prevent this, `textwidget.py` should be renamed to `text_utils.py`, for example.

Akuli:
Fixes #534


porcupine/plugins/autocomplete.py
```diff
@@ -213,7 +213,7 @@ def on_page_up_down(self, event: tkinter.Event[tkinter.Misc]) -> utils.BreakOrNo
             return None
 
         page_count = {"Prior": -1, "Next": 1}[event.keysym]
-        self._doc_text.yview_scroll(page_count, "pages")  # type: ignore[no-untyped-call]
+        self._doc_text.yview_scroll(page_count, "pages")  # type: ignore
```
Akuli, line 216:
```suggestion
        self._doc_text.yview_scroll(page_count, "pages")  # type: ignore[no-untyped-call]
```

PR #543, opened 2021-06-25
Delete porcupine.utils.TemporaryBind

### LINKED ISSUE #542: delete utils.TemporaryBind
It's used only in langserver, and the langserver code would be simpler without it.

Akuli:
Fixes #542


porcupine/plugins/langserver.py
```diff
@@ -573,17 +564,14 @@ def forget_tab(self, tab: tabs.FileTab, *, may_shutdown: bool = True) -> None:
...
-    def request_completions(self, event: utils.EventWithData) -> None:
+    def request_completions(self, tab: tabs.FileTab, event: utils.EventWithData) -> None:
         if self._lsp_client.state != lsp.ClientState.NORMAL:
             self.log.warning(
                 f"autocompletions requested but langserver state == {self._lsp_client.state!r}"
             )
             return
 
-        tab = event.widget
-        assert isinstance(tab, tabs.FileTab) and tab.path is not None
```
Akuli, line 584: Presumably `tab.path is not None` still needs to be asserted to make mypy happy.

porcupine/plugins/langserver.py
```diff
@@ -639,7 +627,7 @@ def stream_to_log(stream: IO[bytes], log: logging.LoggerAdapter) -> None:
         log.info(f"langserver logged: {line}")
 
 
-def get_lang_server(tab: tabs.FileTab) -> Optional[LangServer]:
+def get_lang_server(tab: tabs.FileTab) -> LangServer | None:
```
Akuli, line 630: Might as well change all `Optional`s

PR #549, opened 2021-06-27
Jump to definition

Akuli:
this is awesome

- [x] create chooser menu when multiple definitions
- [x] fix ctrl+enter

PR #553, opened 2021-06-27
properly filter away error from hack

Akuli:
Broken in sansio-lsp-client update of #549


porcupine/plugins/langserver.py
```diff
@@ -39,6 +39,19 @@
 CHUNK_SIZE = 64 * 1024
 
 
+# A hack later in this file causes an event that sansio-lsp-client doesn't understand.
```
Akuli, line 42: maybe move to end? and `later` --> `elsewhere`

PR #567, opened 2021-07-06
Minimap cursor improvement

rdbende:
Not a big improvement, but the `xterm` cursor just looks strange in the minimap.

### DIFF
--- a/porcupine/plugins/minimap.py
+++ b/porcupine/plugins/minimap.py
@@ -34,6 +34,7 @@ class MiniMap(tkinter.Text):
             takefocus=False,
             yscrollcommand=self._update_vast,
             wrap="none",
+            cursor="arrow",
         )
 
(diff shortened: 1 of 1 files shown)


Akuli (approved):
Thanks!

Akuli:
Are you interested in working on Porcupine more? As I mentioned in README, I can help you with whatever it is you want to do.

rdbende:
Yes, I'd like to contribute more. I use Porcupine on a daily basis and would enjoy its development, although I am not very familiar with static typing and mypy, but with tkinter all the more so.

Akuli:
If you want to talk with me in real time to get help as fast as possible, you can click the IRC link in readme. Basics of mypy and static typing aren't that difficult IMO, and easily knowing the types of everything help a lot with larger projects.

PR #568, opened 2021-07-06
Fix url plugin backticks

rdbende:
Before:
image

After:
image

### DIFF
--- a/porcupine/plugins/urls.py
+++ b/porcupine/plugins/urls.py
@@ -29,8 +29,8 @@ def find_urls(text: tkinter.Text, start: str, end: str) -> Iterable[Tuple[str, s
         url = url.split("'")[0]
         url = url.split('"')[0]
 
-        open2close = {"(": ")", "{": "}", "<": ">"}
-        close2open = {")": "(", "}": "{", ">": "<"}
+        open2close = {"(": ")", "{": "}", "<": ">", "`": "`"}
+        close2open = {")": "(", "}": "{", ">": "<", "`": "`"}
 
         if before_url in open2close and open2close[before_url] in url:
             # url is parenthesized
--- a/porcupine/plugins/urls.py
+++ b/porcupine/plugins/urls.py
@@ -28,9 +28,10 @@ def find_urls(text: tkinter.Text, start: str, end: str) -> Iterable[Tuple[str, s
         url = url.split(" ")[0]
         url = url.split("'")[0]
         url = url.split('"')[0]
+        url = url.split('`')[0]
 
-        open2close = {"(": ")", "{": "}", "<": ">", "`": "`"}
-        close2open = {")": "(", "}": "{", ">": "<", "`": "`"}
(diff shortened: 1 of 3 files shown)


Akuli:
Awesome! Can you add a test case to `test_find_urls` in `tests/test_urls_plugin.py`? It can be a lot like the other tests with quotes and parentheses and stuff.

porcupine/plugins/urls.py
```diff
@@ -29,8 +29,8 @@ def find_urls(text: tkinter.Text, start: str, end: str) -> Iterable[Tuple[str, s
         url = url.split("'")[0]
         url = url.split('"')[0]
 
-        open2close = {"(": ")", "{": "}", "<": ">"}
-        close2open = {")": "(", "}": "{", ">": "<"}
+        open2close = {"(": ")", "{": "}", "<": ">", "`": "`"}
+        close2open = {")": "(", "}": "{", ">": "<", "`": "`"}
```
Akuli, line 33: It might be better to add another `.split()`, so that it's handled just like any other quotes. I'm not sure if it makes a practical difference, but I'd like to avoid weird gotchas.

rdbende: I looked over several times to see which was better and thought I had tried every situation, but now I realized that the `.split` solution is still better, because with the open-close method you can get something like this:

image

tests/test_urls_plugin.py
```diff
@@ -46,6 +46,50 @@ def test_find_urls():
...
+    urls = [
+        "https://example.com/single`quote",
+        "https://example.com/double`quote",
+        "https://example.com/backtick`something",
+    ]        
+    test_cases = """\
+          URL
+          URL bla bla
+"See also URL"
+         'URL bla'
```
Akuli, line 61: I don't think we need to copy/paste all this. Just add a line like this to the other test above:
```
         `URL`
```

You will also need to run `black tests/test_urls_plugin.py`

Akuli (changes requested):

porcupine/plugins/urls.py
```diff
@@ -28,6 +28,7 @@ def find_urls(text: tkinter.Text, start: str, end: str) -> Iterable[Tuple[str, s
         url = url.split(" ")[0]
         url = url.split("'")[0]
         url = url.split('"')[0]
+        url = url.split('`')[0]
```
Akuli, line 31: Black likes double quotes (could also run `black urls.py`):

```suggestion
        url = url.split("`")[0]
```

Akuli (approved):

porcupine/plugins/urls.py
rdbende: Thanks. I know it's a bad practice but I normally use single quote because I like it better. Fixed.

Akuli: It's not "bad practice", it's just not how black does it. Black is very opinionated by design, and there's lots of things that aren't bad but it changes anyway.

Akuli:
Thanks!

PR #574, opened 2021-07-07
create blockcursor plugin

rdbende:
Turns the blockcursor on or off for the textwidgets with a checkbutton in the settings.

image


Akuli:
A few problems:
- It looks really thick for some reason. I use a monospace font, but it still touches the letters next to the currently highlighted letter.

https://user-images.githubusercontent.com/18505570/124765866-11d46d00-df3f-11eb-9e32-0a57c27fa36e.mp4

- If I have two tabs opened and I enable/disable blockcursor in one tab, it doesn't get enabled/disabled in the other tab. The solution would be to bind to `<<SettingChanged:blockcursor>>` and use `get_tab_manager().add_filetab_callback()`. These events are documented [here](https://akuli.github.io/porcupine/settings.html#change-events). See e.g. `LineNumbers.__init__` in `linenumbers.py` or `LongLineMarker.setup` in `longlinemarker.py` for examples.
- I don't think pressing the insert key is an appropriate way to choose whether or not you want to use the blockcursor. A checkbox in the settings dialog would be better (see `anchor.py` for an example of how to create a checkbutton there). Many Windows programs change to "typing a character overrides the next character" mode when pressing insert, and even though it's tempting to associate that with a blockcursor, that's not what this plugin does.

porcupine/plugins/blockcursor.py
```diff
@@ -0,0 +1,27 @@
+"""Toggles between the ❚ and | cursor state of the textwidgets."""
+from __future__ import annotations
+
+import tkinter
+
+from porcupine import get_tab_manager, settings, tabs
+
+
+def do_toggle(event: tkinter.Event[tkinter.Text]) -> None:
+    if settings.get("blockcursor", object):
```
Akuli, line 10: Nit:
```suggestion
    if settings.get("blockcursor", bool):
```

porcupine/plugins/blockcursor.py
```diff
@@ -0,0 +1,27 @@
...
+        event.widget.configure(blockcursor=True, insertwidth=0)  # minimize the cursor thickness
+    else:
+        event.widget.configure(
+            blockcursor=False, insertwidth=2
+        )  # set insertwidth back to tk default
+
+
+def on_new_filetab(tab: tabs.FileTab) -> None:
+    is_block = settings.get("blockcursor", object)
+    tab.textwidget.config(blockcursor=is_block, insertwidth=0 if is_block else 2)
```
Akuli, line 20: To avoid fragile copy/pasta, it would be better to do something like this:

```suggestion
    do_toggle(tab)
```

Then `do_toggle()` has to be changed so that it accepts a text widget, not an event. When binding, you can use e.g. `lambda event: do_toggle(tab)`.

porcupine/plugins/blockcursor.py
```diff
@@ -0,0 +1,27 @@
...
+    tab.textwidget.bind("<<SettingChanged:blockcursor>>", do_toggle, add=True)
+
+
+def setup() -> None:
+    settings.add_option("blockcursor", False)
+    settings.add_checkbutton("blockcursor", text="Use blockcursor")
```
Akuli, line 26:
```suggestion
    settings.add_checkbutton("blockcursor", text="Show cursor as a block instead of a thin line")
```

Akuli (commented):
It's starting to look pretty good! A few nits.

porcupine/plugins/blockcursor.py
rdbende: Thanks, it's much better indeed.

porcupine/plugins/blockcursor.py
rdbende: I can't write descriptive texts XD

rdbende:
The usage of the insert key is completely confusing. For example, Gedit and Sublime Text are use it to toggle the insertion cursor, while Vim is use it to overwrite the next char.

Akuli:
dialog

It's a little weird that the checkboxes aren't together. Add `setup_after = ["filetypes"]` along with an appropriate comment for now. It's a bit weird, but it works and it can be changed later.

rdbende:
Thanks. This is something I wanted to do, I just didn’t know about this feature.

Akuli (approved):
Thanks!

PR #575, opened 2021-07-07
comment_block plugin: distinguish '# foo' with one vs many spaces

### LINKED ISSUE #562: comment_block: can't uncomment
uncomment

It will add more `#` when I try to uncomment

Akuli:
Fixes #562

### DIFF
--- a/porcupine/plugins/comment_block.py
+++ b/porcupine/plugins/comment_block.py
@@ -13,12 +13,6 @@ from typing import Optional
 from porcupine import get_tab_manager, menubar, tabs, textutils
 
 
-def is_commented(line: str, comment_prefix: str) -> bool:
-    # Don't ignore indented '#    blah', that is most likely by this plugin
-    # But ignore '# blah' comments because they are likely written by hand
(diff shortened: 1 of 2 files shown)


porcupine/plugins/comment_block.py
```diff
@@ -32,17 +33,19 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
...
         lineno
-        for lineno in all_linenos
-        if tab.textwidget.get(f"{lineno}.0") == comment_prefix
-        # Do not touch comments starting with '# '
-        and tab.textwidget.get(f"{lineno}.1") != " "
+        for lineno, line in enumerate(
+            tab.textwidget.get(f"{start}.0", f"{end}.0").splitlines(), start
+        )
+        # Don't ignore indented '#    blah', that is most likely by this plugin
+        # But ignore '# blah' comments because they are likely written by hand
```
Akuli, line 40:
```suggestion
        # Ignore '# blah' comments because they are likely written by hand
        # But don't ignore indented '#    blah', that is most likely by this plugin
```

PR #583, opened 2021-07-09
fix pluginmanager problems

rdbende:
Fix treeview collapsing by switching to `tkinter.PanedWindow`, set panedwindow bg to ttk style, set minsize to panes, little margin around buttons, disabled button state by default (because it was annoying, that you can disable unselected plugins).

### DIFF
--- a/porcupine/plugins/pluginmanager.py
+++ b/porcupine/plugins/pluginmanager.py
@@ -49,15 +49,23 @@ class PluginDialogContent:
     def __init__(self, master: tkinter.Misc) -> None:
         self.content_frame = ttk.Frame(master)
 
-        panedwindow = ttk.Panedwindow(self.content_frame, orient="horizontal")
+        _cols_width = [120, 150, 180]
+
+        # borrowed code from textutils.create_passive_text_widget
+        ttk_bg = self.content_frame.tk.eval("ttk::style lookup TLabel.label -background")
+        if not ttk_bg:
+            ttk_bg = "white"
+
+        panedwindow = tkinter.PanedWindow(self.content_frame, orient="horizontal", bg=ttk_bg)
         panedwindow.pack(side="top", fill="both", expand=True)
+
         self._plz_restart_label = ttk.Label(self.content_frame)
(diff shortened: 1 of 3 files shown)


rdbende:
lol when code checks are fine then platform tests will start to fail.

rdbende:
`test_pluginmanager.py` fixed. The test didn't know that the buttons were disabled by default.

porcupine/plugins/pluginmanager.py
```diff
@@ -79,43 +87,42 @@ def __init__(self, master: tkinter.Misc) -> None:
...
         scrollbar.pack(side="right", fill="y")
 
-        for index, width in enumerate([100, 150, 180]):
+        for index, width in enumerate(_cols_width):
             self.treeview.column(index, width=width, minwidth=width)
 
-        self.treeview.heading(0, text="Name")
-        self.treeview.heading(1, text="Type")
-        self.treeview.heading(2, text="Status")
+        for index, header_title in enumerate(["Name", "Type", "Status"]):  # be consistent
```
Akuli, line 93: Comment doesn't make sense to me, but this is simple code, and doesn't require a comment IMO:
```suggestion
        for index, header_title in enumerate(["Name", "Type", "Status"]):
```

Akuli (commented):
Looks good! A small nit.

porcupine/plugins/pluginmanager.py
rdbende: lol, that wasn't intentional

tests/test_pluginmanager.py
```diff
@@ -45,6 +45,7 @@ def get_states():
 def test_disable_itself(mocker, dialog_content):
     mock = mocker.patch("tkinter.messagebox.askokcancel", return_value=False)
     dialog_content.treeview.selection_set(["pluginmanager"])
+    dialog_content.disable_button.state(["!disabled"])
```
Akuli, line 48: It might be better to run `_on_select()`, which sets the buttons enabled or disabled:

```suggestion
    dialog_content.treeview.event_generate("<<TreeviewSelect>>")
```

PR #584, opened 2021-07-09
Add missing plugindescriptions

rdbende:
Add missing plugindescriptions for black, directory_tree and sort plgins.

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -1,5 +1,5 @@
 """
-Displays a directory tree on the left side of the editor with the folders containing the 5 most recently edited files.
+Displays a directory tree on the left side of the editor.
 
 You can navigate directories, and open files in Porcupine.
 """
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -1,3 +1,9 @@
+"""
+Displays a directory tree on the left side of the editor with the folders containing the 5 most recently edited files.
+
+You can navigate directories, and open files in Porcupine.
+"""
+
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -1,3 +1,9 @@
+"""
+Displays a directory tree on the left side of the editor with the folders containing the 5 most recently edited files.
```
Akuli, line 2: A couple problems with 5 most recently:
- The 5 most recent things are projects, not files. For this plugin, "project" and "directory" are slightly different concepts.
- The 5 is not true in all cases. There can be more or less. See the comment above `MAX_PROJECTS`.

I think a good solution is to just delete this:

```suggestion
Displays a directory tree on the left side of the editor.
```

porcupine/plugins/sort.py
```diff
@@ -1,3 +1,9 @@
+"""
+Sorts the selected rows, or if there's no selection, to the next blank line.
```
Akuli, line 2: A few nits:
- Sorting happens alphabetically, which is kind of dumb in many contexts, and not necessarily what you want. That should be mentioned.
- rows --> lines. This seems to be more standard and Porcupine already shows "Line 123, column 45" in status bar.
- The logic isn't "to the next blank line". It finds the previous and next blank line, and sorts between them.

```suggestion
Sort the selected lines alphabetically, or if there's no selection, sort between surrounding blank lines.
```

rdbende:
Thanks for the suggestions!. The thing is, I actually wrote these based on my experience, so what others see during use, not what the code exactly does.

Akuli:
Thanks!

PR #589, opened 2021-07-10
Find area improvements

### LINKED ISSUE #479: find/replace: disabling "Replace this match" works inconsistently
It should be grayed out until "Next match" is clicked. So for example, if you just open find area, without clicking anything else, it should be grayed out. But it isn't. We discovered this in #476.

Akuli:
- Fixes #479.
- Adds "Match 2/5" kind of thing when you click "Previous match" or "Next match".
- If cursor is at start of a match, but it isn't highlighted, then "Next match" will choose it instead of whatever comes after it.

### DIFF
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -186,14 +186,11 @@ class Finder(ttk.Frame):
         # "find_highlight_selected". Both have to match.
         locations = [str(loc) for loc in self._textwidget.tag_ranges("sel")]
         locations2 = [str(loc) for loc in self._textwidget.tag_ranges("find_highlight_selected")]
-        if (
+        replace_this_state = "normal" if (
             locations == locations2
             and len(locations) == 2
             and tuple(locations) in self.get_match_ranges()
-        ):
-            replace_this_state = "normal"
-        else:
-            replace_this_state = "disabled"
+        ) else "disabled"
 
         self.previous_button.config(state=matches_something_state)
         self.next_button.config(state=matches_something_state)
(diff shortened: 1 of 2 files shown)


tests/test_find_plugin.py
```diff
@@ -201,9 +200,12 @@ def test_basic_statuses_and_previous_and_next_match_buttons(filetab_and_finder):
     assert finder.statuslabel["text"] == "Found 5 matches."
 
     def get_selected():
```
Akuli, line 202: This function would also be useful in `test_find_next_corner_case()`

porcupine/plugins/find.py
```diff
@@ -180,19 +180,20 @@ def get_match_ranges(self) -> list[tuple[str, str]]:
...
+        locations = [str(loc) for loc in self._textwidget.tag_ranges("sel")]
+        locations2 = [str(loc) for loc in self._textwidget.tag_ranges("find_highlight_selected")]
+        if (
+            locations == locations2
+            and len(locations) == 2
+            and tuple(locations) in self.get_match_ranges()
+        ):
+            replace_this_state = "normal"
+        else:
             replace_this_state = "disabled"
```
Akuli, line 196: A bit long but:
```suggestion
        replace_this_state = "normal" if (
            locations == locations2
            and len(locations) == 2
            and tuple(locations) in self.get_match_ranges()
        ) else "disabled"
```

porcupine/plugins/find.py
```diff
@@ -258,53 +259,49 @@ def highlight_all_matches(self, *junk: object) -> None:
...
-                self._select_match(start, end)
-                break
-        else:
-            # reached end of file, use the first match
-            self._select_match(*pairs[0])
-
-        self.statuslabel.config(text="")
-        self._update_buttons()
+        if pairs:
+            # If no matches highlighted, can highlight match exactly at cursor
```
Akuli, line 279:
```suggestion
            # If no matches highlighted yet, can highlight match exactly at cursor
```

PR #592, opened 2021-07-11
pygments typing fixes

Akuli:
Waiting for https://github.com/python/typeshed/pull/5759


porcupine/plugins/pygments_style.py
```diff
@@ -22,8 +22,9 @@ def get_colors(style_name: str) -> tuple[str, str]:
     style_infos = dict(iter(style))
 
     for tokentype in [token.String, token.Text]:
-        if style_infos[tokentype]["color"] is not None:
-            fg = "#" + style_infos[tokentype]["color"]
+        color = style_infos[tokentype]["color"]
+        if color is not None:
+            fg = "#" + color
```
Akuli, line 27: This code looks like it could be simplified with `or`

PR #595, opened 2021-07-11
Modernize typing

Akuli:
Get rid of List,Dict,Tuple,Union,Optional where possible. They are deprecated in 3.10, and we can already use the lower-case versions with `from __future__ import annotations` in most places.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,7 +81,7 @@ jobs:
       with:
         python-version: "3.9"
     - run: pip install -r requirements-dev.txt
-    - run: python3 -m black --check scripts/*.py porcupine tests more_plugins docs/*.py
+    - run: python3 -m black --check scripts/*.py docs/*.py porcupine tests more_plugins
   isort:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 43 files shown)


.github/workflows/main.yml
```diff
@@ -99,7 +99,7 @@ jobs:
       with:
         python-version: "3.9"
     - run: pip install -r requirements-dev.txt
-    - run: python3 -m pyflakes scripts/*.py porcupine tests more_plugins
+    - run: python3 -m pyflakes scripts/*.py porcupine tests more_plugins docs/extensions.py
```
Akuli, line 102:
```suggestion
    - run: python3 -m pyflakes scripts/*.py porcupine tests more_plugins docs/*.py
```

porcupine/utils.py
```diff
@@ -428,6 +415,12 @@ def handle_event(event: utils.EventWithData) -> None:
         ...
         foos = [Foo('ab', 123), Foo('cd', 456)]
         some_widget.event_generate('<<Thingy>>', data=Bar(foos))
+
+    Note that before Python 3.10, you need ``List[str]`` instead of
+    ``list[str]``, even if you use ``from __future__ import annotations``. This
+    is because Porcupine uses a library that needs to evaluate the type
+    annotations regardless of whether ``from __future__ import annotations``
```
Akuli, line 422:
```suggestion
    annotations even if ``from __future__ import annotations``
```

PR #622, opened 2021-07-24
do not delete newlines when dedenting

### LINKED ISSUE #609: shift+tab deletes blank lines
Seems to happen only if your file is indented with tabs, not spaces, and when some code containing blank lines is selected

Akuli:
Fixes #609

### DIFF
--- a/porcupine/textutils.py
+++ b/porcupine/textutils.py
@@ -668,7 +668,7 @@ class MainText(tkinter.Text):
                 # not deleting from start of line, delete previous char instead
                 location = f"{location} - 1 char"
 
-            if self.get(location).isspace():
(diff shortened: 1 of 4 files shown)


porcupine/textutils.py
```diff
@@ -668,7 +668,7 @@ def dedent(self, location: str) -> bool:
                 # not deleting from start of line, delete previous char instead
                 location = f"{location} - 1 char"
 
-            if self.get(location).isspace():
+            if self.get(location).isspace() and self.get(location) != "\n":
```
Akuli, line 671:
```suggestion
            # Deleting spaces this way is annoying. But that is somewhat
            # intentional because mixing tabs and spaces is bad.
            if self.get(location) in {" ", "\t"}:
```

PR #623, opened 2021-07-24
directory tree: use settings.add_option earlier

Akuli:
Prevents getting an error when loading directory tree from plugin manager

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -542,9 +542,8 @@ def setup() -> None:
         label="Focus directory tree", command=partial(focus_treeview, tree)
     )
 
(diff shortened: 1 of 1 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -540,7 +542,6 @@ def setup() -> None:
         label="Focus directory tree", command=partial(focus_treeview, tree)
     )
 
-    settings.add_option("directory_tree_projects", [], List[str])
     string_paths = settings.get("directory_tree_projects", List[str])
 
     # Must reverse because last added project goes first
```
Akuli, line 547:
```suggestion
    # Must reverse because last added project goes first
    string_paths = settings.get("directory_tree_projects", List[str])
```

PR #646, opened 2021-08-11
fix closing autocompleter when clicked elsewhere

### LINKED ISSUE #615: clicking next to autocomplete popup doesn't make the popup go away

Akuli:
Fixes #615


Akuli:
On Linux, it is now possible to paste text with middle-click. That won't stop completing.

PR #657, opened 2021-08-15
error if msvcrt not installed

### LINKED ISSUE #656: check if msvcrt installed
Python doesn't run without msvcrt, but does the installer run without msvcrt? It should tell the user to download msvcrt if needed.

Akuli:
Fixes #656

### DIFF
--- a/scripts/build-exe-installer.py
+++ b/scripts/build-exe-installer.py
@@ -41,7 +41,7 @@ print(url)
 
 response = requests.get(url)
 response.raise_for_status()
-zipfile.ZipFile(io.BytesIO(response.content)).extractall("build/Python")
+zipfile.ZipFile(io.BytesIO(response.content)).extractall("build/python-first")
 
 print("Copying files")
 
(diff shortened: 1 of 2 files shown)


scripts/build-exe-installer.py
```diff
@@ -52,16 +52,24 @@
...
 # We have a custom Porcupine.exe launcher which has the custom icon and can be called
 # with no arguments to launch Porcupine.
-# I couldn't get python to import from anywhere else than from Python directory, no separate pynsist_pkgs
+#
+# I couldn't get python to import from anywhere else than from Python directory,
+# so no separate pynsist_pkgs.
+#
+# When installing, python and python-libs get merged together, but a minimal
+# python is needed to fail setup early if it can't run.
+os.mkdir("build/python-second")
```
Akuli, line 66: mkdir is separate from rest, could split up big comment block in half

Akuli:
If you have Windows 7 without KB2533623 installed (which should be quite rare nowadays), Porcupine will end up in a weird state where it is installed but silently fails to start. I don't think I will ever fix this, because nobody else cares about Windows 7 with missing updates.

PR #660, opened 2021-08-16
windows installer: associate known file types

### LINKED ISSUE #637: Associate python files (and possibly other codes) with porcupine
Arrinao tried to do this and gave up

```
<Arrinao> where's the freaking porcupine executable??
<Arrinao> there's none over here
<Arrinao> C:\Users\Martin\AppData\Local\Programs\Porcupine
<Akuli> why do you need to know?
<Arrinao> I want to set up a registry entry to make porcupine open .py files
<Akuli> you can right-click porcupine in start menu
<Akuli> that's a good idea though, please create an issue about it :)
<Akuli> maybe porcupine should do it by default in windows
<Akuli> if you right-click and choose properties, it should show
<Arrinao> I can, but that doesn't point me at an executable but at a shortcut
<Akuli> so where does the shortcut point? look at its properties
<Arrinao> it points into the directory I've just posted :D
<Arrinao> but there's no executable
<Arrinao> C:\Users\Martin\AppData\Local\Programs\Porcupine\Python\pythonw.exe "C:\Users\Martin\AppData\Local\Programs\Porcupine\Porcupine.launch.pyw"
<Akuli> so... did you read the path?
<Arrinao> I've read only the first part, it didn' display as a whole until I copy pasted here
<Akuli> lol
<Arrinao> phew
<Arrinao> so what now?
<Akuli> to open a file, you need to call ...\Porcupine\Python\pythonw.exe with two arguments. ...\Porcupine\Porcupine.launch.pyw and the file you want to edit
<Akuli> i have no idea how you teach that to windows registry :)
<Arrinao> I don't even think I can
<Arrinao> whatever
<Akuli> i'll make the issue for you
```

Akuli:
Fixes #637

### DIFF
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -47,7 +47,7 @@ Var extension
   StrCpy $extension ""
   StrCpy $0 0  ; $0 = start index of extension
   ${Do}
-    StrCpy $1 "${FOO}" 1 $0   ; $1 = next char of extension
+    StrCpy $1 "${EXTENSIONS}" 1 $0   ; $1 = next char of extension
 
     ${If} $1 == ","
     ${OrIf} $1 == ""
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -1,5 +1,7 @@
 ; Based on a file that pynsist generated
 
+!include 'LogicLib.nsh'  ; TODO: use this to clean up gotos
+
 SetCompressor lzma
 
 Unicode true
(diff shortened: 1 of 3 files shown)


scripts/installer.nsi
```diff
@@ -57,8 +82,9 @@ Section "!Porcupine" sec_app
   DetailPrint "Running a sanity-check with the extracted Python..."
   ClearErrors
   ExecWait '"$INSTDIR\Python\pythonw.exe" -c pass'
-  IfErrors 0 +2
+  ${If} ${Errors}
     Abort
+  ${EndIf}
```
Akuli, line 87: I'm pretty sure this works, but should test

scripts/installer.nsi
```diff
@@ -39,6 +41,29 @@ Section -SETTINGS
...
+  StrCpy $extension ""
+  StrCpy $0 0  ; $0 = index into EXTENSIONS
+  ${Do}
+    StrCpy $1 "${EXTENSIONS}" 1 $0   ; $1 = next char of extension
+
+    ${If} $1 == ","
+    ${OrIf} $1 == ""
+      ; End of extension, run loop body
+      !macroend
+      !macro EndExtensionLoop
```
Akuli, line 56: indentation too clever

```suggestion
!macroend
!macro EndExtensionLoop
```

PR #661, opened 2021-08-16
speed up windows build by avoiding unnecessary pip install

Akuli:
Windows build is currently the slowest thing in CI.


Akuli:
Not much more than 30sec of speedup, does not make a big difference unfortunately

PR #664, opened 2021-08-16
geometry plugin: do not place window outside screen

### LINKED ISSUE #596: window can go off screen
1. Open Porcupine
2. Put porcupine to right half of screen
3. Close Porcupine
4. Adjust screen resolution so that width is half of what it was previously
5. Open Porcupine

Akuli:
Fixes #596

### DIFF
--- a/porcupine/plugins/geometry.py
+++ b/porcupine/plugins/geometry.py
@@ -13,8 +13,13 @@ def geometry_is_within_screen(geometry: str) -> bool:
         # Does not specify location, size only
         return True
 
-    width, height, x,  y = map(int, match.groups())
-    return (0 <= x and x+width <= get_main_window().winfo_screenwidth() and 0 <= y and y <= get_main_window().winfo_screenheight())
+    width, height, x, y = map(int, match.groups())
+    return (
+        0 <= x
+        and x + width <= get_main_window().winfo_screenwidth()
+        and 0 <= y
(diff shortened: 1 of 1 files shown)


porcupine/plugins/geometry.py
```diff
@@ -1,17 +1,37 @@
...
 
+def geometry_is_within_screen(geometry: str) -> bool:
+    match = re.fullmatch(r"(.*)x(.*)\+(.*)\+(.*)", geometry)
+    if match is None:
+        # Does not specify location, size only
+        return True
+
+    width, height, x, y = map(int, match.groups())
+    return (
+        0 <= x
```
Akuli, line 18:
```suggestion
        x >= 0
```

porcupine/plugins/geometry.py
```diff
@@ -1,17 +1,37 @@
...
+        and x + width <= get_main_window().winfo_screenwidth()
+        and 0 <= y
```
Akuli, line 20:
```suggestion
        and y >= 0
```

porcupine/plugins/geometry.py
```diff
@@ -1,17 +1,37 @@
...
+        and y <= get_main_window().winfo_screenheight()
```
Akuli, line 21:
```suggestion
        and y + height <= get_main_window().winfo_screenheight()
```

PR #665, opened 2021-08-16
Do not use broken bit haxors to detect if Alt was held down

### LINKED ISSUE #663: auto indent broken
Seems to happen only ~with porcupine installed using the  windows installer~ on windows

Put to new file `def f():` and press enter with cursor at end of line

Akuli:
Fixes #663

### DIFF
--- a/porcupine/plugins/autoindent.py
+++ b/porcupine/plugins/autoindent.py
@@ -17,10 +17,6 @@ from porcupine import get_tab_manager, tabs
 setup_before = ["rstrip"]
 
 log = logging.getLogger(__name__)
-if sys.platform == 'win32':
-    ALT_BIT = 0x20000
-else:
-    ALT_BIT = 0b1000
 
 
 def leading_whitespace(string: str) -> str:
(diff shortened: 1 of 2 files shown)


porcupine/plugins/autoindent.py
```diff
@@ -4,6 +4,7 @@
 import dataclasses
 import logging
 import re
+import sys
```
Akuli, line 7:
```suggestion
```

porcupine/plugins/autoindent.py
```diff
@@ -88,9 +88,8 @@ def after_enter(tab: tabs.FileTab, alt_pressed: bool) -> None:
         tab.textwidget.dedent("insert")
 
 
-def on_enter_press(tab: tabs.FileTab, event: tkinter.Event[tkinter.Text]) -> None:
+def on_enter_press(tab: tabs.FileTab, alt_pressed: bool, event: tkinter.Event[tkinter.Text]) -> None:
     assert isinstance(event.state, int)
```
Akuli, line 92:
```suggestion
```

Akuli:
Tetris plugin does the same with control, but apparently that's more cross-platform and correct on both windows and linux. (Haven't tried tetris plugin on mac)

Issue #439, opened 2021-05-25
foo/install.sh and bar/install.sh


Akuli:
Suppose `foo/install.sh` and `bar/install.sh` exist. Open them in Porcupine. Both show up as `install.sh` at first, but they show up as `foo/install.sh` and `bar/install.sh` when you save either one of them.

They should always show as `foo/install.sh` and `bar/install.sh`.

Akuli:
can't reproduce now lol

Akuli:
If I have `foo1/bar/baz.py` and `foo2/baz.py`, then show up as `bar/baz.py` and `baz.py` which is a bit confusing too

PR #673, opened 2021-08-19
Fix groupby bug

Akuli:
Maybe this is what caused #439


Akuli:
It seems like `itertools.groupby` is just shit. It only recognizes adjacent items as similar, so you need to sort first. And if you're going to sort anyway, you might as well use a simple linear-time algorithm instead that puts them to lists in a dictionary.

PR #678, opened 2021-08-19
rename modified stuff and fix shit docstring

Akuli:


porcupine/tabs.py
```diff
@@ -619,12 +619,8 @@ def _set_saved_state(self, state: tuple[os.stat_result | None, int, str]) -> Non
...
         self._update_titles()
 
-    def is_modified(self) -> bool:
-        """Return False if the text has changed since previous save.
-
-        This is set to False automagically when the content is modified.
-        Use :meth:`mark_saved` to set this to True.
-        """
+    def has_unsaved_changes(self) -> bool:
+        """Return True if the text has changed since previous save."""
```
Akuli, line 623:
```suggestion
        """Return True if the text in the editor has changed since the previous save."""
```

Issue #627, opened 2021-07-28
"Wrap long lines" option unsets when dragging tab out of porcupine


Akuli:
Would be nice if it was remembered

PR #688, opened 2021-08-21
include settings in saved state

### LINKED ISSUE #687: encoding asked upon restart
1. Create non-utf8 file
2. Open it in Porcupine and enter the encoding when asked
3. Close Porcupine with Alt+F4
4. Open Porcupine

It remembers you had the file open, but does not remember the encoding it used and asks again.

Akuli:
Fixes #687 (would fix also #627 with a little bit more code)

### DIFF
--- a/porcupine/settings.py
+++ b/porcupine/settings.py
@@ -380,19 +380,16 @@ def save() -> None:
     Note that :func:`porcupine.run` always calls this before it returns,
     so usually you don't need to worry about calling this yourself.
     """
-    currently_known_defaults = {
-        name: opt.default for name, opt in _global_settings._options.items()
-    }
-
-    writing: dict[str, Any] = {
+    writing: dict[str, object] = {
         name: unknown.value for name, unknown in _global_settings._unknown_options.items()
     }
-    writing.update({name: get(name, object) for name in currently_known_defaults.keys()})
-
-    # don't store anything that doesn't differ from defaults
-    for name, default in currently_known_defaults.items():
-        if name in writing and writing[name] == default:
-            del writing[name]
+    writing.update(
+        {
+            name: get(name, object)
+            for name, option in _global_settings._options.items()
+            if get(name, object) != option.default
+        }
+    )
 
     with get_json_path().open("w", encoding="utf-8") as file:
         json.dump({name: _value_to_save(value) for name, value in writing.items()}, file)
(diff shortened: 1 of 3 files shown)


porcupine/settings.py
```diff
@@ -201,12 +219,18 @@ def set(self, option_name: str, value: object, *, from_config: bool = False) ->
...
         if option_name not in self._options and from_config:
-            self._unknown_options[option_name] = value
+            self._unknown_options[option_name] = _UnknownOption(
+                value, converter_called=already_converted
+            )
             return
 
         option = self._options[option_name]
-        if from_config:
+        if from_config and not already_converted:
```
Akuli, line 233: Should refactor it so that this line says `if not already_converted`. The `already_converted` boolean only makes sense with `from_config` should change so it is always good

porcupine/settings.py
```diff
@@ -336,9 +381,9 @@ def save() -> None:
         name: opt.default for name, opt in _global_settings._options.items()
     }
 
-    writing: dict[str, Any] = _global_settings._unknown_options.copy()
-
-    # don't wipe unknown stuff from config file
+    writing: dict[str, Any] = {
+        name: unknown.value for name, unknown in _global_settings._unknown_options.items()
```
Akuli, line 385: This should be cleaned up

porcupine/tabs.py
```diff
@@ -949,6 +956,7 @@ def from_state(
         assert isinstance(state, _FileTabState)  # not namedtuple in older porcupines
 
         tab = cls(manager, content=(state.content or ""), path=state.path)
+        tab.settings.set_state(state.settings_state)
```
Akuli, line 959:
```suggestion
        tab.settings.set_state(state.settings_state)  # must be before reload()
```

porcupine/settings.py
```diff
@@ -292,10 +321,28 @@ def debug_dump(self) -> None:
...
+            print(string)
         print()
 
+    # TODO: document state methods?
+    def get_state(self) -> dict[str, _UnknownOption]:
+        result = self._unknown_options.copy()
+        result.update(
+            {
+                name: _UnknownOption(option.value, call_converter=False)
+                for name, option in self._options.items()
```
Akuli, line 337: Maybe this should skip defaults, and fwiw just be same logic as in `save()`

PR #689, opened 2021-08-21
delete settings debug_dump() method

Akuli:
I never actually used it.


Akuli:
Added back, was useful

PR #692, opened 2021-08-21
rewrite long line marker logic

### LINKED ISSUE #157: too long line marker can be in wrong place if there are very long lines
hard to reproduce

closing and reopening the file fixes the problem

Akuli:
Fixes #157 

The problem was that the location of the character at top left corner was used to calculate stuff, and when scrolled right, it might not be visible on the screen when there are blank lines (and it is "rounded" to the beginning of the line). This was hard to reproduce because the result was cached in `self._width`, apparently for no good reason.

### DIFF
--- a/porcupine/plugins/longlinemarker.py
+++ b/porcupine/plugins/longlinemarker.py
@@ -43,23 +43,20 @@ class LongLineMarker:
         font = tkfont.Font(name=self.tab.textwidget["font"], exists=True)
         marker_x = font.measure(" " * max_line_length)
 
+        # Placing, scrolling etc are relative to what's inside these paddings
         padding_on_each_side = (
             self.tab.textwidget["padx"]
             + self.tab.textwidget["borderwidth"]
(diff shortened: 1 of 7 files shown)


porcupine/plugins/longlinemarker.py
```diff
@@ -22,15 +22,13 @@ def __init__(self, filetab: tabs.FileTab) -> None:
         # this must not be a ttk frame because the background color
         # comes from the pygments style, not from the ttk theme
         self.frame = tkinter.Frame(filetab.textwidget, width=1)
-        self._width = self._height = 1  # on_configure() will run soon
 
     def setup(self) -> None:
         utils.add_scroll_command(self.tab.textwidget, "xscrollcommand", self.do_update)
```
Akuli, line 27:
```suggestion
        # xscrollcommand runs when the text widget resizes
        utils.add_scroll_command(self.tab.textwidget, "xscrollcommand", self.do_update)
```

PR #695, opened 2021-08-21
github actions: add timeouts to tests

Akuli:
Not only for #379 but also because I sometimes make tests frozen by messing up with the tests. For example, if I forget to mock a function that shows a dialog to the user.


Akuli:
This just happened in #694 and CI was stuck waiting for a non-existing user to click a button for about an hour. It didn't run newer jobs that I actually care about, because each repository has a limited number of running available at a time or something (which makes sense).

PR #696, opened 2021-08-23
add debug prints to debug #379

Akuli:


porcupine/__main__.py
```diff
@@ -159,6 +160,18 @@ def main() -> None:
...
+            t for t in threading.enumerate() if t != threading.current_thread() and not t.daemon
+        ]
+        if non_daemon_threads:
+            print("Non-daemon threads running while Porcupine exits:")
+            for thread in non_daemon_threads:
+                if getattr(thread.run, "__self__", None) is thread:
+                    print(thread.name, thread._target, thread._args, thread._kwargs)  # type: ignore
+                else:
+                    # run method overrided
+                    print(thread.name, thread.run)
```
Akuli, line 174:
```suggestion
                    print(thread.name, thread.run, flush=True)
```

porcupine/__main__.py
```diff
@@ -159,6 +160,18 @@ def main() -> None:
...
         settings.save()
+
+        non_daemon_threads = [
+            t for t in threading.enumerate() if t != threading.current_thread() and not t.daemon
+        ]
+        if non_daemon_threads:
+            print("Non-daemon threads running while Porcupine exits:")
+            for thread in non_daemon_threads:
+                if getattr(thread.run, "__self__", None) is thread:
+                    print(thread.name, thread._target, thread._args, thread._kwargs)  # type: ignore
```
Akuli, line 171:
```suggestion
                    print(thread.name, thread._target, thread._args, thread._kwargs, flush=True)  # type: ignore
```

porcupine/__main__.py
```diff
@@ -159,6 +160,18 @@ def main() -> None:
         get_main_window().mainloop()
     finally:
         settings.save()
+
+        non_daemon_threads = [
+            t for t in threading.enumerate() if t != threading.current_thread() and not t.daemon
+        ]
+        if non_daemon_threads:
+            print("Non-daemon threads running while Porcupine exits:")
```
Akuli, line 168:
```suggestion
            print("Non-daemon threads running while Porcupine exits:", flush=True)
```

Akuli (commented):
Add `flush=True` so that this will not be lost in github actions weirdness, no matter what

PR #698, opened 2021-08-23
Create panedwindow to control minimap size

Akuli:
Fixes #60

PR #700, opened 2021-08-23
Improve *args and **kwargs type hints

Akuli:
Contains some of the same changes as #698

### DIFF
--- a/porcupine/plugins/pastebin.py
+++ b/porcupine/plugins/pastebin.py
@@ -146,7 +146,7 @@ class DPaste(Paste):
 class SuccessDialog(tkinter.Toplevel):
     def __init__(self, url: str):
         super().__init__()
-        self.url = url
+        self._url = url
 
         content = ttk.Frame(self)
         content.pack(fill="both", expand=True)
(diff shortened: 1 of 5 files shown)


porcupine/plugins/pastebin.py
```diff
@@ -144,8 +144,8 @@ def run(self, code: str, lexer_class: LexerMeta) -> str:
 
 
 class SuccessDialog(tkinter.Toplevel):
-    def __init__(self, url: str, *args: Any, **kwargs: Any):
-        super().__init__(*args, **kwargs)
+    def __init__(self, url: str):
+        super().__init__()
         self.url = url
```
Akuli, line 149: could this be `self._url`

porcupine/utils.py
```diff
@@ -84,6 +84,28 @@ def quote(string: str) -> str:
     quote = shlex.quote
 
 
+# https://github.com/python/typing/issues/769
+def copy_type(f: _T) -> Callable[[Any], _T]:
```
Akuli, line 88: Not referenced in `docs/utils.rst`?

PR #704, opened 2021-08-23
add json settings file indentation

rdbende:
make json settings file readable by adding indentation


Akuli (approved):
Thanks!

Akuli:
If you don't want to get old commits showing up in every pull request, you can delete the branch that Git on your computer currently calls "master" (`git branch -D master`, after switching to some other non-master branch) and then make a new master branch based on mine (`git checkout --track Akuli/master`). To see what is going on, I recommend [git lola](https://akuli.github.io/git-guide/branches.html#git-lola) if you don't already have something similar.

rdbende:
Thanks for the advice!

PR #708, opened 2021-08-23
encoding dropdown

### LINKED ISSUE #705: encoding dropdown
would be nice to have a combobox with the available codecs in the popup window instead of the entry

rdbende:
btw fixes #705

### DIFF
--- a/porcupine/tabs.py
+++ b/porcupine/tabs.py
@@ -12,6 +12,7 @@ import os
 import pathlib
 import tkinter
 import traceback
+import encodings
 from tkinter import filedialog, messagebox, ttk
 from typing import Any, Callable, Iterable, NamedTuple, Optional, Sequence, Type, TypeVar
 
@@ -413,6 +414,16 @@ class ReloadInfo(utils.EventDataclass):
 
 
 def _ask_encoding(path: pathlib.Path, encoding_that_didnt_work: str) -> str | None:
+    list_of_encodings = [x.replace("_", "-") for x in set(encodings.aliases.aliases.values())]
+
+    # must do this, because in list_of_encodings there are encodings, that codecs don't know
+    # but probably this isn't the appropriate place to do it
+    for encoding in list_of_encodings:
+        try:
+            codecs.lookup(encoding)
+        except LookupError:
+            list_of_encodings.remove(encoding)
+
     label_width = 400
 
     dialog = tkinter.Toplevel()
@@ -432,9 +443,9 @@ def _ask_encoding(path: pathlib.Path, encoding_that_didnt_work: str) -> str | No
     ).pack(fill="x", padx=10, pady=10)
 
     var = tkinter.StringVar()
-    entry = ttk.Entry(big_frame, textvariable=var)
-    entry.pack(pady=50)
-    entry.insert(0, encoding_that_didnt_work)  # type: ignore[no-untyped-call]
+    combobox = ttk.Combobox(big_frame, values=sorted(list_of_encodings), textvariable=var)
+    combobox.pack(pady=40)
+    combobox.set(encoding_that_didnt_work)  # type: ignore[no-untyped-call]
 
     ttk.Label(
         big_frame,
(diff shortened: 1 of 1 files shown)


porcupine/tabs.py
```diff
@@ -432,9 +443,9 @@ def _ask_encoding(path: pathlib.Path, encoding_that_didnt_work: str) -> str | No
     ).pack(fill="x", padx=10, pady=10)
 
     var = tkinter.StringVar()
-    entry = ttk.Entry(big_frame, textvariable=var)
-    entry.pack(pady=50)
-    entry.insert(0, encoding_that_didnt_work)  # type: ignore[no-untyped-call]
+    combobox = ttk.Combobox(big_frame, values=sorted(list_of_encodings), textvariable=var)
```
Akuli, line 446: Maybe the whole `textvariable` should be deleted, since this uses `combobox.get()` anyway.

porcupine/tabs.py
```diff
@@ -413,6 +414,16 @@ class ReloadInfo(utils.EventDataclass):
 
 
 def _ask_encoding(path: pathlib.Path, encoding_that_didnt_work: str) -> str | None:
+    list_of_encodings = [x.replace("_", "-") for x in set(encodings.aliases.aliases.values())]
+
+    # must do this, because in list_of_encodings there are encodings, that codecs don't know
+    # but probably this isn't the appropriate place to do it
+    for encoding in list_of_encodings:
```
Akuli, line 421: This loop doesn't work. For a simpler example, consider:

```
>>> foo = ['a', 'b', 'c']
>>> for item in foo: foo.remove(item)
... 
>>> foo
['b']
```

The for loop internally creates a list iterator, which has an  index pointing into the list. Let's say that the index is `0`, and so we delete `foo[0]`, making `foo` to be `['b', 'c']`. So far so good. But then the index increments for the next iteration, and now we are deleting `foo[1]`, which is now `'c'` because `foo` is `['b', 'c']`. In general, loops like this can't delete two adjacent elements, because once an element is deleted, the index incrementing will cause the next element to be skipped.

To fix this, you can loop backwards, since deleting an element does not mess up indexes before whatever it is you deleted:

```suggestion
    for encoding in reversed(list_of_encodings):
```

Or you can loop over a copy of the list:

```suggestion
    for encoding in list_of_encodings.copy():
```

porcupine/tabs.py
```diff
@@ -631,15 +641,25 @@ def __init__(
...
-            self, width=1, height=1, wrap="none", undo=True, padx=3
+            self,
+            width=1,
+            height=1,
+            wrap="none",
+            undo=True,
+            padx=3,
+            borderwidth=0,
+            relief="solid",
+            highlightthickness=0,
```
Akuli, line 662: Some of these changes should be reflected in the minimap plugin. Currently it does not line up nicely with the main text widget:

Screenshot at 2021-08-24 15-30-56

What is the reason for changing all this in the first place? What was wrong with the default values?

porcupine/tabs.py
Akuli: Looking into this more, it seems like the right solution would be to use a hard-coded list: https://stackoverflow.com/a/25584253

Currently Python 3.7 is the oldest Python that Porcupine supports. If you add a comment that mentions `3.7`, the encoding list will get updated when Porcupine drops support for Python 3.7.

porcupine/tabs.py
rdbende: I kept the `textvariable` so that if the user types an unknown encoding, it checks and disables the Ok button.

porcupine/tabs.py
rdbende: lol, it wasn't intended, I just played around with the border, because I don't like the `highlightthickness`, but I didn't want to commit it

porcupine/tabs.py
```diff
@@ -412,6 +412,109 @@ class ReloadInfo(utils.EventDataclass):
     had_unsaved_changes: bool
 
 
+# list of encodings supported by python 3.7
```
Akuli, line 415:
```suggestion
# list of encodings supported by python 3.7 https://stackoverflow.com/a/25584253
```

porcupine/tabs.py
Akuli: You might like `git diff --cached` (or `--staged`) before committing so you actually see what you are about to commit. To include only a part of a file for committing, you can use `git add -p file.py`, and it will ask about each change whether you want it to be in the commit or not.

rdbende: The funny thing is that I did `git diff --staged` but didn't notice

PR #712, opened 2021-08-24
add dialog title

### LINKED ISSUE #711: encoding dialog: title is stupid
Plasma puts a number after the title
image

should be something like "select an encoding"

rdbende:
Fixes #711


Akuli:
I ran `git grep 'tkinter.Toplevel[()]'`, and apparently the plugin manager has the same problem. Can you fix that too?

PR #722, opened 2021-09-02
isort plugin

rdbende:
I needed an isort plugin, so this is just black with ctrl-f and replace


Akuli:
Thanks! It would be nice to combine the two into one plugin, maybe `python_tools.py`, but that can be a separate PR.

PR #724, opened 2021-09-03
rewrite wrap plugin to use settings

### LINKED ISSUE #627: "Wrap long lines" option unsets when dragging tab out of porcupine

Akuli:
Fixes #627

### DIFF
--- a/porcupine/plugins/wrap.py
+++ b/porcupine/plugins/wrap.py
@@ -6,11 +6,11 @@ from functools import partial
 
 from porcupine import get_tab_manager, menubar, tabs
 
-
 # unfortunately wrap info be synced between 3 places:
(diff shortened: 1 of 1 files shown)


porcupine/plugins/wrap.py
```diff
@@ -7,22 +7,46 @@
...
-        return True
-    return False
+        try:
+            value = tab.settings.get("wrap", bool)
+        except KeyError:
+            # Happens when adding new tab to tab manager before on_new_filetab() runs
+            return
+        var.set(value)
+    else:
+        var.set(False)
```
Akuli, line 31:
```suggestion
        var.set(False)  # menu item will be disabled
```

PR #725, opened 2021-09-03
combine black and isort plugins

Akuli:
Cleans up copy/pasta created in #722


porcupine/plugins/python_tools.py
```diff
@@ -19,7 +19,7 @@
 log = logging.getLogger(__name__)
 
 
-def run_black(code: str, path: Path | None) -> str:
+def run_tool(tool: str, code: str, path: Path | None) -> str:
     python = python_venv.find_python(None if path is None else utils.find_project_root(path))
     if python is None:
         messagebox.showerror(
```
Akuli, line 25: The line below (which I can't comment on github) says black, even if the user tries to isort.

PR #731, opened 2021-09-03
update line numbers when unfolding

### LINKED ISSUE #709: folding: line numbers don't appear when unfolding in small files
line numbers don't appear when unfolding in short files

folded:
image

unfolded:
image

the code is there, but the line numbers are not

Akuli:
Fixes #709

### DIFF
--- a/porcupine/plugins/fold.py
+++ b/porcupine/plugins/fold.py
@@ -47,11 +47,6 @@ def update_line_numbers(tab: tabs.FileTab) -> None:
             child.do_update()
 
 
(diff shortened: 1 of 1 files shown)


porcupine/plugins/fold.py
```diff
@@ -41,6 +41,17 @@ def find_indented_block(tab: tabs.FileTab, lineno: int) -> int | None:
     return last_lineno
 
 
+def update_line_numbers(tab: tabs.FileTab):
```
Akuli, line 44:
```suggestion
def update_line_numbers(tab: tabs.FileTab) -> None:
```

PR #733, opened 2021-09-04
Directory tree: right-click options to delete or trash files and folders

Akuli:


requirements.txt
```diff
@@ -11,3 +11,4 @@ typing_extensions
 dacite>=1.5.1,<2.0.0
 toml
 ttkthemes>=3.0.0
+send2trash[nativeLib]>=1.8.0,<2.0.0
```
Akuli, line 14: need to test windows installer

porcupine/plugins/directory_tree_delete.py
```diff
@@ -0,0 +1,99 @@
...
+    if not close_tabs(find_tabs_by_parent_path(path)):
+        return
+
+    try:
+        send2trash(path)
+    except Exception as e:
+        log.exception(f"can't trash {path}")
+        messagebox.showerror(
+            f"Moving to {trash_name} failed",
+            f"Moving {path} to {trash_name} failed:\n\n{type(e).__name__}: {e}",
```
Akuli, line 54:
```suggestion
            f"Moving {path} to {trash_name} failed.\n\n{type(e).__name__}: {e}",
```

porcupine/plugins/python_venv.py
```diff
@@ -109,25 +109,24 @@ def _on_folder_refreshed(event: utils.EventWithData) -> None:
...
+    if path == project_root or not path.is_dir():
+        return
+
+    is_used_var = tkinter.BooleanVar(value=(get_venv(project_root) == path))
+    cast(Any, tree.contextmenu).garbage_collection_is_lol = is_used_var
+
+    tree.contextmenu.add_checkbutton(
+        label="Use this Python venv",
+        variable=is_used_var,
+        state=("normal" if is_venv(path) else "disabled"),
```
Akuli, line 126: Should be also disabled whenever it is checked. Otherwise user thinks that clicking it will somehow stop using the venv, and that's just not true.

porcupine/plugins/directory_tree.py
```diff
@@ -546,3 +562,6 @@ def setup() -> None:
         if path.is_absolute() and path.is_dir():
             tree.add_project(path, refresh=False)
     tree.refresh()
+
+    # TODO: mac right click = button 2?
```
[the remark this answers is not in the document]
Akuli: #737

PR #734, opened 2021-09-04
Move git stuff from directory_tree plugin into a separate plugin

Akuli:


porcupine/plugins/directory_tree.py
```diff
@@ -342,35 +238,9 @@ def find_project_id(self, item_id: str) -> str:
     # The following two methods call each other recursively.
 
     def _update_tags_and_content(self, project_root: Path, child_id: str) -> None:
```
Akuli, line 240: This function is now very short. Is it still needed?

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+            result[path] = "git_ignored"
+        elif line[:2] in {"AA", "UU"}:
+            result[path] = "git_mergeconflict"
+        else:
+            log.warning(f"unknown git status line: {repr(line)}")
+    return result
+
+
+# TODO: copy/pasted from directory_tree plugin
+def _path_to_root_inclusive(path: Path, root: Path) -> Iterator[Path]:
```
Akuli, line 70: Used in only one function, consider rewriting that

In directory_tree.py it is also used in exactly one place and that might benefit from some rewriting too

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+                    status = "git_modified"
+                elif "git_added" in substatuses:
+                    status = "git_added"
+                else:
+                    assert not substatuses
+                    status = None
+
+            # TODO: allow other plugins to create tags too
+            # Ideally without making two tcl calls.
+            # If I remember correctly this code is a bit perf critical
```
Akuli, line 138:
```suggestion
            # If I remember correctly this code was perf critical, maybe isn't anymore after #734
```

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+    def __init__(self, tree: DirectoryTree):
+        self.tree = tree
+        self.project_specific_colorers: dict[str, ProjectColorer] = {}
+
+    def config_color_tags(self, junk: object = None) -> None:
+        fg = self.tree.tk.eval("ttk::style lookup Treeview -foreground")
+        bg = self.tree.tk.eval("ttk::style lookup Treeview -background")
+        gray = utils.mix_colors(fg, bg, 0.5)
+
+        if sum(self.tree.winfo_rgb(fg)) > 3 * 0x7FFF:
```
Akuli, line 161: aboutdialog has similar code, should make util

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+    def start_status_coloring_for_all_projects(self, junk_event: object) -> None:
+        old_project_ids = set(self.project_specific_colorers.keys())
+        new_project_ids = set(self.tree.get_children())
+        for removed in old_project_ids - new_project_ids:
+            del self.project_specific_colorers[removed]
+        for added in new_project_ids - old_project_ids:
+            self.project_specific_colorers[added] = ProjectColorer(self.tree, added)
+
+        for colorer in self.project_specific_colorers.values():
+            colorer.start_running_git_status()
```
Akuli, line 184: I'm not sure if this makes sense. It can make old ProjectColorers to start running `git status` even if it's already running. I think a better solution would be to stop all old coloring and start anew.

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def update_git_tags_for_item(self, item_id: str) -> None:
```
Akuli, line 186: Maybe rename to `color_item`?

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+                    status = None
+
+            # TODO: allow other plugins to create tags too
+            # Ideally without making two tcl calls.
+            # If I remember correctly this code is a bit perf critical
+            self.tree.item(item_id, tags=([] if status is None else status))
+            if item_id in selection:
+                update_tree_selection_color(self.tree)
+
+    def update_colors_now_or_later(self, item_id: str) -> None:
```
Akuli, line 143: Rename to `update_color_tags_now_or_later`?

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+        path = path.parent
+
+
+class ProjectColorer:
+    def __init__(self, tree: DirectoryTree, project_id: str):
+        self.tree = tree
+        self.project_id = project_id
+        self.project_path = get_path(project_id)
+        self.git_status_future: Future[dict[Path, str]] | None = None
+        self.coloring_queue: list[str] = []
```
Akuli, line 85: Maybe this should be a set?

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+        "git_mergeconflict",
+        None,
+        "git_untracked",
+        "git_ignored",
+    ].index(git_tag)
+
+
+def setup() -> None:
+    for tree in utils.get_children_recursively(get_paned_window()):
+        if isinstance(tree, DirectoryTree):
```
Akuli, line 231: This logic appears in many places now, including tests. Maybe there should be a function that finds and returns the tree from the currently running Porcupine?

porcupine/plugins/git_status.py
```diff
@@ -0,0 +1,250 @@
...
+            tree.config(style="DirectoryTreeGitStatus.Treeview")
+
+            main_colorer = TreeColorer(tree)
+            tree.bind(
+                "<<RefreshBegins>>", main_colorer.start_status_coloring_for_all_projects, add=True
+            )
+            utils.bind_with_data(
+                tree,
+                "<<UpdateItemTags>>",
+                lambda e: main_colorer.update_git_tags_for_item(e.data_string),
```
Akuli, line 241: e --> event

Akuli:
This seems to make everything a lot less fragile, and also more responsive because the whole refreshing no longer waits until `git status` has been ran.

porcupine/plugins/directory_tree.py
Akuli: ye

porcupine/plugins/directory_tree.py
```diff
@@ -342,35 +226,9 @@ def find_project_id(self, item_id: str) -> str:
...
-                status = "git_added"
-            else:
-                assert not substatuses
-                status = None
+        self.event_generate("<<UpdateItemTags>>", data=child_id)
 
-        self.item(child_id, tags=([] if status is None else status))
+        child_path = get_path(child_id)
         if child_id.startswith(("dir:", "project:")) and not self.contains_dummy(child_id):
             self._open_and_refresh_directory(child_path, child_id)
```
Akuli, line 233: Maybe `_open_and_refresh_directory` shouldn't take a separate `child_path` argument?

porcupine/plugins/directory_tree.py
```diff
@@ -410,7 +268,12 @@ def _open_and_refresh_directory(self, dir_path: Path, dir_id: str) -> None:
         for child_path, child_id in path2id.items():
             self._update_tags_and_content(project_root, child_id)
 
-        for index, child_id in enumerate(sorted(self.get_children(dir_id), key=self._sorting_key)):
+        for index, child_id in enumerate(
+            sorted(
+                self.get_children(dir_id),
+                key=(lambda item_id: [f(item_id) for f in self.sorting_keys]),
```
Akuli, line 274: Output of sorting keys may depend on git tags which are set afterwards. Can this cause wrong order bug?

Issue #30, opened 2020-07-05
better git integration


Akuli:
at the very least, auto-save files as the user types, and auto-reload from disk when the files change

Akuli:
I'm starting to think that auto-reload is a bad thing, at least if it can't be undone. I lost some changes with git, but I still had them in Porcupine and I just saved the file to restore them. If I ever add auto-reload to Porcupine, it should be undoable because of this.

PR #739, opened 2021-09-05
directory tree: Git options in right click menu

Akuli:
Really improves #30

### DIFF
--- a/porcupine/plugins/git_right_click.py
+++ b/porcupine/plugins/git_right_click.py
@@ -45,8 +45,7 @@ def populate_menu(event: tkinter.Event[DirectoryTree]) -> None:
         ["git", "status"], cwd=project_root, stdout=subprocess.DEVNULL, **utils.subprocess_kwargs
     )
     if run_result.returncode == 0:
-        if tree.contextmenu.index("end") is not None:  # type: ignore[no-untyped-call]
-            # menu not empty
+        if tree.contextmenu.index("end") is not None:  # menu not empty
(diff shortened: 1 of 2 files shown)


porcupine/plugins/git_right_click.py
```diff
@@ -0,0 +1,66 @@
...
+    "python_venv",
+]
+
+log = logging.getLogger(__name__)
+
+if sys.platform == "win32":
+    trash_name = "recycle bin"
+else:
+    trash_name = "trash"
+
```
Akuli, line 28: These don't belong here

Akuli (commented):
x

porcupine/plugins/git_right_click.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        tree.contextmenu.add_command(
+            label="git restore (discard non-added changes)",
+            command=partial(run_git, "checkout", "--", path),
+        )
+
+
+def setup() -> None:
+    for tree in utils.get_children_recursively(get_paned_window()):
+        if isinstance(tree, DirectoryTree):
```
Akuli, line 65: should be replaced by `get_directory_tree()`, defined in #734

Akuli:
I'm not sure if this is a good idea. There's still nothing to commit from the directory tree, for example.

Akuli:
I think I'll close the PR but not delete the branch yet. I might revisit this if I ever decide that Porcupine needs more git features.

PR #741, opened 2021-09-05
send2trash: smaller install

Akuli:
Works on windows and linux. Not sure about mac


Akuli:
Reduces windows installer size a lot: 31.4MB --> 22.8MB

PR #750, opened 2021-09-09
change to "from pathlib import Path"

### LINKED ISSUE #745: code cleanup: pathlib import
```
akuli@akuli-desktop:~/porcu$ git grep "from pathlib import Path" | wc -l
12
akuli@akuli-desktop:~/porcu$ git grep "import pathlib" | wc -l
23
```

I now prefer `from pathlib import Path`, would be good to consistently use it

Akuli:
fixes #745


porcupine/plugins/run/terminal.py
```diff
@@ -3,30 +3,30 @@
...
 
-_this_dir = pathlib.Path(__file__).absolute().parent
+_this_dir = Path(__file__).absolute().parent
 if sys.platform == "win32":
     run_script = _this_dir / "windows_run.py"
 else:
     run_script = _this_dir / "bash_run.sh"
 
-CommandArgument = Union[str, pathlib.Path]
+CommandArgument = Union[str, Path]
```
Akuli, line 26: not used
```suggestion
```

porcupine/settings.py
```diff
@@ -130,7 +130,7 @@ def add_option(
         calling :func:`set_` with a value that isn't a string or
         calling :func:`get` with the type set to something else than ``str``
         is an error. You can also provide a custom type with the *type*
-        argument, e.g. ``add_option('foo', None, Optional[pathlib.Path])``.
+        argument, e.g. ``add_option('foo', None, Optional[Path])``.
```
Akuli, line 133: For docs, I think `pathlib.Path` is better because then it is clear what it means

```suggestion
        argument, e.g. ``add_option('foo', None, Optional[pathlib.Path])``.
```

porcupine/settings.py
```diff
@@ -284,14 +284,14 @@ def get(self, option_name: str, type_: Any) -> Any:
             foo = settings.get('something', str)
             reveal_type(foo)  # str
 
-            shitty_bar = settings.get('something', Optional[pathlib.Path])
+            shitty_bar = settings.get('something', Optional[Path])
```
Akuli, line 287:
```suggestion
            from pathlib import Path
            shitty_bar = settings.get('something', Optional[Path])
```

porcupine/tabs.py
```diff
@@ -528,7 +528,7 @@ class FileTab(Tab):
         Things like :source:`the line number plugin <porcupine/plugins/linenumbers.py>` use this attribute.
 
     .. attribute:: path
-        :type: pathlib.Path | None
+        :type: Path | None
```
Akuli, line 531: I don't think the types display very well anyway, but let's not make it any worse than it is:

```suggestion
        :type: pathlib.Path | None
```

PR #753, opened 2021-09-09
Directory tree cleanups

Akuli:


porcupine/plugins/git_status.py
```diff
@@ -191,11 +192,12 @@ def start_status_coloring_for_all_projects(self, junk_event: object) -> None:
         for project_id in self.tree.get_children():
             colorer = ProjectColorer(self.tree, project_id)
             self.project_specific_colorers[project_id] = colorer
+            colorer.coloring_queue.add(project_id)
```
Akuli, line 195: variable names could be simplified

PR #755, opened 2021-09-09
create hide_project plugin

Akuli:


porcupine/plugins/directory_tree.py
```diff
@@ -150,18 +150,30 @@ def add_project(self, root_path: Path, *, refresh: bool = True) -> None:
...
+        return result
+
+    def _find_project_id_by_path(self, path: Path) -> str | None:
         matching_projects = [
             project_id for project_id in self.get_children() if get_path(project_id) in path.parents
         ]
         if not matching_projects:
+            return None
+
+        # When opening ~/foo/bar/lol.py, use ~/foo/bar instead of ~/foo
```
Akuli, line 166:
```suggestion
        # For ~/foo/bar/lol.py, use ~/foo/bar instead of ~/foo
```

porcupine/plugins/directory_tree.py
```diff
@@ -192,6 +205,15 @@ def contains_dummy(self, parent: str) -> bool:
         children = self.get_children(parent)
         return len(children) == 1 and self.tag_has("dummy", children[0])
 
+    def project_has_opened_tabs(self, project_id: str) -> bool:
```
Akuli, line 208: rename to `project_has_open_filetabs`

Akuli, line 208:
```suggestion
    # TODO: it's not great how only the directory tree knows this
    def project_has_opened_tabs(self, project_id: str) -> bool:
```

porcupine/plugins/hide_project.py
```diff
@@ -0,0 +1,29 @@
...
+
+import tkinter
+
+from porcupine.plugins.directory_tree import DirectoryTree, get_directory_tree
+
+setup_after = ["directory_tree"]
+
+
+def add_hide_option(event: tkinter.Event[DirectoryTree]) -> None:
+    tree: DirectoryTree = event.widget
```
Akuli, line 12:
```suggestion
    tree = event.widget
```

PR #762, opened 2021-09-19
use more types-foo packages

Akuli:


Akuli:
types-appdirs and types-toposort should be available tomorrow

PR #771, opened 2021-09-26
git_status plugin: fix slowness in projects with many ignored files

Akuli:
Porcupine was behaving really slowly, like 5sec of unresponsiveness after clicking it. It only happened after opening a project named `~/tex`. It contains many ignored files, and looping through them for each directory tree item just isn't practical:
```
akuli@akuli-desktop:~/tex$ git status --porcelain | wc -l
1
akuli@akuli-desktop:~/tex$ git status --porcelain --ignored | wc -l
490
```

### DIFF
--- a/porcupine/plugins/git_status.py
+++ b/porcupine/plugins/git_status.py
@@ -70,6 +70,27 @@ def run_git_status(project_root: Path) -> dict[Path, str]:
             result[path] = "git_mergeconflict"
         else:
             log.warning(f"unknown git status line: {repr(line)}")
+
+    # When a folder contains files with different statuses, decide status of folder
+    # There can be lots of statuses, not good to loop through them in gui thread
+    folder_to_content_statuses: dict[Path, set[str]] = {}
+    for path, status in result.items():
+        if status in {"git_added", "git_modified", "git_mergeconflict"}:
+            for folder in path.parents:
+                folder_to_content_statuses.setdefault(folder, set()).add(status)
+
+    assert not (folder_to_content_statuses.keys() & result.keys())
+
+    for folder, content_statuses in folder_to_content_statuses.items():
+        if "git_mergeconflict" in content_statuses:
+            result[folder] = "git_mergeconflict"
+        elif "git_modified" in content_statuses:
+            result[folder] = "git_modified"
+        elif "git_added" in content_statuses:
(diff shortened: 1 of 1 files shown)


porcupine/plugins/git_status.py
```diff
@@ -78,34 +78,41 @@ def __init__(self, tree: DirectoryTree, project_id: str):
...
+            # status of a folder applies to everything it contains
+            for folder in item_path.parents:
+                try:
+                    return path_to_status[folder]
+                except KeyError:
+                    if folder == self.project_path:
+                        break
 
         # Handle directories containing files with different statuses
         substatuses = {
```
Akuli, line 118: The code below loops over all the statuses and still isn't really ideal. I'm starting to think that `dict[Path, status string]` is maybe not the ideal data structure for this,.

porcupine/plugins/git_status.py
```diff
@@ -70,6 +70,27 @@ def run_git_status(project_root: Path) -> dict[Path, str]:
...
         else:
             log.warning(f"unknown git status line: {repr(line)}")
+
+    # When a folder contains files with different statuses, decide status of folder
+    # There can be lots of statuses, not good to loop through them in gui thread
+    folder_to_content_statuses: dict[Path, set[str]] = {}
+    for path, status in result.items():
+        if status in {"git_added", "git_modified", "git_mergeconflict"}:
+            for folder in path.parents:
+                folder_to_content_statuses.setdefault(folder, set()).add(status)
```
Akuli, line 80:
```suggestion
                folder_to_content_statuses.setdefault(folder, set()).add(status)
                if folder == project_root:
                    break
```

porcupine/plugins/git_status.py
```diff
@@ -78,51 +99,30 @@ def __init__(self, tree: DirectoryTree, project_id: str):
...
-        if "git_modified" in substatuses:
-            return "git_modified"
-        if "git_added" in substatuses:
-            return "git_added"
-
-        assert not substatuses
-        return None
+        assert self._git_status_future is not None
+        path_to_status = self._git_status_future.result(timeout=0)
+        return path_to_status.get(item_path)
```
Akuli, line 125:
```suggestion
        return path_to_status.get(item_path, None)
```

PR #774, opened 2021-09-27
directory tree: single-letter key bindings to avoid pressing arrow keys many times

### LINKED ISSUE #564: directory tree: finding something from a long directory listing
Would be nice to just start typing something and find it

Akuli:
fixes #564

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -339,7 +339,25 @@ class DirectoryTree(ttk.Treeview):
             return result
         return None
 
-    # TODO: invoking context menu from keyboard
+    def _cycle_through_items(self, event: tkinter.Event[DirectoryTree]) -> None:
+        if len(event.char) != 1:
+            return
+
+        try:
+            [item] = self.selection()
+        except ValueError:  # nothing selected
+            return
+
+        children = [c for c in self.get_children(self.parent(item)) if self.item(c, 'text').startswith(event.char)]
+        if not children:
+            return
+
+        try:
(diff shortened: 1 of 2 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -432,8 +452,11 @@ def setup() -> None:
     tree.refresh()
 
     # TODO: mac right click = button 2?
+    # TODO: invoking context menu from keyboard
     tree.bind("<Button-3>", tree._on_right_click, add=True)
 
+    tree.bind("<Key>", partial(tree._cycle_through_items), add=True)
```
Akuli, line 458:
```suggestion
    tree.bind("<Key>", tree._cycle_through_items, add=True)
```

tests/test_directory_tree_plugin.py
```diff
@@ -177,3 +177,31 @@ def test_home_folder_displaying():
...
+    tree.event_generate("b")
+    assert get_path(tree.selection()[0]) == tmp_path / "bar.txt"
+    tree.event_generate("b")
+    assert get_path(tree.selection()[0]) == tmp_path / "baz.txt"
+    tree.event_generate("b")
+    assert get_path(tree.selection()[0]) == tmp_path / "bar.txt"
+    tree.event_generate("R")
+    assert get_path(tree.selection()[0]) == tmp_path / "README"
+    tree.event_generate("R")
+    assert get_path(tree.selection()[0]) == tmp_path / "README"
```
Akuli, line 207: should add test case with presing some letter that doesnt exist like `x`

PR #790, opened 2021-10-02
Run plugin rewrite

### LINKED ISSUE #32: idea: command runner thingy for better workflow
My workflow is typically like this:

```
while True:
    edit a file in porcupine
    run a command in a terminal
    if command succeeded:
        break
    look up a file name and a line number from the output of the command (e.g. compiler error or Python error message)
    find the file among the many tabs I have open in porcupine
    Ctrl+L to jump to the line number
```

My idea is to make this possible without a terminal. To run a command, I would type the command into a text entry in porcupine, and then porcupine would display the output of the command in the terminal. There should also be a convenient keyboard shortcut for running the same command again, which is what pressing up arrow and then enter does in a terminal.

When displaying the output, any file name and line number pairs (e.g. `File "foo.py", line "123"` or `bar.c:123`) should be clickable, and clicking that would open the corresponding file at the given line number in Porcupine.

Challenging things with displaying command output in tkinter:
- `input()`
- infinite loops and everything else that print stuff very fast (need blocking write?)
- printing colors on terminal
- making it appear as an actual terminal for `sys.stdout.isatty()` and similar things in other programming languages (many programs display different output depending on this, e.g. grep displays no colors)
- curses programs (these just won't run in porcupine window lol)

The problem is that I should implement some reasonable subset of the features available in an actual terminal.

Akuli:
fixes #32 
- [x] F5 = repeat previous command (or one of the default commands)
- [ ] F6,F7,F8 = repeat a different command
- [ ] venv support, sourcing rc scripts in general?
- [ ] rerunning two commands conveniently?
- [ ] make file names and line numbers clickable
- [ ] improve how dialog looks with long commands
- [x] test on windows

Created separate issues for unchecked items


porcupine/default_filetypes.toml
```diff
@@ -209,74 +188,110 @@ autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = '(case
...
+command = "make"
+working_directory = "{project_path}"
+external_terminal = false
+
 # TODO: Windows batch files and powershell files
-#
-# i'm not trying to discriminate anyone with pygments_lexer and
-# commands, override them in your user-specific config you want to
-#
 # shebang_regex is partly copy/pasted from nano's default config
```
Akuli, line 270:
```suggestion
# override pygments_lexer if you want to
# shebang_regex is partly copy/pasted from nano's default config
```

PR #792, opened 2021-10-02
MacOS right click

### LINKED ISSUE #751: directory tree mac-specific code
Developed without having a Mac user to test all this:
- [ ] Does right-clicking directory tree bring up a menu?
- [ ] Does "Move to trash" work?
- [ ] Does "Open in file manager" work? (directories only)

1anakin20:
Thanks @Akuli for your help with #751. I have added the right click event to the default_keybinds. The right click in the tree works in windows and macOS.

Fixes #751

### DIFF
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -61,13 +61,10 @@ for {set i 1} {$i <= 9} {incr i} {
 
 # tab_closing plugin
 event add "<<TabClosing:XButtonClickClose>>" <Button-1>
-if {[tk windowingsystem] == "aqua"} {
-    # right-click is Button-2, no wheel-click (afaik)
-    event add "<<TabClosing:ShowMenu>>" <Button-2>
-} else {
-    # right-click is Button-3, wheel-click is Button-2
+# no wheel click on mac (afaik)
+if {[tk windowingsystem] != "aqua"} {
+    # Wheel-click is Button-2
     event add "<<TabClosing:ShowMenu>>" <Button-3>
-    event add "<<TabClosing:HeaderClickClose>>" <Button-2>
 }
 
 # sort plugin
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -3,8 +3,10 @@
 # Use Command on mac, Control on other systems
 if {[tk windowingsystem] == "aqua"} {
     set contmand Command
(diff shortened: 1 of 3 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -461,9 +461,7 @@ def setup() -> None:
             tree.add_project(path, refresh=False)
     tree.refresh()
 
-    # TODO: mac right click = button 2?
-    # TODO: invoking context menu from keyboard
```
Akuli, line 465: It would be good to leave the keyboard comment here. You currently need a mouse click to open the menu, even though the directory tree can otherwise be used with just the keyboard.

porcupine/default_keybindings.tcl
```diff
@@ -3,8 +3,10 @@
 # Use Command on mac, Control on other systems
 if {[tk windowingsystem] == "aqua"} {
     set contmand Command
+    event add "<<RightClick>>" <Button-2>
 } else {
     set contmand Control
+    event add "<<RightClick>>" <Button-3>
 }
```
Akuli, line 10: `<<TabClosing:ShowMenu>>` is defined below in the same file. Can you delete it and change the `tab_closing` plugin to use `<<RightClick>>` instead?

Akuli (commented):
Thanks! A couple nits.

porcupine/default_keybindings.tcl
```diff
@@ -59,13 +61,10 @@ for {set i 1} {$i <= 9} {incr i} {
...
 event add "<<TabClosing:XButtonClickClose>>" <Button-1>
-if {[tk windowingsystem] == "aqua"} {
-    # right-click is Button-2, no wheel-click (afaik)
-    event add "<<TabClosing:ShowMenu>>" <Button-2>
-} else {
-    # right-click is Button-3, wheel-click is Button-2
+# no wheel click on mac (afaik)
+if {[tk windowingsystem] != "aqua"} {
+    # Wheel-click is Button-2
     event add "<<TabClosing:ShowMenu>>" <Button-3>
```
Akuli, line 67: I think you deleted the wrong line. The following suggestion changes it so that there is no `<<TabClosing:ShowMenu>>` but `<<TabClosing:HeaderClickClose>>` is unchanged:
```suggestion
    event add "<<TabClosing:HeaderClickClose>>" <Button-2>
```

Akuli (approved):
Thanks!

PR #794, opened 2021-10-03
UI fixes

rdbende:


rdbende:
I fixed many dialog's buttons, but the builtin messageboxes are still awkward with their tiny buttons :(
image

porcupine/plugins/pluginmanager.py
```diff
@@ -56,12 +56,12 @@ def __init__(self, master: tkinter.Misc) -> None:
         self._plz_restart_label = ttk.Label(self.content_frame)
         self._plz_restart_label.pack(side="bottom", fill="x")
 
-        column_sizes = [120, 150, 180]
+        column_sizes = (120, 150, 180)
 
-        left_side = ttk.Frame(panedwindow)
+        left_side = ttk.Frame(panedwindow, padding=10)
```
Akuli, line 61: The left side of the plugin manager looks a bit ridiculous IMO with that much padding. UIs with too much padding make me feel like I am struggling to see the content of a tiny box.

```suggestion
        left_side = ttk.Frame(panedwindow, padding=5)
```

porcupine/plugins/pluginmanager.py
```diff
@@ -71,15 +71,14 @@ def __init__(self, master: tkinter.Misc) -> None:
...
+        self._search_var = tkinter.StringVar(value="Filter by name, type or status...")
         search_entry = ttk.Entry(left_side, textvariable=self._search_var)
         search_entry.bind(
             "<FocusIn>", (lambda event: search_entry.selection_range(0, "end")), add=True
         )
-        search_entry.insert(0, "Filter by name, type or status...")
         self._search_var.trace_add("write", self._search)
 
-        search_entry.pack(side="bottom", fill="x")
+        search_entry.pack(side="bottom", fill="x", pady=(10, 0))
```
Akuli, line 81: Same here:
```suggestion
        search_entry.pack(side="bottom", fill="x", pady=(5, 0))
```

porcupine/settings.py
```diff
@@ -591,7 +591,7 @@ def _grid_widgets(
 ) -> None:
     label = ttk.Label(chooser.master, text=label_text)
     label.grid(column=0, sticky="w")
-    chooser.grid(row=label.grid_info()["row"], column=1, sticky="we")
+    chooser.grid(row=label.grid_info()["row"], column=1, sticky="we", pady=5)
```
Akuli, line 594: This `pady` does not affect checkboxes, because they don't have a separate label next to them. This means that checkboxes are a bit closer together than everything else. Ideally there would be a way to set a spacing between all items, but if there isn't anything like that, we can just add the same `pady` to checkboxes too.

Akuli (commented):
Thanks! I like most of the new settings.

Akuli:
I don't think we should do anything about built-in message boxes. Making our own message boxes instead would be a lot of work and it would likely cause weird inconsistencies on different operating systems.

porcupine/settings.py
Akuli: One option might be to loop through `winfo_children()` when the dialog is shown.

porcupine/plugins/pluginmanager.py
rdbende: Yeah, it looks indeed better with just 5 px

Akuli (approved):
Thanks! Sometimes the CI fails randomly due to a timeout, see #379

Issue #793, opened 2021-10-03
copy file


Akuli:
It would be nice if there was an easy way to copy a file by right-clicking it. You can already rename, and I imagine copying would work very similarly.

PR #808, opened 2021-10-03
Copy-paste on right-click

nicolafan:
Fixes: #793


porcupine/plugins/filemanager.py
```diff
@@ -122,6 +125,24 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    global copy_path
+    new_file_path = os.path.join(new_path, os.path.basename(copy_path))
```
Akuli, line 130: Porcupine uses pathlib, which makes things like this a lot easier. You can probably find some guides for how to use it with google. For example, this line is best written like this:

```suggestion
    new_file_path = new_path / copy_path.name
```

Let me know if you need help with rewriting the code to use pathlib.

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +125,24 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    global copy_path
```
Akuli, line 129: Globalling here is not needed, because this function only uses the current value of `copy_path`. You need `global` only when modifying. But to make the type checker happy, you need to ensure it is not `None`.

```suggestion
    assert copy_path is not None
```

porcupine/plugins/filemanager.py
```diff
@@ -25,6 +26,8 @@
 else:
     trash_name = "trash"
 
+copy_path = ""
```
Akuli, line 29: It is better to use `None` as the "missing" value. You also need to add an annotation for the type checker to know what else it can be than `None`.

```suggestion
copy_path: Path | None = None
```

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +125,24 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    global copy_path
+    new_file_path = os.path.join(new_path, os.path.basename(copy_path))
+
+    if os.path.exists(new_file_path):
+        log.error("already exists - copy aborted")
+        return
```
Akuli, line 134: IMO it shouldn't be an error to copy from a directory into itself. The file should just get a different name, e.g. `foo.py` --> `foo (2).py` or something. Or we could ask the user how to name the file.

porcupine/plugins/filemanager.py
```diff
@@ -221,6 +242,8 @@ def is_NOT_project_root(path: Path) -> bool:
     # Doing something to an entire project is more difficult than you would think.
     # For example, if the project is renamed, venv locations don't update.
     # TODO: update venv locations when the venv is renamed
+    Command("Copy", None, is_NOT_project_root, copy),
+    Command("Paste", None, (lambda p: p.is_dir() and copy_path != ''), paste),
```
Akuli, line 246: Several suggestions:
- Use `<<Copy>>` and `<<Paste>>` virtual events which are Command+C on Mac, and Ctrl+C on windows and linux. Copy/pasting without Ctrl+C Ctrl+V just wouldn't feel right :)
- Use `None` as the "missing" value for `copy_path`
- Rename "Paste" to "Paste here", because the destination must be a directory

```suggestion
    Command("Copy", "<<Copy>>", is_NOT_project_root, copy),
    Command("Paste here", "<<Paste>>", (lambda p: p.is_dir() and copy_path is not None), paste),
```

Akuli (commented):
Thanks! I somehow didn't think of copying a file from a directory

porcupine/plugins/filemanager.py
nicolafan: Exactly, I wanted to ask you about it. I'm a Ubuntu user and the file explorer asks me how to call the new file when I add it to a folder which contains a file with the same name. What do you think is the right approach for Porcupine? I would be happy to implement it.

porcupine/plugins/filemanager.py
nicolafan: Didn't know about it. I'll check it!

porcupine/plugins/filemanager.py
Akuli: I think it would be best to modify the renaming code so that the same dialog can be used to ask the new name when copying. If my copy was named `foo (2).py`, I would have to rename it myself anyway.

nicolafan:
Thank you for all your suggestions @Akuli . I'd like to work on this issue to implement the functionality in the best way. 
Your explanations have been very useful.

nicolafan:
Schermata da 2021-10-03 22-42-43

porcupine/plugins/filemanager.py
```diff
@@ -78,7 +109,7 @@ def validate_name(*junk: object) -> None:
             ok_button.config(state="disabled")
             return
 
-        if possible_new_path.exists():
+        if choice.get() == 1 and possible_new_path.exists():
```
Akuli, line 112: It's not clear what the integer `1` does here. Instead of values `0` and `1`, it might be simpler to use values `"overwrite"` and `"change_name"`, or something like that. You will need to change `choice` to be a `StringVar`.

Akuli (commented):
Not a full review, will review `validate_name()` and `paste()` tomorrow

porcupine/plugins/filemanager.py
Akuli: Actually, since it has two values, a `BooleanVar` would make sense the most. It could be named so that it is obvious which choice is which, e.g. `overwrite_var`.

porcupine/plugins/filemanager.py
```diff
@@ -221,6 +281,8 @@ def is_NOT_project_root(path: Path) -> bool:
     # Doing something to an entire project is more difficult than you would think.
     # For example, if the project is renamed, venv locations don't update.
     # TODO: update venv locations when the venv is renamed
+    Command("Copy", "<<Copy>>", (lambda p: not p.is_dir()), copy),
+    Command("Paste here", "<<Paste>>", (lambda p: p.is_dir() and copy_path is not None), paste),
```
Akuli, line 285: After trying this out, I found "Paste here" to be quite confusing. If I have a file selected and I press `Ctrl+C Ctrl+V`, I would expect to get a copy of the file in the same directory where it is. Maybe we should have two options, "Paste" (possible whenever the selection is not a project, I would prefer this to be Ctrl+V) and "Paste inside selected directory" (the current "Paste here")?

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,31 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    assert copy_path is not None
+    new_file_path = new_path / copy_path.name
+
+    if new_file_path.exists():
+        if new_file_path.parent == copy_path.parent:  # copying file in the same dir (sequential number copy)
```
Akuli, line 165: I don't like the sequential numbering, as a file named `foo (1).py` is unacceptable to me and I would have to rename it anyway. Can we just always ask for a new name?

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,31 @@ def rename(old_path: Path) -> None:
...
+            path_stem = new_file_path.stem
+            path_suffix = new_file_path.suffix
+            while new_file_path.exists():
+                path_number += 1
+                new_file_path = new_file_path.with_name(path_stem + f" ({path_number})" + path_suffix)
+        else:
+            new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
+
+    if new_file_path is not None:
+        new_file_path.touch()
```
Akuli, line 176: This shouldn't be necessary, `shutil.copy` creates the file if it doesn't exist.

porcupine/plugins/filemanager.py
```diff
@@ -70,6 +83,24 @@ def select_name() -> None:
     ok_button = ttk.Button(button_frame, text="OK", command=select_name, state="disabled", width=1)
     ok_button.pack(side="right", expand=True, fill="x", padx=(5, 0))
 
+    def overwriting() -> None:
+        entry.config(state="disabled")
+        ok_button.config(state="enabled")
+
+    def renaming() -> None:
+        entry.config(state="enabled")
+        validate_name()
```
Akuli, line 92: Maybe `validate_name()` should run every time the `choice` variable changes? You can do that with `choice.trace_add`, just like how there is already `var.trace_add`.

Feel free to rename `var`. It is no longer clear what it is because there are two tkinter variables.

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,31 @@ def rename(old_path: Path) -> None:
...
+            path_suffix = new_file_path.suffix
+            while new_file_path.exists():
+                path_number += 1
+                new_file_path = new_file_path.with_name(path_stem + f" ({path_number})" + path_suffix)
+        else:
+            new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
+
+    if new_file_path is not None:
+        new_file_path.touch()
+        shutil.copy(copy_path, new_file_path)
```
Akuli, line 177: The new file does not appear in the directory tree when the copying is done. I think adding `get_directory_tree().refresh()` should fix it.

porcupine/plugins/filemanager.py
nicolafan: Hi @Akuli, I think you suggested the floowing behaviour:
(Ctrl + V) + no dir selected -> copy file inside same directory
(Ctrl + V or (right-click + "Paste inside selected directory)) + dir selected -> copy file inside selected directory

P.S.
What to do in case: (Ctrl + V) + file selected? (I think that if the selected file is in another directory and we press Ctrl + V and paste in the same directory as in the first rule it would be a little confusing... Should we block Ctrl + V or should we let the copy be pasted in the directory which contains the selected file (this is how IntelliJ works) ? )

porcupine/plugins/filemanager.py
nicolafan: Perfect

porcupine/plugins/filemanager.py
nicolafan: This is strange, in my case it automatically refreshes without the instruction

porcupine/plugins/filemanager.py
Akuli: If I have `foo/bar.py` selected and I press Ctrl+V, it will copy to `foo/blah.py`.

If I have `foo/subfolder/` selected and I press Ctrl+V, that too will copy to `foo/blah.py`.

If I have `foo/subfolder/` selected and I select "Paste inside selected directory", it will copy to `foo/subfolder/blah.py`.

So it does not matter what you have selected, what matters is where the selected file or directory is. In all cases, the new "Paste here" would do same as Ctrl+V. If you have a better suggestion to use instead of "Paste here", feel free to name it better.

porcupine/plugins/filemanager.py
Akuli: It depends on how focusing events work with the window manager you use. It refreshes when the Porcupine window is focused:

https://github.com/Akuli/porcupine/blob/719da27426350b503c910850fb757716b366efe0/porcupine/plugins/directory_tree.py#L442

porcupine/plugins/filemanager.py
nicolafan: Not always when it changes, but only when passing from `overwrite_var = False` to `True` (this is the new name of the variable).

Akuli: Why not always? Why should it update when switching from `True` to `False`, but stay as is if you then switch back to `True`?

Akuli: Here's how I imagined this:
- `overwriting()` and `renaming()` set the entry to enabled or disabled.
- `validate_name()` should sets `ok_button` to enabled or disabled. Feel free to rename it to `update_ok_button_state()` or something.

This way it would be very clear what each function does, because the state of `ok_button` would be updated from one function only. Then it makes sense to run `validate_name()` whenever any of the two variables change, simply because values of both variables are used when deciding the state of the button.

nicolafan: So `overwrite_var == True` if the overwrite `RadioButton` is checked. When it's checked I thought that the `OK` button to go on should always be enabled since it's always possible to overwrite. Let's say that I want to copy `file.txt` where there's another file with this name. I have overwrite checked as default. Then I check 'change name of destination' and modify the new name to `filex.txt`. OK is enabled. Then i modify back to `file.txt` and `validate_name()` disables `OK`. Now `overwrite_var == False` since I'm modifying the name. When i click again the `RadioButton` for overwriting I want the `OK` button to be enabled no matter what's written in the entry field (also the entry field is disabled) so if I call again `validate_name()` it will disable `OK`.

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,28 @@ def rename(old_path: Path) -> None:
...
+
+
+def paste_here(new_path: Path) -> None:
+    assert copy_path is not None
+    new_file_path = new_path / copy_path.name
+
+    if new_file_path.exists():
+        new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
+
+    if new_file_path is not None and not copy_path == new_file_path:
```
Akuli, line 172: This is a bit hard to read, because the `None` check is not near the code that can set `new_file_path` to `None`. How about this?

```suggestion
    if new_file_path.exists():
        new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
        if new_file_path is None:
            return

    if copy_path != new_file_path:
```

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,28 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    assert copy_path is not None
```
Akuli, line 161: Not needed for type checker to be happy, because this function doesn't use `copy_path`.

```suggestion
```

porcupine/plugins/filemanager.py
nicolafan: You wrote while I was writing :) yes, I can refactor it like this.

porcupine/plugins/filemanager.py
```diff
@@ -25,51 +25,84 @@
...
 
+    if is_paste and old_path.parent != copy_path.parent:
+        dialog.title("File conflict")
+        ttk.Label(
+            big_frame, text=f"{old_path.parent} already has a file named {old_path.name}."
+                            f"\nDo you want to overwrite it?",
+            wraplength=label_width
+        ).pack(fill="x")
+    else:
+        dialog.title("Rename")
```
Akuli, line 59: The title gets set to "Rename" when I paste a file into the same directory as where it is. This isn't great because I'm renaming, not copying. I think we need a third case for pastes within the same parent directory.

porcupine/plugins/filemanager.py
```diff
@@ -25,51 +25,84 @@
...
+        ttk.Label(
+            big_frame, text=f"Enter a new name for {old_path.name}:", wraplength=label_width
+        ).pack(fill="x")
+
+    entry_frame = ttk.Frame(big_frame)
+    entry_frame.pack(fill="x")
     var = tkinter.StringVar()
```
Akuli, line 66: Now there is `var` and `overwrite_var`. I think this original `var` needs to be renamed. How about `file_name_var`?

porcupine/plugins/filemanager.py
```diff
@@ -25,51 +25,84 @@
...
     cancel_button.pack(side="left", expand=True, fill="x", padx=(0, 5))
     ok_button = ttk.Button(button_frame, text="OK", command=select_name, state="disabled", width=1)
     ok_button.pack(side="right", expand=True, fill="x", padx=(5, 0))
 
+    def overwriting() -> None:
+        entry.config(state="disabled")
+        ok_button.config(state="enabled")
+
+    def renaming() -> None:
+        entry.config(state="enabled")
```
Akuli, line 95: I think the correct state is `"normal"`. Apparently non-existent states default to same as `"normal"`, so `"enabled"` works too, but `"normal"` would be more explicit.

Akuli (commented):
This PR turned out to be trickier than I expected. Great work so far! Let me know if you at any time feel like you are no longer want to work on it, and I'll finish it.

porcupine/plugins/filemanager.py
```diff
@@ -25,52 +25,84 @@
...
+            new_path = old_path.with_name(entry.get())
         dialog.destroy()
 
     cancel_button = ttk.Button(button_frame, text="Cancel", command=dialog.destroy, width=1)
     cancel_button.pack(side="left", expand=True, fill="x", padx=(0, 5))
     ok_button = ttk.Button(button_frame, text="OK", command=select_name, state="disabled", width=1)
     ok_button.pack(side="right", expand=True, fill="x", padx=(5, 0))
 
-    def validate_name(*junk: object) -> None:
+    if is_paste and copy_path.parent != old_path.parent:
```
Akuli, line 90: I don't like how `ask_name_for_renaming()` uses `copy_path` only to do this check. It has 3 different modes:
- Renaming
- Pasting into the same directory (do not allow overwriting, as that would make no sense)
- Pasting into a different directory (ask if the user wants to overwrite)

I think one boolean and one global variable is not a good way to describe the 3 things this function can do. Splitting into more functions doesn't seem great either, because there's a lot of code shared between all the 3 different modes. It would be nice if I could see the "3 modes" idea without having to read the whole function, as it is quite long.

A couple ideas, pick your favorite or do something even better:
- Two booleans: `def ask_name_for_renaming(old_path: Path, is_paste: bool, show_overwriting_option: bool)`
- Using an enum with 3 members (feels like overkill to me)

The function name isn't great either, since the dialog is no longer all about renaming. How about `ask_file_name()` or something?

porcupine/plugins/filemanager.py
```diff
@@ -83,20 +115,23 @@ def validate_name(*junk: object) -> None:
...
     dialog.wait_window()
+
     return new_path
 
 
 def rename(old_path: Path) -> None:
     new_path = ask_name_for_renaming(old_path)
     if new_path is None:
         return
+    clean_copy_path(old_path)
```
Akuli, line 134: This rename function is not the only way how a file can get renamed. You can still use other programs to rename files. This means that `clean_copy_path()` won't work correctly in all cases, and at best it works most of the time.

If I understand correctly, the purpose of `clean_copy_path()` is to prevent being able to paste after `Ctrl+C`-ing a file and then deleting or renaming it, as a paste like that would result in a `FileNotFoundError`. I think it would be better to instead check if `copy_path.exists()` before using `copy_path`. We already check whether `copy_path is not None`, so that would mean replacing every `copy_path is not None` check with `copy_path is not None and copy_path.exists()`. Instead of `exists()`, it might be even better to use `is_file()`.

Akuli (commented):
It's starting to look pretty good!

porcupine/plugins/filemanager.py
nicolafan: Somehow i didn't think about it D:

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +157,34 @@ def rename(old_path: Path) -> None:
...
+def paste_here(new_path: Path) -> None:
+    assert copy_path is not None
+    assert new_path.is_dir()
+
+    new_file_path = new_path / copy_path.name
+
+    if new_file_path.exists():
+        new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
+        if new_file_path is None:
+            return
```
Akuli, line 173: This currently creates a mypy error. Because `ask_name_for_renaming` can return `None` but `new_file_path` has type `Path` (as it was initially set to `new_path / copy_path.name`), mypy complains that `new_file_path` can become `None`, even though it is only temporarily `None` and the rest of the code can still assume it is `Path`. We could report this to mypy's issue tracker, but it is likely that someone has already reported, and even more likely that nobody would fix it (there's lots of open issues).

Here is a workaround:

```suggestion
        path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
        if path is None:
            return
        new_file_path = path
```

Now `path` will have type `Path | None` before the `return` statement, but because we check for `None`, it has type `Path` after the `return` statement and we can assign it to `new_file_path`.

porcupine/plugins/filemanager.py
```diff
@@ -162,20 +161,24 @@ def paste(new_path: Path) -> None:
...
 
     new_file_path = new_path / copy_path.name
 
     if new_file_path.exists():
-        new_file_path = ask_name_for_renaming(new_file_path, is_paste=True)  # rename-overwrite file conflict
-        if new_file_path is None:
+        show_overwriting_option = True
+        if new_file_path.parent == copy_path.parent:
+            show_overwriting_option = False
+        path = ask_file_name(new_file_path, is_paste=True, show_overwriting_option=show_overwriting_option)
```
Akuli, line 173: A simpler way to view this: whether to show overwriting option is same as whether we are copying into a different directory.
```suggestion
        path = ask_file_name(new_file_path, is_paste=True, show_overwriting_option=(new_file_path.parent != copy_path.parent))
```
The resulting line of code is now a bit long, but `black` will split this up on several lines.

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +156,38 @@ def rename(old_path: Path) -> None:
...
+        if path is None:
+            return
+        new_file_path = path
+
+    if copy_path != new_file_path:
+        shutil.copy(copy_path, new_file_path)
+        get_directory_tree().refresh()
+        if copy_path == new_file_path:  # if overwrite and file is open refresh page
+            close_tabs(find_tabs_by_parent_path(new_file_path))
+            get_tab_manager().open_file(new_file_path)
```
Akuli, line 183: These 3 lines can be deleted. The paths will never be the same, since all this is inside an `if` statement that makes it run only when the paths differ. With copying, it's fine if the original file remains open, and the user has to open the newly created copy explicitly.

porcupine/plugins/filemanager.py
```diff
@@ -217,10 +283,17 @@ def is_NOT_project_root(path: Path) -> bool:
...
+    return is_NOT_project_root(path) and copy_path is not None and copy_path.is_file()
+
+
 commands = [
     # Doing something to an entire project is more difficult than you would think.
     # For example, if the project is renamed, venv locations don't update.
     # TODO: update venv locations when the venv is renamed
+    Command("Copy", "<<Copy>>", (lambda p: not p.is_dir()), copy),
+    Command("Paste", "<<Paste>>", can_paste, paste),
+    Command("Paste here", None, (lambda p: p.is_dir() and copy_path is not None and copy_path.is_file()), paste_here),
```
Akuli, line 296: The condition is a bit long. It might be nicer to put it into a function, similarly to `can_paste()`.

Akuli (commented):
The final nits...

Before merging, you will have to run `black`, e.g. `black porcupine/`. This formats the code in a specific way that you might not like, but the way it formats is consistent across the whole project, and I can't complain about your code formatting as that is done by `black`.

porcupine/plugins/filemanager.py
nicolafan: Yes, here I actually made a conceptual error. Before the last commit this check was `if copy_path.name == new_file_path.name`. Then I changed it like this without thinking more.
The purpose I want to reach with these actions is:
check if the operation is an overwrite -> if `the overwritten file is open in the editor` then `refresh the tab of the file since its content changed`. It is not related to the first copied file. I'll think on how to implement it or I'll implement your solution.

porcupine/plugins/filemanager.py
nicolafan: Sure

porcupine/plugins/filemanager.py
nicolafan: Actually it was working when the condition was `if copy_path.name == new_file_path.name`. The problem that made me erroneously changed it was that it did this operation even in the case the copied file was pasted without problems with its name in another dir. 
So I have to check if the names are the same and if there was a conflict before (or you think there's another way we can talk about it).

Akuli: Ah, overwriting an open file is something I didn't think of. We can fix this in a separate PR.

nicolafan: This PR had so many hidden edge cases. Hopefully we solved (almost) all of them :)

Akuli: Yeah, I didn't realize it was going to be this tricky either.

nicolafan:
Maybe I guess right. It is not possible to put `is not None` in a function that is called by `assert` (?).
I run black on `filemanager.py` btw.

porcupine/plugins/filemanager.py
```diff
@@ -122,6 +169,38 @@ def rename(old_path: Path) -> None:
         tab.path = new_path / tab.path.relative_to(old_path)
 
 
+def paste(new_path: Path) -> None:
+    paste_here(new_path.parent)
+
+
+def paste_here(new_path: Path) -> None:
+    assert is_copy_path_valid()
+    assert new_path.is_dir()
```
Akuli, line 178: Unfortunately mypy (the type checker) isn't clever enough to see how `assert is_copy_path_valid()` ensures thjat `copy_path` can't be `None`. That currently causes a mypy error. The rest of the checks are not needed, because theu don't affect type checking and they are already handled by `can_paste_here(new_path)`.

```suggestion
    assert copy_path is not None
```

nicolafan: I'm sorry for these repeated problems. I've never used mypy before. Should check smth about it btw.

Akuli: It's ok, I'm here to help :)

nicolafan: I don't know why **black check** wasn't passed. I ran it and now that I'm trying again it says that there's nothing to change.

Akuli (approved):
Thanks for your work!

nicolafan:
Thank you @Akuli . I've learned so much with your help! 😀

Issue #432, opened 2021-05-02
emoji not working


Akuli:
In general, Tk doesn't support emoji characters (afaik anything beyond `U+FFFF`). Even if those characters aren't fully supported, it would be nice if Porcupine could edit files containing them.

Moosems:
They now support it a little but it causes an issue if there's a tag on any of them.

Akuli:
I guess one dumb workaround is to open the file as latin-1. That way emojies will show up as multiple garbage characters, but at least you won't get surprising errors and other problems.

Moosems:
Try editing the files now without highlighting ;).

benjamin-kirkbride:
Why does this not work? I would like to be able to use emoji's in the Markdown docs I'm working on.

rdbende:
Tk in general doesn't really support emojis. There are a couple of them that work (kinda, I mean they doesn't cause the app to crash), but even those behave really weirdly when 3diting

Moosems:
Why I recommend we remove tags from emojis. It removes most of the actual bugs other than the cut off visual appearance.

PR #809, opened 2021-10-03
Black error messagebox

### LINKED ISSUE #796: black: bad error message when file has syntax error
`black` prints a reasonable error message to the terminal, but Porcupine doesn't capture it and instead shows a confusing python traceback

rdbende:
Fixes #796

I think it's good enough

### DIFF
--- a/porcupine/plugins/python_tools.py
+++ b/porcupine/plugins/python_tools.py
@@ -27,6 +27,8 @@ def run_tool(tool: str, code: str, path: Path | None) -> str:
         )
         return code
 
+    fail_str = f"Running {tool} failed"
+
     try:
         # run in subprocess just to make sure that it can't crash porcupine
         # set cwd so that black/isort finds its config in pyproject.toml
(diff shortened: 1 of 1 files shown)


porcupine/plugins/python_tools.py
```diff
@@ -36,13 +36,23 @@ def run_tool(tool: str, code: str, path: Path | None) -> str:
...
+            stderr=subprocess.PIPE,
             cwd=(Path.home() if path is None else path.parent),
             input=code.encode("utf-8"),
         )
         return result.stdout.decode("utf-8")
     except Exception:
         log.exception(f"running {tool} failed")
-        messagebox.showerror(f"Running {tool} failed", traceback.format_exc())
+        if tool == "black":
+            messagebox.showerror(
```
Akuli, line 47: The messagebox should probably not go under `except Exception:`. It should instead catch whatever more specific error gets raised when the subprocess exits with an error status. I think it is `subprocess.CalledProcessError`, but I'm not sure. This way it's less likely to show an error message telling about syntax errors when the code doesn't have syntax errors. Also, I don't see any reason to not do this with `isort`.

So instead of "oh no error for `black` and traceback error for `isort`", I think the logic should be "oh no error for `subprocess.CalledProcessError` and traceback error for other exceptions".

porcupine/plugins/python_tools.py
```diff
@@ -36,13 +36,23 @@ def run_tool(tool: str, code: str, path: Path | None) -> str:
...
+                f"Running {tool} failed",
+                # can't black's use emojis in tkinter :(
```
Akuli, line 49: The error message is:

```
_tkinter.TclError: character U+10000 is above the range (U+0000-U+FFFF) allowed by Tcl
```

So you can simply remove the bad characters, like this for example:
```python
tkinter_safe_string = "".join(char for char in black_output if ord(char) <= 0xFFFF)
```

There is also #432 which is mostly about editing files containing emojies.

Akuli (approved):
Thanks!

PR #814, opened 2021-10-03
clickable links to command output

### LINKED ISSUE #805: clickable file names and line numbers
Would be nice to run some command like `mypy porcupine/`, then click the file names it prints to see code causing warnings

Akuli:
Fixes #805

### DIFF
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -53,12 +53,11 @@ class NoTerminalRunner:
         self.textwidget.tag_bind("link", "<Leave>", self._leave_link)
         self.textwidget.tag_bind("link", "<Button-1>", self._open_link)
 
-        self._output_queue: queue.Queue[tuple[str, str]] = queue.Queue()
+        self._output_queue: queue.Queue[tuple[Path, str, str]] = queue.Queue()
         self._running_process: subprocess.Popen[bytes] | None = None
         self._queue_handler()
 
         self._links: dict[str, tuple[Path, int]] = {}
-        self._cwd: Path | None = None
 
     # TODO: much links code copy/pasted from aboutdialog plugin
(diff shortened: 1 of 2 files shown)


porcupine/plugins/run/no_terminal.py
```diff
@@ -3,27 +3,84 @@
...
 
 from porcupine import get_tab_manager, images, utils
 
 log = logging.getLogger(__name__)
 
 
+filename_regex_parts = [
+    # c compiler output
+    # playground.c:4:9: warning: ...
+    r'^([^:]+):([0-9]+)(?=:)'
```
Akuli, line 21: should also add python syntax

porcupine/plugins/run/no_terminal.py
```diff
@@ -3,27 +3,84 @@
...
+                tab.textwidget.tag_remove("sel", "1.0", "end")
+                tab.textwidget.tag_add("sel", "insert", "insert lineend")
+                break
+
+    def run_command(self, cwd: Path, command: str) -> None:
+        self._cwd = cwd
+
+        # this is a daemon thread because i don't care what the fuck
+        # happens to it when python exits
+        threading.Thread(target=self._runner_thread, args=[cwd, command], daemon=True).start()
```
Akuli, line 83: It is a bit weird to pass cwd to two places. Maybe instead put it to queue and use it for links that way?

porcupine/plugins/run/no_terminal.py
```diff
@@ -83,11 +135,22 @@ def _queue_clearer(self) -> None:
                 if tag == "clear":
                     assert not text
                     self.textwidget.delete("1.0", "end")
+                    for tag in self._links.keys():
+                        self.textwidget.tag_delete(tag)
```
Akuli, line 139: Not really necessary, the link-specific tags don't have any bindings or configuration associated with them

```suggestion
```

PR #815, opened 2021-10-03
run plugin: set PYTHONUNBUFFERED when not using external terminal

Akuli:
Fixes #802

PR #816, opened 2021-10-03
better tests for run plugin

Akuli:
Will hopefully make #814 and #815 testable

### DIFF
--- a/tests/test_run_plugin.py
+++ b/tests/test_run_plugin.py
@@ -10,15 +10,8 @@ from porcupine import get_main_window, utils
 from porcupine.plugins.run import no_terminal, settings, terminal
 
 
-def copy_file_with_modification(source, destination, old_string, new_string):
-    content = source.read_text()
-    assert content.count(old_string) == 1
-    assert not destination.exists()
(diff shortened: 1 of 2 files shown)


tests/test_run_plugin.py
```diff
@@ -1,7 +1,72 @@
...
+from typing import Any, List
 
+import pytest
 
-def test_no_previous_command_error(filetab, tmp_path, mocker):
+from porcupine import get_main_window, utils
+from porcupine.plugins.run import no_terminal, settings, terminal
+
+
+def copy_file_with_modification(source, destination, old_string, new_string):
```
Akuli, line 12: not used

tests/test_run_plugin.py
```diff
@@ -1,7 +1,72 @@
...
+    content = source.read_text()
+    assert content.count(old_string) == 1
+    assert not destination.exists()
+    destination.write_text(content.replace(old_string, new_string))
+
+
+@pytest.fixture
+def isolated_history():
```
Akuli, line 20: `autouse=True` would be good

Issue #303, opened 2021-02-08
Closing tabs with <Button-2> on Mac


Akuli:
Found while grepping for `TODO.*Mac` in #297 

On Windows and Linux, tabs can be closed by moving the mouse pointer over them (not necessarily over the `X` button, it also works where it shows the file name) and pressing the mouse wheel. I don't know whether that works on Mac. I don't even know whether macs have a mouse wheel.

Akuli:
Looking at [this code](https://github.com/ThePhilgrim/minesweeper/blob/e9241de6dfcb10231cf48b7ed01adf853d1a47a9/minesweeper.py#L237), it seems like the tabs should close when they are right-clicked. @ThePhilgrim Can you confirm that this works?

Akuli:
```
<Akuli> is closing with right-click a sane thing to do at all?
<Akuli> on windows and linux, it closes with wheel-click, which seems to be more or less standard
<ThePhilgrim> nah, I would accidentally close it when trying to access some kind of options or something
```

1anakin20:
In mac the bind for middle mouse click is `<Button-3>`. I can update the default key binds accordingly.

Akuli:
Ah, so `<Button-2>` and `<Button-3>` are basically backwards on Mac vs other operating systems. Thanks, good to know!

The new key binding could be very similar to `<<RightClick>>`.

PR #818, opened 2021-10-03
mac mouse wheel click

1anakin20:
I have replaced the `TabClosing:HeaderClickClose` by `<<WheelClick>>` binding who will set the mouse button to 2 if it is mac, else 3.
The tab closing plugin has been updated accordingly.
# Referenced issues
#303


Akuli (approved):
Thanks!

Issue #813, opened 2021-10-03
cannot copy/paste from passive text widgets


Akuli:
Would be nice to copy/paste output of running command within porcupine window. Affects everything that uses `textutils.create_passive_text_widget()`.

rdbende:
I can't figure out, how to do it. Can't bind it as it can't be in focus

Akuli:
Apparently `takefocus=True` is enough. For example:

```
>>> from porcupine.textutils import create_passive_text_widget
>>> t = create_passive_text_widget(None)
>>> t.pack()
>>> t['state'] = 'normal'
>>> t.insert('end', 'asdf')
>>> t['state'] = 'disabled'
>>> import tkinter
>>> tkinter.Entry().pack()
>>> t['takefocus'] = True
```

After running the last line, it is possible to switch between the entry and the text widget using tab. It is also possible to copy the text.

To get Ctrl+A to select all in the text widget, like you would expect it to do, you can load default_keybindings.tcl:

```
>>> t.tk.eval('source porcupine/default_keybindings.tcl')
''
```

Akuli:
Turns out that the run plugin doesn't use `textutils.create_passive_text_widget`, but it should be changed to use it.

PR #822, opened 2021-10-04
Run dialog tweaks

rdbende:
Substitutions now appear in a list in a passive text widget, so you should be able to copy/paste them, AFTER #813

### DIFF
--- a/porcupine/plugins/run/dialog.py
+++ b/porcupine/plugins/run/dialog.py
@@ -109,7 +109,7 @@ class CommandAsker:
 
         sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
         sub_textbox = textutils.create_passive_text_widget(
-            content_frame, height=len(substitutions), width=1
+            content_frame, height=len(substitutions), width=1, wrap="none"
         )
         sub_textbox.pack(fill="x", padx=(15, 0), pady=(0, 20))
         sub_textbox.config(state="normal")
--- a/porcupine/plugins/run/dialog.py
+++ b/porcupine/plugins/run/dialog.py
@@ -105,13 +105,15 @@ class CommandAsker:
             validated_callback=self.update_run_button,
         )
 
-        sub_text = "\n".join("\t•  {%s} = %s" % pair for pair in substitutions.items())
+        sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
         sub_textbox = textutils.create_passive_text_widget(
-            content_frame, width=1, height=len(substitutions) + 2, tabs=10
+            content_frame, height=len(substitutions) + 2
         )
         sub_textbox.pack(fill="x")
         sub_textbox.config(state="normal")
         sub_textbox.insert("1.0", f"Substitutions:\n{sub_text}\n")
+        sub_textbox.tag_configure("indent", lmargin1=15)
+        sub_textbox.tag_add("indent", "2.0", "end")
         sub_textbox.config(state="disabled")
 
         # TODO: remember value with settings
--- a/porcupine/plugins/run/dialog.py
+++ b/porcupine/plugins/run/dialog.py
@@ -111,7 +111,7 @@ class CommandAsker:
         sub_textbox = textutils.create_passive_text_widget(
             content_frame, height=len(substitutions), width=1
         )
-        sub_textbox.pack(fill="x", padx=15, pady=(0, 20))
+        sub_textbox.pack(fill="x", padx=(15, 0), pady=(0, 20))
         sub_textbox.config(state="normal")
         sub_textbox.insert("1.0", sub_text)
         sub_textbox.config(state="disabled")
--- a/porcupine/plugins/run/dialog.py
+++ b/porcupine/plugins/run/dialog.py
@@ -205,6 +205,10 @@ def ask_command(tab: tabs.FileTab, project_path: Path) -> history.Command | None
     asker = CommandAsker(tab.path, project_path, history.get(tab, project_path))
     asker.window.title("Run command")
     asker.window.transient(get_main_window())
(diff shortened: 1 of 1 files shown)


porcupine/plugins/run/dialog.py
```diff
@@ -105,8 +105,14 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
             validated_callback=self.update_run_button,
         )
 
-        sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
-        ttk.Label(content_frame, text=f"Substitutions:\n{sub_text}\n").pack(fill="x")
```
Akuli, line 109: With a label, the width of the dialog is determined by the longest substitution text, or the width of "Display the output inside the Porcupine window (does not support keyboard input)", whichever is wider. This might be a good thing, because if you have very long file paths, then the dialog will be bigger, and they will always fit.

On the other hand, as you say, a text widget will hopefully allow copy/pasting in the future.

Which is better overall? What do you think?

porcupine/plugins/run/dialog.py
```diff
@@ -105,8 +105,14 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
...
+        sub_text = "\n".join("\t•  {%s} = %s" % pair for pair in substitutions.items())
```
Akuli, line 108: Instead of `\t•  `, I think it would be better to use padding on the left side to indent the items, and display no black dots. That way it will be very easy to copy/paste the substitution text into Porcupine.

porcupine/plugins/run/dialog.py
```diff
@@ -105,8 +105,14 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
...
+        sub_textbox = textutils.create_passive_text_widget(
+            content_frame, height=len(substitutions) + 2, tabs=10
```
Akuli, line 110: This makes the dialog a bit too wide to my taste. It looks better when the width is determined from the radio button texts.
```suggestion
            content_frame, width=1, height=len(substitutions) + 2, tabs=10
```

porcupine/plugins/run/dialog.py
rdbende: If the user has that long file paths, he deserves to resize the window to see it properly. So I vote for the text widget.

porcupine/plugins/run/dialog.py
```diff
@@ -106,7 +106,15 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
         )
 
         sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
-        ttk.Label(content_frame, text=f"Substitutions:\n{sub_text}\n").pack(fill="x")
+        sub_textbox = textutils.create_passive_text_widget(
+            content_frame, height=len(substitutions) + 2
```
Akuli, line 110: Looks like you deleted `width=1` by accident?

porcupine/plugins/run/dialog.py
```diff
@@ -106,7 +106,15 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
...
+        )
+        sub_textbox.pack(fill="x")
+        sub_textbox.config(state="normal")
+        sub_textbox.insert("1.0", f"Substitutions:\n{sub_text}\n")
+        sub_textbox.tag_configure("indent", lmargin1=15)
+        sub_textbox.tag_add("indent", "2.0", "end")
```
Akuli, line 116: It might be easier to use a separate label for the initial `"Substitutions:"`, since that part doesn't need to be copy/pasteable.

porcupine/plugins/run/dialog.py
rdbende: I think I messed up something with git

porcupine/plugins/run/dialog.py
```diff
@@ -105,8 +105,16 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
...
         )
 
+        ttk.Label(content_frame, text="Substitutions:").pack(anchor="w")
+
         sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
-        ttk.Label(content_frame, text=f"Substitutions:\n{sub_text}\n").pack(fill="x")
+        sub_textbox = textutils.create_passive_text_widget(
+            content_frame, height=len(substitutions), width=1
+        )
+        sub_textbox.pack(fill="x", padx=15, pady=(0, 20))
```
Akuli, line 114: For long file paths, it's better to have the padding only on the left side.
```suggestion
        sub_textbox.pack(fill="x", padx=(15, 0), pady=(0, 20))
```

porcupine/plugins/run/dialog.py
```diff
@@ -105,8 +105,16 @@ def __init__(self, file_path: Path, project_path: Path, suggestions: list[histor
             validated_callback=self.update_run_button,
         )
 
+        ttk.Label(content_frame, text="Substitutions:").pack(anchor="w")
+
         sub_text = "\n".join("{%s} = %s" % pair for pair in substitutions.items())
-        ttk.Label(content_frame, text=f"Substitutions:\n{sub_text}\n").pack(fill="x")
+        sub_textbox = textutils.create_passive_text_widget(
+            content_frame, height=len(substitutions), width=1
```
Akuli, line 112: Currently it is always 7 lines tall, but wraps the text into more than 7 lines when you have long file paths or the dialog is very narrow. This makes some lines at the end invisible. We can disable wrapping to fix this:

```suggestion
            content_frame, height=len(substitutions), width=1, wrap="none"
```

Akuli (approved):
Thanks!

PR #823, opened 2021-10-04
Pastebin dialog

rdbende:
just a small ui fix, because place is terrible

### DIFF
--- a/tests/test_pastebin_plugin.py
+++ b/tests/test_pastebin_plugin.py
@@ -152,10 +152,11 @@ def test_lots_of_stuff_with_localhost_termbin(filetab, monkeypatch, tabmanager, 
 
 
 def test_paste_error_handling(monkeypatch, caplog, mocker, tabmanager, filetab, dont_run_in_thread):
+    monkeypatch.setattr(pastebin_module, "DPASTE_URL", "ThisIsNotValidUrlStart://wat")
     mocker.patch("tkinter.messagebox.showerror")
 
-    get_main_window().tk.call("tk", "busy", "hold", get_main_window())
-    pastebin_module.pasting_done_callback(pastebin_module.DPaste(), tkinter.Toplevel(), False, "")
+    tabmanager.select(filetab)
+    get_main_window().event_generate("<<Menubar:Pastebin/dpaste.com>>")
 
     tkinter.messagebox.showerror.assert_called_once_with(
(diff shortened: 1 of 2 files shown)


Akuli:
The `errordialog()` function is a mess. It is called in one place only, and it has a couple features that are never used:
- Main window is always viewable when the function gets called, so `if get_main_window().winfo_viewable()` could be cleaned up.
- `monospace_text` is never None.

This is because in older Porcupines, the `errordialog()` function was in `porcupine/utils.py` and it was used for many other things too. I moved it to the pastebin plugin after converting other `errordialog()` calls to use `tkinter.messagebox`.

During a few years of using the pastebin plugin, I have never actually needed to read the error message it displays. The solution has always been to make sure I'm connected to internet or try a different pastebin.

So I think we should:
- Delete the whole `errordialog()` function
- Add a call like `messagebox.showerror("Pasting failed", "Check your internet connection or try a different pastebin.")`
- Write the full error message to `logging`, in case someone wants to see it
- Update `test_pastebin_plugin.py` (I think it mocks the `errordialog` function and checks it was called with the correct arguments)

tests/test_pastebin_plugin.py
```diff
@@ -152,21 +152,14 @@ def fake_wait_window(success_dialog):
...
 def test_paste_error_handling(monkeypatch, caplog, mocker, tabmanager, filetab, dont_run_in_thread):
-    monkeypatch.setattr(pastebin_module, "DPASTE_URL", "ThisIsNotValidUrlStart://wat")
-    mock = mocker.patch("porcupine.plugins.pastebin.errordialog")
+    mocker.patch("tkinter.messagebox.showerror")
 
-    tabmanager.select(filetab)
-    get_main_window().event_generate("<<Menubar:Pastebin/dpaste.com>>")
-    get_main_window().update()
+    get_main_window().tk.call("tk", "busy", "hold", get_main_window())
+    pastebin_module.pasting_done_callback(pastebin_module.DPaste(), tkinter.Toplevel(), False, "")
```
Akuli, line 158: I don't really like how a test about error handling has to know how exactly `pasting_done_callback` gets called, and how exactly the main window is made busy.

Is there a reason to rewrite this part of the test? Was there a problem with invoking `"<<Menubar:Pastebin/dpaste.com>>"`?

Akuli (approved):
I modified the test to be like I wanted it. The rest looks good. Thanks!

PR #826, opened 2021-10-05
Link cleanup

Akuli:
aboutdialog and run plugins had very similar code for creating links. Move it into `utils` and change it so that the same code can be used for both.

### DIFF
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -37,10 +37,7 @@ class NoTerminalRunner:
 
         self._cwd: Path | None = None  # can't pass data to callbacks when adding link
         self._link_manager = textutils.LinkManager(
-            self.textwidget,
-            filename_regex,
-            get_text=(lambda m: m.group(0)),
-            get_click_callback=self._get_link_opener,
+            self.textwidget, filename_regex, self._get_link_opener
         )
         self.textwidget.tag_config("link", underline=True)
 
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -27,7 +27,7 @@ filename_regex_parts = [
 filename_regex = "|".join(r"(?:" + part + r")" for part in filename_regex_parts)
 
 
-def open_file_with_line_number(self, path: Path, lineno: int) -> None:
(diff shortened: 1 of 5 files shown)


porcupine/plugins/run/no_terminal.py
```diff
@@ -26,59 +27,44 @@
...
+            return None
+        return partial(self._open_link, path, int(lineno))
+
+    def _open_link(self, path: Path, lineno: int) -> None:
+        tab = get_tab_manager().open_file(path)
+        if tab is not None:
+            tab.textwidget.mark_set("insert", f"{lineno}.0")
+            tab.textwidget.see("insert")
+            tab.textwidget.tag_remove("sel", "1.0", "end")
+            tab.textwidget.tag_add("sel", "insert", "insert lineend")
```
Akuli, line 63: Doesn't use self, consider moving outside class

porcupine/plugins/run/no_terminal.py
```diff
@@ -136,20 +122,13 @@ def _queue_handler(self) -> None:
 
         if messages:
             self.textwidget.config(state="normal")
-            for cwd, tag, output_line in messages:
+            for tag, output_line in messages:
                 if tag == "clear":
                     assert not output_line
+                    self._link_manager.delete_all_links()
                     self.textwidget.delete("1.0", "end")
```
Akuli, line 129:
```suggestion
                    self.textwidget.delete("1.0", "end")
                    self._link_manager.delete_all_links()  # prevent memory leak
```

tests/test_about_dialog.py
```diff
@@ -1,41 +1,33 @@
...
         nonlocal called
         called += 1
-        self.destroy()
+        self.destroy()  # can't do this with mock objects
 
-    monkeypatch.setattr(tkinter.Toplevel, "wait_window", fake_wait_window)
-
-    get_main_window().event_generate("<<Menubar:Help/About Porcupine>>")
+    monkeypatch.setattr("tkinter.Toplevel.wait_window", fake_wait_window)
+    aboutdialog.show_about_dialog()
```
Akuli, line 18:
```suggestion
    get_main_window().event_generate("<<Menubar:Help/About Porcupine>>")
```

PR #827, opened 2021-10-05
Passive text widget copy-paste

### LINKED ISSUE #813: cannot copy/paste from passive text widgets

rdbende:
Fixes #813

### DIFF
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -45,7 +45,9 @@ def parse_paths(cwd: Path, line: str) -> Iterator[tuple[str, Path, int] | str]:
 class NoTerminalRunner:
     def __init__(self, master: tkinter.Misc) -> None:
         # TODO: better coloring that follows the pygments theme
-        self.textwidget = create_passive_text_widget(master, height=12, name="run_output")
+        self.textwidget = create_passive_text_widget(
+            master, height=12, name="run_output", is_focusable=True
+        )
         self.textwidget.tag_config("info", foreground="blue")
         self.textwidget.tag_config("output")  # use default colors
         self.textwidget.tag_config("error", foreground="red")
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -41,7 +41,7 @@ class NoTerminalRunner:
     def __init__(self, master: tkinter.Misc) -> None:
         # TODO: better coloring that follows the pygments theme
         self.textwidget = create_passive_text_widget(
-            master, height=12, name="run_output", is_focusable=True
+            master, font="TkFixedFont", height=12, is_focusable=True, name="run_output"
         )
         self.textwidget.tag_config("info", foreground="blue")
         self.textwidget.tag_config("output")  # use default colors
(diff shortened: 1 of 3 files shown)


Akuli:
Can you also change the run plugin to actually use `create_passive_text_widget()`? Currently it manually sets `state="disabled"`, so this PR as is doesn't actually fix #813.

porcupine/plugins/run/no_terminal.py
```diff
@@ -12,7 +12,8 @@
 from pathlib import Path
 from typing import Callable
 
-from porcupine import get_tab_manager, images, textutils, utils
+from porcupine import get_tab_manager, images, utils
```
Akuli, line 15: textutils needed for `LinkManager`
```suggestion
from porcupine import get_tab_manager, images, textutils, utils
```

porcupine/plugins/run/no_terminal.py
```diff
@@ -39,7 +40,7 @@ def open_file_with_line_number(path: Path, lineno: int) -> None:
 class NoTerminalRunner:
     def __init__(self, master: tkinter.Misc) -> None:
         # TODO: better coloring that follows the pygments theme
-        self.textwidget = tkinter.Text(master, height=12, state="disabled", name="run_output")
+        self.textwidget = create_passive_text_widget(master, height=12, name="run_output")
```
Akuli, line 43: I would prefer a fixed-width (monospace) font, so prevent `create_passive_text_widget` from setting it to `TkDefaultFont` which is not fixed-width. For text widgets, `TkFixedFont` is the default.

```suggestion
        self.textwidget = create_passive_text_widget(master, height=12, name="run_output", font="TkFixedFont")
```

porcupine/textutils.py
```diff
@@ -790,6 +790,8 @@ def create_passive_text_widget(parent: tkinter.Widget, **kwargs: Any) -> tkinter
     kwargs.setdefault("relief", "flat")
     kwargs.setdefault("wrap", "word")
     kwargs.setdefault("state", "disabled")
+    kwargs.setdefault("takefocus", True)
+    kwargs.setdefault("highlightthickness", 0)
```
Akuli, line 794: I think setting `higlightthickness=0` and `takefocus=True` in the same text widget isn't great. You can select the widget by pressing Tab or Shift+Tab several times or by clicking it, but there's no visual indicator for it being selected. Whether it is selected or not is important for the user, because you can press Ctrl+A only in a selected widget, and you want to see where you are when navigating widgets by pressing Tab.

A better solution might be to delete these two lines:

```suggestion
```

So now, text widgets don't get focus by default, but we can add `takefocus=True` to select which widgets should be focused, and get the default visible border when focused.

`takefocus=True` is good for:
- The run output widget, so that you can copy/paste the output
- Substitutions in the "Run command" dialog, so that you can copy/paste them (#822)

`takefocus=True` is bad for:
- The two text widgets in "Run command" dialog that show the command after substituting it. I want to jump from the first entry to the second with a single Tab key press.
- The about dialog. There's no need to copy/paste its text, and a highlighted border would just look weird.

rdbende: yeah, that's a fair point i'll make modifications

Akuli (approved):
Thanks!

Issue #829, opened 2021-10-05
Overwriting currently opened file


Akuli:
Corner case not handled in #808

PR #831, opened 2021-10-05
if necessary, reload file once command completes

### LINKED ISSUE #819: refresh current file when command finished
Would be useful for running commands that modify currently opened file

Already works when using external terminal, because file refreshes when focusing porcupine window

Akuli:
Fixes #819 

Makes #829 easier to fix

### DIFF
--- a/porcupine/tabs.py
+++ b/porcupine/tabs.py
@@ -745,6 +745,15 @@ class FileTab(Tab):
             #   - Indicates that something isn't saved
             return True
 
+    # TODO: document this
+    def reload_if_necessary(self) -> None:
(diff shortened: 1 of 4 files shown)


porcupine/tabs.py
```diff
@@ -745,6 +745,15 @@ def other_program_changed_file(self) -> bool:
             #   - Indicates that something isn't saved
             return True
 
+    # TODO: document this
```
Akuli, line 748:
```suggestion
    # TODO: document this
    # TODO: should cursor and scrolling stuff be a part of reload()?
```

PR #832, opened 2021-10-05
create utils.tkinter_safe_string and use it in run plugin

### LINKED ISSUE #811: emoji output
In #809 I realized that running black with output going into Porcupine window doesn't work.

Akuli:
Fixes #811

### DIFF
--- a/tests/test_run_plugin.py
+++ b/tests/test_run_plugin.py
@@ -70,7 +70,7 @@ import sys
 
 # Test error handling for badly printed bytes
 # All bytes that are invalid utf-8 AND invalid cp1252: 81, 8D, 8F, 90, 9D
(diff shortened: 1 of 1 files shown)


porcupine/utils.py
```diff
@@ -335,6 +335,11 @@ def get_binding(virtual_event: str, *, menu: bool = False) -> str:
     return _format_binding(bindings[0], menu) if bindings else ""
 
 
+# TODO: document this
+def tkinter_safe_string(string: str) -> str:
+    return "".join(char for char in string if ord(char) <= 0xFFFF)
```
Akuli, line 340: It might be better to replace the unsupported chars with a "?" character.

PR #835, opened 2021-10-06
run plugin: source venv activate script before running commands

### LINKED ISSUE #804: venv support
The venv plugin should be able to prefix something like `source /blah/blah/venv/bin/activate` into shell commands

Akuli:
Fixes #804 

- [ ] test on windows

### DIFF
--- a/porcupine/plugins/run/__init__.py
+++ b/porcupine/plugins/run/__init__.py
@@ -1,22 +1,36 @@
 """Compile, run and lint files."""
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 from tkinter import messagebox
 from typing import Any, List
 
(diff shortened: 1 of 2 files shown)


porcupine/plugins/run/__init__.py
```diff
@@ -1,42 +1,58 @@
...
+        command_string = command.command
+    else:
+        if sys.platform == "win32":
+            activate = utils.quote(str(venv / "Scripts" / "activate"))
+            command_string = f"{activate}\n{command.command}"
+        else:
+            activate = utils.quote(str(venv / "bin" / "activate"))
+            command_string = f". {activate}\n{command.command}"
+
     # FIXME: python_venv plugin integration goes everywhere
```
Akuli, line 29:
```suggestion
```

PR #840, opened 2021-10-06
venv choosing in directory tree: refresh and allow unchecking

### LINKED ISSUE #836: directory tree doesn't refresh after switching venv

### LINKED ISSUE #839: can't unselect env
When a python venv is selected in directory tree, it shows up as grayed out. In some weird cases, it might be useful to use no venv at all, even if a project has a venv.

Akuli:
Fixes #836 
Fixes #839

### DIFF
--- a/porcupine/plugins/python_venv.py
+++ b/porcupine/plugins/python_venv.py
@@ -126,7 +126,7 @@ def _populate_menu(event: tkinter.Event[dirtree.DirectoryTree]) -> None:
 
     def on_change(*junk: object) -> None:
         set_venv(project_root, path if var.get() else None)
-        tree.refresh()  # needed only on windows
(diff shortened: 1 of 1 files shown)


porcupine/plugins/python_venv.py
```diff
@@ -118,23 +124,19 @@ def _populate_menu(event: tkinter.Event[dirtree.DirectoryTree]) -> None:
     if not is_venv(path):
         return
 
-    is_used = get_venv(project_root) == path
+    def on_change(*junk: object) -> None:
+        set_venv(project_root, path if var.get() else None)
+        tree.refresh()  # needed only on windows
```
Akuli, line 129:
```suggestion
        tree.refresh()  # needed on windows
```

Issue #527, opened 2021-06-17
directory tree: dotfiles are annoying


Akuli:
Similar to #399 

File names starting with `.` should be last, even in directories that don't have `.gitignore`. This is my home folder, viewed in Porcupine's directory tree:

![dotfiles](https://user-images.githubusercontent.com/18505570/122474160-8022b100-cfcb-11eb-84dc-e427cb95f111.png)

PR #852, opened 2021-10-08
Change directory tree order (dotfiles after normal files)

nicolafan:
Fixes: #527


Akuli:
It might be nicest to put dotted items last, regardless of whether they are directories or not. This way everything you will need is likely going to be in the beginning of the list.

nicolafan:
Schermata da 2021-10-08 13-25-13

PR #857, opened 2021-10-08
Implement cut-paste on right-click

nicolafan:
Makes #467 easier.


nicolafan:
The code has been simplified thanks to the fact that we want to cut-paste with rename if pasting is in the same dir, since the operations and checks were already implemented.

PR #896, opened 2021-10-10
messagebox `detail` argument

rdbende:


Akuli:
Test failures are a known problem and not related to your changes.

Issue #614, opened 2021-07-18
find plugin jumping


Akuli:
If I find matches in two different functions, I often want to replace them in just one of the functions. When I click "Replace this match" for the last time, the find area takes me to a different function and I have to navigate back to the function I was working on. I can use anchors and gotoline plugins to help with this, but it's still annoying.

PR #936, opened 2021-10-16
make find box take up less space

rdbende:
and other improvements

### DIFF
--- a/porcupine/plugins/find.py
+++ b/porcupine/plugins/find.py
@@ -27,23 +27,22 @@ class Finder(ttk.Frame):
         self._textwidget = textwidget
 
         # grid layout:
-        #         column 0        column 1     column 2        column 3
-        #     ,---------------------------------------------------------------.
-        # row0|     Find:     | text entry    |       | [x] Full words only   |
-        #     |---------------|---------------|-------|-----------------------|
-        # row1| Replace with: | text entry    |       | [x] Ignore case       |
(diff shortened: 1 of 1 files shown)


Akuli:
New and old find area:

finds

- This is not "less space". In fact, the new UI is a few pixels taller.
- The entries are now ridiculously short, so replacing longer pieces of code is going to be annoying.
- Why remove the dots? Maybe it's just me, but I expect status bars (and similar things) to display `.` at the end.
- I don't like the padding. This area of the editor should be squeezed to small space, so that it fits nicely on small laptop screens.

Unlike most other things in Porcupine, the find area is exactly like I want it. It's not something I want changed. Even though this mostly applies to how it behaves (`test_find_plugin.py` is huge), it also applies to the UI to some extent. So even though I appreciate your work and I have liked every pull request you made so far, this one just isn't good.

If you are interested in tweaking the find area, I would appreciate adding a "Replace all visible matches" button or some other solution to #614.

Akuli:
We can fix the unused space thing. The entries just have to be wide enough when Porcupine is as wide as in my screenshot. I usually have Porcupine and something else (e.g. web browser) next to each other, taking up half of the screen width each.

If the buttons really need vertical padding, please make it like 1 pixel or so. More horizontal padding than that is fine, even if there will be one more button.

It might be easier to make a new PR than revert everything else, but I can reopen this one if you want to keep what you have so far.

PR #938, opened 2021-10-16
fix cwd bug in git_right_click plugin

rdbende:
Fixes that you can't do right-click git stuff outside of the Porcupine repo


Akuli:
Ideally there would be a test for this, but it would be more complicated than just fixing this.

rdbende:
Yes :)

Issue #893, opened 2021-10-10
Refresh tree doesn't work for empty folder


nicolafan:
To reproduce the problem:

1. Copy or cut a file
2. Paste it inside a folder which is empty (identified by `(empty)`)

The file won't appear in the directory tree until it gets refreshed manually (by closing and opening the folder for example).

nicolafan:
I think the second condition on this line could have something to do with this issue.
https://github.com/Akuli/porcupine/blob/f0f4c7eb5270dbf20abcfaee296d91d6f84c5368/porcupine/plugins/directory_tree.py#L259
Removing the condition (`not self.contains_dummy(child_id)`) makes porcupine crash, so it's not a sol.

PR #945, opened 2021-10-17
Change refreshing condition

nicolafan:
Fixes: #893 

@Akuli can you review this? It seems to work.
After looking at the code I think that the whole point of that condition was to identify if the directory to refresh was open by checking the dummy child. This seems to solve the problem without having to specify the dummy type. I don't know if there's something that I'm not taking into account.

### DIFF
--- a/tests/test_directory_tree_plugin.py
+++ b/tests/test_directory_tree_plugin.py
@@ -212,7 +212,7 @@ def test_cycling_through_items(tree, tmp_path, tabmanager):
 
 def test_empty_directory_refreshing(tree, tmp_path):
     new_dir_path = tmp_path / "a"
-    new_dir_path.mkdir(parents=True)
+    new_dir_path.mkdir()
 
     tree.add_project(tmp_path)
     [project_id] = [id for id in tree.get_children("") if get_path(id) == tmp_path]


nicolafan:
Mh they're not perfectly equal: the dummy is present only before the dir is loaded the first time, after that the directory will have all its content in the tree so in next refreshes it will be refreshed even if it's close... At this point, maybe, checking if it's open will even reduce the refreshing time since closed directory won't be refreshed (should we refresh them?)

Akuli:
Not refreshing opened directories is good. When you open them, a `<<TreeviewOpen>>` event is generated, and there's a binding that calls `_open_and_refresh_directory` when that happens.

Can you add a test to `test_directory_tree.py`? Something like this:
1. Make an empty directory
2. Open it using `open_as_if_user_clicked()`
3. Create a file inside the directory
4. Refresh
5. Assert that the directory now has one child item whose path is what you would expect it to be (this is enough to ensure it is not a dummy item)

nicolafan:
I ran the test as example and it was passed

Akuli:
It also fails if you run it without the fix, which is good.

nicolafan:
Thank you for your help!

PR #951, opened 2021-10-17
HTML autoindent

### LINKED ISSUE #841: HTML autoindent
it'd be nice if the `autoindent` plugin would work with HTML, it's very annoying, that I can't use it there

rdbende:
This is absolutely enough

Fixes #841

### DIFF
--- a/tests/test_indent_dedent.py
+++ b/tests/test_indent_dedent.py
@@ -270,6 +270,9 @@ blah
 <p>
 blah blah
 </p>
+<div>
+<a>lol</a>
+</div>
 </body>
 </html>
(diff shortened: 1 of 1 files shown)


Akuli:
Thanks! Can you add a test to `tests/test_indent_dedent.py`?

rdbende:
Yeah, my regex absolutely not working.

```html
<img src="foo.png">
  <div>
```

It indents after self-closing tags, regex would be really complicated

Akuli:
You can explicitly close the tags, like ``. I think it would be enough to support that.

PR #957, opened 2021-10-18
Add Open in webbrowser plugin

rdbende:
I find this very useful

### DIFF
--- a/porcupine/plugins/open_in_browser.py
+++ b/porcupine/plugins/open_in_browser.py
@@ -14,7 +14,7 @@ from porcupine import menubar, tabs
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


Akuli:
I personally see tool plugins like this as deprecated, since you can now very easily run `python3 -c 'webbrowser.open("{file_path}")'` or `x-www-browser {file_path}` with the run plugin.

rdbende:
Yeah, you can. I just haven't thought of that so far :)

rdbende:
But I think the other 3 shortcuts should also have a default command (like `mypy`, `black`, `isort`, `./run` (since F5 is compile for that langs)), and then we can get rid of  `black` and `isort` plugins.

PR #991, opened 2021-10-26
beginner friendliness

rdbende:
Not sure if actually useful, but a nice touch.

### DIFF
--- a/porcupine/pluginloader.py
+++ b/porcupine/pluginloader.py
@@ -187,7 +187,10 @@ def _run_setup_and_set_status(info: PluginInfo) -> None:
     if not hasattr(info.module, "setup"):
         info.status = Status.SETUP_FAILED
         info.error = "missing"
-        log.warning(f"{info.name} plugin has no setup(). See https://akuli.github.io/porcupine/plugin-intro.html for more.")
+        log.warning(
+            f"{info.name} plugin has no setup(). See"
+            " https://akuli.github.io/porcupine/plugin-intro.html for more."
+        )
         return
 
     start = time.perf_counter()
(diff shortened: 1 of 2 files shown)


porcupine/plugins/pluginmanager.py
```diff
@@ -190,7 +190,18 @@ def _on_select(self, junk: object = None) -> None:
...
                 text = f"Importing the plugin failed.\n\n{info.error}"
             elif info.status == pluginloader.Status.SETUP_FAILED:
-                text = f"The plugin's setup() function failed.\n\n{info.error}"
+                text = "Calling the plugin's setup() function failed.\n\n"
+                if info.error is None:
+                    # no setup(). probably a beginner tries to make a plugin
+                    text += (
+                        "There is no setup() function. Make sure to include a setup function into"
+                        " your plugin.\n\nTo learn more about Porcupine's plugin API, visit"
+                        " https://akuli.github.io/porcupine"
```
Akuli, line 199: Why not just set this text to `info.error`?

It might be simplest to just put all this text to `log.error()` in `pluginloader.py`, so that the same exact text shows up in the terminal and in the plugin manager.

Akuli (commented):
It's finally starting to look short enough for something that most people will never notice :)

porcupine/pluginloader.py
```diff
@@ -183,6 +183,16 @@ def _run_setup_and_set_status(info: PluginInfo) -> None:
...
 
+    if not hasattr(info.module, "setup"):
+        info.status = Status.SETUP_FAILED
+        info.error = (
+            "There is no setup() function. Make sure to include a setup function into your"
+            " plugin.\nTo learn more about Porcupine's plugin API, visit"
+            " https://akuli.github.io/porcupine/plugin-intro.html"
+        )
+        log.warning(f"Calling {info.name!r} plugin's setup() function failed.\n{info.error}")
+        return
```
Akuli, line 194: This early-return skips code that removes a logging handler. It would be better to remove it, even though a plugin without a `setup()` likely won't log much.

Akuli (approved):
Thanks!

PR #994, opened 2021-10-30
New file option when right-clicking existing file

### LINKED ISSUE #993: new file when right-clicking existing file
If I right-click `foo/bar/baz.py`, it would be nice to get an option to create a new file, so that entering `blah.py` as file name creates `foo/bar/blah.py`

rdbende:
Fixes #993

### DIFF
--- a/porcupine/plugins/filemanager.py
+++ b/porcupine/plugins/filemanager.py
@@ -329,12 +329,12 @@ def get_selected_path(tree: DirectoryTree) -> Path | None:
 class Command:
     name: str
     virtual_event_name: str | None
-    condition: Callable[[Path], bool]
+    condition: Callable[[Path], bool] | None
     callback: Callable[[Path], None | str]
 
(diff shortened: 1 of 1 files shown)


porcupine/plugins/filemanager.py
```diff
@@ -360,7 +363,7 @@ def new_file_here(path: Path) -> None:
     # Doing something to an entire project is more difficult than you would think.
     # For example, if the project is renamed, venv locations don't update.
     # TODO: update venv locations when the venv is renamed
-    Command("New file", "<<FileManager:New file>>", (lambda p: p.is_dir()), new_file_here),
+    Command("New file here", "<<FileManager:New file>>", None, new_file_here),
```
Akuli, line 366: This way you don't need to add `None` checks everywhere else:

```suggestion
    Command("New file here", "<<FileManager:New file>>", (lambda p: True), new_file_here),
```

Akuli (approved):
Thanks!

Issue #911, opened 2021-10-12
right-clicking (empty) dummy


Akuli:
1. Create empty folder
2. Click it open in directory tree
3. Right-click the `(empty)` dummy item (or press Ctrl+C, or Ctrl+V, or do anything else with it)

You get an error from trying to get the path of the dummy item.

I think the correct fix is to check for dummy items in `get_path()`, and return None when the item is a dummy. This must be done in a way that doesn't slow down `get_path()`, so maybe like this:

```python
try:
    item_type, project_number, path = item_id.split(":", maxsplit=2)
except ValueError:
    ensure its a dummy item, raising error otherwise
    return None
```

PR #1004, opened 2021-12-15
Add check on get_path

nicolafan:
Fixes: #911 

It will raise many `mypy` errors. I did the simplest thing by adding the `None` checks to not have errors in the console... But I'm not sure about adding so many asserts together with the checks.

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -51,11 +51,9 @@ def get_path(item_id: str) -> Path | None:
     try:
         item_type, project_number, path = item_id.split(":", maxsplit=2)
     except ValueError as e:
-        # We found a dummy which doesn't represent any file or directory
+        # This should happen only if we found a dummy which doesn't represent any file or directory.
         if "dummy" not in get_directory_tree().item(item_id)["tags"]:
-            # But wait, it wasn't supposed to be there!
             raise e
-        # Otherwise, the dummy isn't anything to write home about.
         return None
     return Path(path)
 
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -51,8 +51,11 @@ def get_path(item_id: str) -> Path | None:
     try:
         item_type, project_number, path = item_id.split(":", maxsplit=2)
     except ValueError as e:
+        # We found a dummy which doesn't represent any file or directory
         if "dummy" not in get_directory_tree().item(item_id)["tags"]:
+            # But wait, it wasn't supposed to be there!
             raise e
+        # Otherwise, the dummy isn't anything to write home about.
         return None
     return Path(path)
 
@@ -236,7 +239,8 @@ class DirectoryTree(ttk.Treeview):
     def _hide_old_projects(self, junk: object = None) -> None:
         for project_id in self.get_children(""):
             project_path = get_path(project_id)
-            if project_path is None or not project_path.is_dir():
+            assert project_path is not None
+            if not project_path.is_dir():
                 self.delete(project_id)
 
         # To avoid getting rid of existing projects when not necessary, we do
@@ -258,11 +262,11 @@ class DirectoryTree(ttk.Treeview):
         self._hide_old_projects()
         self.event_generate("<<RefreshBegins>>")
         for project_id in self.get_children():
-            self._update_tags_and_content(get_path(project_id), project_id)
+            self._update_tags_and_content(project_id)
 
     # The following two methods call each other recursively.
 
-    def _update_tags_and_content(self, project_root: Path|None, child_id: str) -> None:
+    def _update_tags_and_content(self, child_id: str) -> None:
         if child_id.startswith(("dir:", "project:")) and self.item(child_id, "open"):
             self._open_and_refresh_directory(child_id)
 
@@ -299,7 +303,7 @@ class DirectoryTree(ttk.Treeview):
         project_id = self.find_project_id(dir_id)
         project_root = get_path(project_id)
         for child_path, child_id in path2id.items():
-            self._update_tags_and_content(project_root, child_id)
+            self._update_tags_and_content(child_id)
         self.sort_folder_contents(dir_id)
 
         if not self.get_children(dir_id):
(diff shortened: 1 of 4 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -47,8 +47,13 @@
...
 #   - type is "file", "dir", "project"
 #   - project_number is unique to each project
-def get_path(item_id: str) -> Path:
-    item_type, project_number, path = item_id.split(":", maxsplit=2)
+def get_path(item_id: str) -> Path | None:
+    try:
+        item_type, project_number, path = item_id.split(":", maxsplit=2)
+    except ValueError:
+        if "dummy" not in get_directory_tree().item(item_id)["tags"]:
+            raise ValueError
```
Akuli, line 55: Nit: It might be best to re-raise the same error. This doesn't really matter though, because if you raise a new erro, the error message will contain both errors.

```suggestion
    except ValueError as e:
        if "dummy" not in get_directory_tree().item(item_id)["tags"]:
            raise e
```

Akuli:
Thanks! I can go through some of the mypy failures myself.

porcupine/plugins/directory_tree.py
nicolafan: Oh I didn't know about this possibility. Thanks!

Moosems:
Any progress?

Akuli:
I will get back to this and Porcupine in general "eventually" :)

Akuli:
Applying a patch to fix mypy errors that my friend wrote, will then review...

porcupine/plugins/directory_tree.py
```diff
@@ -47,8 +47,13 @@
...
 #   - project_number is unique to each project
-def get_path(item_id: str) -> Path:
-    item_type, project_number, path = item_id.split(":", maxsplit=2)
+def get_path(item_id: str) -> Path | None:
+    try:
+        item_type, project_number, path = item_id.split(":", maxsplit=2)
+    except ValueError as e:
+        if "dummy" not in get_directory_tree().item(item_id)["tags"]:
+            raise e
+        return None
```
Akuli, line 56:
```suggestion
        # It is a "(dummy)" item that does not represent a file or directory
        return None
```

porcupine/plugins/directory_tree.py
```diff
@@ -229,7 +235,8 @@ def project_has_open_filetabs(self, project_id: str) -> bool:
 
     def _hide_old_projects(self, junk: object = None) -> None:
         for project_id in self.get_children(""):
-            if not get_path(project_id).is_dir():
+            project_path = get_path(project_id)
+            if project_path is None or not project_path.is_dir():
```
Akuli, line 239: It cannot be `None`.

```suggestion
            assert project_path is not None
            if not project_path.is_dir():
```

porcupine/plugins/directory_tree.py
```diff
@@ -255,12 +262,13 @@ def refresh(self, junk: object = None) -> None:
 
     # The following two methods call each other recursively.
 
-    def _update_tags_and_content(self, project_root: Path, child_id: str) -> None:
+    def _update_tags_and_content(self, project_root: Path|None, child_id: str) -> None:
```
Akuli, line 265: It would be better to assert that it isn't `None` when calling this method. Every file belongs to some project and so the project path is never `None`.

porcupine/plugins/git_status.py
```diff
@@ -147,13 +147,17 @@ def _handle_queue(self) -> None:
             if not self.tree.contains_dummy(dir_id):
                 tags_changed = False
                 for item_id in self.tree.get_children(dir_id):
-                    if self._set_tag(item_id, self._choose_tag(get_path(item_id))):
-                        tags_changed = True
+                    item_path = get_path(item_id)
+                    if item_path is not None:
+                        if self._set_tag(item_id, self._choose_tag(item_path)):
+                            tags_changed = True
```
Akuli, line 153: This is already under `if not self.tree.contains_dummy(dir_id)`, so we don't get dummy items here.

```suggestion
                    assert item_path is not None
                    if self._set_tag(item_id, self._choose_tag(item_path)):
                        tags_changed = True
```

porcupine/plugins/python_venv.py
```diff
@@ -120,11 +122,16 @@ def _populate_menu(event: tkinter.Event[dirtree.DirectoryTree]) -> None:
     tree: dirtree.DirectoryTree = event.widget
     [item] = tree.selection()
     path = dirtree.get_path(item)
+
+    if path is None:
+        return
+
     project_root = dirtree.get_path(tree.find_project_id(item))
-    if not is_venv(path):
+    if project_root is None or not is_venv(path):
```
Akuli, line 130: - `project_root` can't be `None`, as a project item is never a dummy item.
- We can do the `is_venv()` check earlier, as in `if path is None or not is_venv(path)`

porcupine/plugins/python_venv.py
```diff
@@ -120,11 +122,16 @@ def _populate_menu(event: tkinter.Event[dirtree.DirectoryTree]) -> None:
...
     [item] = tree.selection()
     path = dirtree.get_path(item)
+
+    if path is None or not is_venv(path):
+        return
+
     project_root = dirtree.get_path(tree.find_project_id(item))
-    if not is_venv(path):
+    if project_root is None:
         return
```
Akuli, line 131:
```suggestion
    assert project_root is not None
```

porcupine/plugins/directory_tree.py
```diff
@@ -47,8 +47,16 @@
...
-    item_type, project_number, path = item_id.split(":", maxsplit=2)
+def get_path(item_id: str) -> Path | None:
+    try:
+        item_type, project_number, path = item_id.split(":", maxsplit=2)
+    except ValueError as e:
+        # We found a dummy which doesn't represent any file or directory
+        if "dummy" not in get_directory_tree().item(item_id)["tags"]:
+            # But wait, it wasn't supposed to be there!
+            raise e
+        # Otherwise, the dummy isn't anything to write home about.
```
Akuli, line 58:
```suggestion
        # This should happen only if we found a dummy which doesn't represent any file or directory.
        if "dummy" not in get_directory_tree().item(item_id)["tags"]:
            raise e
```

Akuli:
Thanks nicolafan and Tuomas!

PR #1006, opened 2021-12-16
upgrade mypy

Akuli:


porcupine/plugins/directory_tree.py
```diff
@@ -108,7 +108,7 @@ def _on_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
 
         # Man page says identify_row is "obsolescent" but tkinter doesn't have the new thing yet
         item = self.identify_row(event.y)
-        if item is None:
+        if not item:  # TODO: check if typeshed is correct here
```
Akuli, line 111: must check

porcupine/plugins/git_status.py
```diff
@@ -229,7 +229,7 @@ def update_tree_selection_color(tree: DirectoryTree, event: object = None) -> No
 
 
 def sorting_key(tree: DirectoryTree, item_id: str) -> Any:
-    [git_tag] = [t for t in tree.item(item_id, "tags") if t.startswith("git_")] or [None]
+    git_tag = next((t for t in tree.item(item_id, "tags") if t.startswith("git_")), None)
```
Akuli, line 232: Ideally this would error if there's multiple `git_foo` tags

PR #1021, opened 2022-01-15
Add menu key functionality to directory tree

### LINKED ISSUE #1020: "Right-clicking" without mouse
On my keyboard, next to right Control, there's a key that opens up the right-click menu in most places. For example, I just pressed it while writing this text and selected "Undo". It deleted most of what I wrote.

In Porcupine's directory tree, the key does nothing, and I have to reach the mouse instead.

Kawaritai:
Resolves #1020 

(first time contributing, any feedback is welcome)

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -387,19 +387,22 @@ class DirectoryTree(ttk.Treeview):
         # Behaviour for menu/application key
         if menu_key:
             # Get relative position of focus item
-            x, y, width, height = self.bbox(self.focus())
+            try:
+                x, y, width, height = self.bbox(self.focus())
+            except ValueError:
+                x, y, width, height = (0, 0, 0, 0)
 
             # Apply menu position offsets
             menu_x = self.winfo_rootx() + x
             menu_y = self.winfo_rooty() + y + height
-        
+
         # Behaviour for a mouse right-click
         else:
             item = self.identify_row(event.y)
 
             # Update selected item
             self.set_the_selection_correctly(item)
-            
+
             # Set menu position to cursor position
             menu_x = event.x_root
             menu_y = event.y_root
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -392,6 +392,8 @@ class DirectoryTree(ttk.Treeview):
             except ValueError:
                 x, y, width, height = (0, 0, 0, 0)
 
+            width  # Silence pyflakes unused variable warning
+
             # Apply menu position offsets
             menu_x = self.winfo_rootx() + x
             menu_y = self.winfo_rooty() + y + height
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -381,24 +381,28 @@ class DirectoryTree(ttk.Treeview):
         self.contextmenu.delete(0, "end")
         self.event_generate("<<PopulateContextMenu>>")
 
-    def _on_right_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
+    def _on_right_click(self, event: tkinter.Event[DirectoryTree], menu_key: bool) -> str | None:
         self.tk.call("focus", self)
 
-        item: str = self.identify_row(event.y)
+        # Behaviour for menu/application key
+        if menu_key:
+            # Get relative position of focus item
+            x, y, width, height = self.bbox(self.focus())
 
+            # Apply menu position offsets
+            menu_x = self.winfo_rootx() + x
+            menu_y = self.winfo_rooty() + y + height
+        
         # Behaviour for a mouse right-click
-        if event.keysym != "App":
+        else:
+            item = self.identify_row(event.y)
+
             # Update selected item
             self.set_the_selection_correctly(item)
+            
             # Set menu position to cursor position
-            menu_x=event.x_root
-            menu_y=event.y_root
-
-        # Behaviour for menu/application key
-        else:
-            menu_x=self.winfo_rootx()
-            # Vertical menu offset by the focus item's relative position
-            menu_y=self.winfo_rooty()+self.bbox(self.focus())[1]+self.bbox(self.focus())[3]
+            menu_x = event.x_root
+            menu_y = event.y_root
 
         self._populate_contextmenu()
         if self.contextmenu.index("end") is not None:
@@ -473,9 +477,9 @@ def setup() -> None:
     tree.refresh()
 
     # Invoke context menu from right-click
-    tree.bind("<<RightClick>>", tree._on_right_click, add=True)
+    tree.bind("<<RightClick>>", partial(tree._on_right_click, menu_key=False), add=True)
     # Invoke context menu from menu key
-    tree.bind("<App>", tree._on_right_click, add=True)
+    tree.bind("<<MenuKey>>", partial(tree._on_right_click, menu_key=True), add=True)
 
     tree.bind("<Key>", tree._cycle_through_items, add=True)
 
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -388,7 +388,7 @@ class DirectoryTree(ttk.Treeview):
         if menu_key:
             # Get relative position of focus item
             try:
-                x, y, width, height = self.bbox(self.focus())
+                x, y, width, height = self.bbox(self.focus())  # type: ignore
             except ValueError:
                 x, y, width, height = (0, 0, 0, 0)
 
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -385,12 +385,25 @@ class DirectoryTree(ttk.Treeview):
         self.tk.call("focus", self)
 
         item: str = self.identify_row(event.y)
-        self.set_the_selection_correctly(item)
+
+        # Behaviour for a mouse right-click
+        if event.keysym != "App":
+            # Update selected item
+            self.set_the_selection_correctly(item)
+            # Set menu position to cursor position
+            menu_x=event.x_root
+            menu_y=event.y_root
+
+        # Behaviour for menu/application key
+        else:
+            menu_x=self.winfo_rootx()
+            # Vertical menu offset by the focus item's relative position
+            menu_y=self.winfo_rooty()+self.bbox(self.focus())[1]+self.bbox(self.focus())[3]
 
         self._populate_contextmenu()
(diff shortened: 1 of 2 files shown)


porcupine/plugins/directory_tree.py
```diff
@@ -459,8 +472,10 @@ def setup() -> None:
             tree.add_project(path, refresh=False)
     tree.refresh()
 
-    # TODO: invoking context menu from keyboard
+    # Invoke context menu from right-click
     tree.bind("<<RightClick>>", tree._on_right_click, add=True)
+    # Invoke context menu from menu key
+    tree.bind("<App>", tree._on_right_click, add=True)
```
Akuli, line 478: The correct key binding seems to be platform specific. On my computer (Linux), I get:

```
porcupine.pluginloader ERROR: directory_tree.setup() doesn't work
  File "/home/akuli/porcu/porcupine/pluginloader.py", line 190, in _run_setup_and_set_status
    info.module.setup()
  File "/home/akuli/porcu/porcupine/plugins/directory_tree.py", line 478, in setup
    tree.bind("<App>", tree._on_right_click, add=True)
  File "/usr/lib/python3.9/tkinter/__init__.py", line 1392, in bind
    return self._bind(('bind', self._w), sequence, func, add)
  File "/usr/lib/python3.9/tkinter/__init__.py", line 1346, in _bind
    self.tk.call(what + (sequence, cmd))
_tkinter.TclError: bad event type or keysym "App"
```

and the correct binding seems to be `<Menu>` instead of `<App>`. 

Does `<Menu>` work on your computer? If it doesn't, we can define a platform-specific virtual event in `porcupine/default_keybindings.tcl`, just like how `<<RightClick>>` is defined.

porcupine/plugins/directory_tree.py
```diff
@@ -385,12 +385,25 @@ def _on_right_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
         self.tk.call("focus", self)
 
         item: str = self.identify_row(event.y)
```
Akuli, line 387: The `identify_row()` doesn't make sense when this is triggered with a key press, so it should be moved under `if event.keysym != "App"`. You can also delete `: str` while you're at it. It should no longer be needed with the latest version of mypy.

porcupine/plugins/directory_tree.py
```diff
@@ -385,12 +385,25 @@ def _on_right_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
...
+            self.set_the_selection_correctly(item)
+            # Set menu position to cursor position
+            menu_x=event.x_root
+            menu_y=event.y_root
+
+        # Behaviour for menu/application key
+        else:
+            menu_x=self.winfo_rootx()
+            # Vertical menu offset by the focus item's relative position
+            menu_y=self.winfo_rooty()+self.bbox(self.focus())[1]+self.bbox(self.focus())[3]
```
Akuli, line 401: - It might be good to offset `menu_x` as well. I think it would look a little better, and it would still work if I later decide to move the directory tree somewhere else than to the left edge of the window.
- While `self.bbox(self.focus())[3]` works, it isn't very readable. Consider something like `x, y, width, height = self.bbox(self.focus())`. This unpacks the returned tuple into separate variables.

Akuli (commented):
Thanks for contributing!

porcupine/plugins/directory_tree.py
Kawaritai: I use Windows 10, `<App>` works but not `<Menu>`. I'm unfamiliar with virtual events, but I tried the following:
```
    set contmand Command
    event add "<<RightClick>>" <Button-2>
    event add "<<RightClick>>" <Control-Button-1>
    event add "<<WheelClick>>" <Button-3>
} else {
    set contmand Control
    event add "<<RightClick>>" <Button-3>
    event add "<<WheelClick>>" <Button-2>
    event add "<<MenuKey>>" <App>                             <--- ADDED
}
```
with 

```suggestion
    tree.bind("<<MenuKey>>", tree._on_right_click, add=True)
```
And the feature continued to work on my computer. 
I could use help on where and how to correctly define the event on other platforms.

porcupine/plugins/directory_tree.py
Kawaritai: Thanks for the feedback. Something like, -? I'm not sure how it appears on non-Windows 10 platforms.
```suggestion
            # Get relative position of focus item
            x, y, width, height = self.bbox(self.focus())

            # Apply menu position offsets
            menu_x = int(self.winfo_rootx() + (width / 8))
            menu_y = int(self.winfo_rooty() + y + height)
```

porcupine/plugins/directory_tree.py
Kawaritai: Okay, here's the new change:
```suggestion
        # Behaviour for a mouse right-click
        if event.keysym != "App":
            item = self.identify_row(event.y)
```

PS I'm still figuring out how to use GitHub right, and I'm not 100% sure on where to go from here...

porcupine/plugins/directory_tree.py
Akuli: Yeah, the Tcl code is a bit hard to read :)

The first line `if {[tk windowingsystem] == "aqua"} {` checks if Porcupine is running on MacOS. There are 3 possible values that `[tk windowingsystem]` can return: `aqua` (macos), `x11` (linux) and `win32` (windows). It's documented here: https://www.tcl.tk/man/tcl/TkCmd/tk.html#M12

So this if,else is for MacOS specific things. The `<App>` vs `<Menu>` difference isn't a MacOS thing though, because it differs between Windows and Linux. I put "macos keyboard" to google images search, and there doesn't seem to be a menu key on Mac.

I think we should just add a new if statement below, something like this:

```tcl
if {[tk windowingsystem] == "win32"} {
    event add "<<MenuKey>>" <App>
} elseif {[tk windowingsystem] == "x11"} {
    event add "<<MenuKey>>" <Menu>
}
```

Also, you aren't the first contributor who struggles with this Tcl code. Feel free to add a couple short comments to explain it for non-Tcl programmers.

porcupine/plugins/directory_tree.py
Akuli: It's ok to make a mess with Git and GitHub. I can help if you get stuck :)

Instead of checking this with `event.keysym`, we could pass a boolean into the function. Something like this (with lots of details omitted):

```
class DirectoryTree:
    def on_right_click(self, menu_key: bool, event):
        ...
        if menu_key:
            ...
        else:
            ...

tree.bind("<<RightClick>>", partial(tree._on_right_click, False), add=True)
tree.bind("<<MenuKey>>", partial(tree._on_right_click, True), add=True)
```

Now right-click will run `tree._on_right_click(False, event object)`, and the menu key will run `tree._on_right_click(True, event object)`.

porcupine/plugins/directory_tree.py
Akuli: We probably don't need to use the width in menu_x at all, because we want the menu to still appear in the left side of the selected item:

```
menu_x = self.winfo_rootx() + x
menu_y = self.winfo_rooty() + y + height
```

porcupine/plugins/directory_tree.py
Akuli: > I'm unfamiliar with virtual events

I have explained them here: https://akuli.github.io/porcupine/plugin-intro.html#virtual-events But that documentation is a bit outdated, because it doesn't mention that `default_keybindings.tcl` associates many virtual events with physical events (i.e. non-virtual events like `<App>` or `<Button-2>`).

Feel free to make a pull request that updates the documentation, or an issue that reminds me to update it. The documentation is somewhere in `docs/`. To find it, you can e.g. search for its title "Virtual Events" in all files, e.g. `git grep "Virtual Events"`.

Kawaritai: > Feel free to make a pull request that updates the documentation, or an issue that reminds me to update it.

Okay, I've made the issue #1028 for this.

TYSM for all your help! LMK about any other issues.

porcupine/plugins/directory_tree.py
```diff
@@ -381,16 +381,33 @@ def _populate_contextmenu(self) -> None:
...
-    def _on_right_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
+    def _on_right_click(self, event: tkinter.Event[DirectoryTree], menu_key: bool) -> str | None:
         self.tk.call("focus", self)
 
-        item: str = self.identify_row(event.y)
-        self.set_the_selection_correctly(item)
+        # Behaviour for menu/application key
+        if menu_key:
+            # Get relative position of focus item
+            x, y, width, height = self.bbox(self.focus())
```
Akuli, line 390: Currently the mypy check is failing with this mypy error:

```
porcupine/plugins/directory_tree.py:390: error: "Literal['']" object is not iterable  [misc]
```

This points out a real problem: if you have so many items in the treeview that not all of them fit on the screen, you select an invisible item, and then you press the menu key, you get an error:

```
    return self.func(*args)
  File "/home/akuli/porcu/porcupine/plugins/directory_tree.py", line 390, in _on_right_click
    x, y, width, height = self.bbox(self.focus())
ValueError: not enough values to unpack (expected 4, got 0)
```

In this case, `self.bbox()` returns an empty string, and so it's trying to do `x, y, width, height = ""`, and that fails. You need to add an if statement to check that.

Once you have fixed this, you need to run `black porcupine/plugins/directory_tree.py` and commit the result. Then this PR should be done :)

Kawaritai: Committed 4783f94d77caedfd01c711c5d9831de334cdf062.

BTW on a similar note, for me, right-clicking outside of a tree item raises errors such as:
```
  File "...\porcupine\porcupine\plugins\hide_project.py", line 13, in add_hide_option
    [project_id] = tree.selection()
ValueError: not enough values to unpack (expected 1, got 0)
```
```
  File ""...\porcupine\porcupine\plugins\python_venv.py", line 121, in _populate_menu
    [item] = tree.selection()
ValueError: not enough values to unpack (expected 1, got 0)
```
```
  File ""...\porcupine\porcupine\plugins\git_right_click.py", line 33, in populate_menu
    [item] = tree.selection()
ValueError: not enough values to unpack (expected 1, got 0)
```

These can be fixed with a try/except, as done in `porcupine\plugins\directory_tree.py` lines 318-322 and lines 358-361. 
Thoughts?

Akuli: Make a separate issue about your other errors :)

Akuli:
Thanks for working on this! I added a couple commits to silence warnings that don't indicate actual bugs. I'll merge this soon :)

PR #1024, opened 2022-01-16
Google text search feature added

### LINKED ISSUE #1018: Ctrl+G google searches the selected text
Would be handy, when better documentation doesn't show up in your langserver. We all use google a lot anyway...

okankamilsen:
PR for Issue resolve #1018

### DIFF
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -40,7 +40,7 @@ event add "<<Run:Repeat3>>" <F8>
 event add "<<Menubar:Edit/Go to Line>>" <$contmand-l>
 
 # google search plugin
-event add "<<Menubar:Tools/Google Search>>" <$contmand-g>
+event add "<<Menubar:Tools/Search selected text on Google>>" <$contmand-g>
 
 # fullscreen plugin
 event add "<<Menubar:View/Full Screen>>" <F11>
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -39,6 +39,9 @@ event add "<<Run:Repeat3>>" <F8>
 # gotoline plugin
 event add "<<Menubar:Edit/Go to Line>>" <$contmand-l>
 
+# google search plugin
+event add "<<Menubar:Tools/Google Search>>" <$contmand-g>
+
 # fullscreen plugin
 event add "<<Menubar:View/Full Screen>>" <F11>
 
--- a/porcupine/plugins/text_google_search.py
+++ b/porcupine/plugins/text_google_search.py
@@ -1,9 +1,10 @@
 """Search selected text on Google."""
 
 import tkinter as tk
-from porcupine import menubar, tabs
 import webbrowser
 import urllib.parse
+
+from porcupine import menubar, tabs
 
 def google_search(tab: tabs.FileTab) -> None:
     try:
--- a/porcupine/plugins/text_google_search.py
+++ b/porcupine/plugins/text_google_search.py
@@ -6,9 +6,12 @@ import webbrowser
 import urllib.parse
 
 def google_search(tab: tabs.FileTab) -> None:
-    selected_text = tab.textwidget.get("sel.first", "sel.last")
-    url = "https://www.google.com/search?q={}".format(urllib.parse.quote_plus(selected_text))
-    webbrowser.open_new_tab(url)
-
+    try:
+        selected_text = tab.textwidget.get("sel.first", "sel.last")
+        url = "https://www.google.com/search?q={}".format(urllib.parse.quote_plus(selected_text))
+        webbrowser.open_new_tab(url)
+    except:
+        pass
+    
 def setup() -> None:
     menubar.add_filetab_command("Tools/Search selected text on Google", google_search)
--- a/porcupine/plugins/text_google_search.py
+++ b/porcupine/plugins/text_google_search.py
@@ -1,0 +1,21 @@
+"""Search selected text on Google."""
+
+import tkinter as tk
+from porcupine import menubar, tabs
+import webbrowser
+
+def google_search(tab: tabs.FileTab) -> None:
+    try:
+        selected_text_tuple = tab.textwidget.tag_ranges(tk.SEL)
+        # Check if text is selected to search on Google
+        if len(selected_text_tuple)>0:
+            selected_text = tab.textwidget.get(*selected_text_tuple)
+            url = "https://www.google.com/search?q={}".format(selected_text)
+            webbrowser.open_new_tab(url)
+        else:
+            tk.messagebox.showwarning("Warning","Please select text!")
+    except:
+        tk.messagebox.showerror("Error", "Something went wrong!")
+
+def setup() -> None:
+    menubar.add_filetab_command("Tools/Google Search", google_search)
--- a/porcupine/plugins/text_google_search.py
+++ b/porcupine/plugins/text_google_search.py
@@ -5,17 +5,9 @@ from porcupine import menubar, tabs
 import webbrowser
 
 def google_search(tab: tabs.FileTab) -> None:
-    try:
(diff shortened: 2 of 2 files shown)


porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,21 @@
...
+
+import tkinter as tk
+from porcupine import menubar, tabs
+import webbrowser
+
+def google_search(tab: tabs.FileTab) -> None:
+    try:
+        selected_text_tuple = tab.textwidget.tag_ranges(tk.SEL)
+        # Check if text is selected to search on google
+        if len(selected_text_tuple)>0:
```
Akuli, line 11: We probably shouldn't do the search if multiple lines of text are selected, or if the selected text consists entirely of whitespace characters (e.g. spaces or tabs, basically whatever `string.strip()` removes).

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,21 @@
...
+            selected_text = tab.textwidget.get(*selected_text_tuple)
```
Akuli, line 12: Slightly simpler:

```suggestion
            selected_text = tab.textwidget.get("sel.first", "sel.last")
```

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,21 @@
...
+            url = "https://www.google.com/search?q={}".format(selected_text)
```
Akuli, line 13: Shouldn't this quote the text somehow? For example, if you google search for `std::vector<int>& values`, the URL in browser starts with `https://www.google.com/search?q=std%3A%3Avector%3Cint%3E%26+values`, not `https://www.google.com/search?q=std::vector<int>& values`.

I think you can use `urllib.parse.quote_plus()` for this:

```
>>> import urllib.parse
>>> urllib.parse.quote_plus('hello world&foo=bar')
'hello+world%26foo%3Dbar'
```

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,21 @@
...
+            webbrowser.open_new_tab(url)
+        else:
+            tk.messagebox.showwarning("Warning","Please select text!")
+    except:
+        tk.messagebox.showerror("Error", "Something went wrong!")
```
Akuli, line 18: What's the purpose of these message boxes? I think it would be better without them:
- If you accidentally press Ctrl+G when you didn't mean to google anything, you get a message box saying "Please select text!". How are you supposed to know that it was because Porcupine tried to do a Google search? It would be better to just not bother the user with unnecessary message boxes.
- I don't like "Something went wrong" error messages. They don't tell what went wrong. Also, the user will notice if something went wrong, even without message box, when their browser doesn't open. If you don't use `try`/`except` at all, Porcupine will print the full error message to terminal and write it to a log file, in case the user wants to figure out why Ctrl+G isn't working.

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,21 @@
...
+
+def setup() -> None:
+    menubar.add_filetab_command("Tools/Google Search", google_search)
```
Akuli, line 21:
```suggestion
    menubar.add_filetab_command("Tools/Search selected text on Google", google_search)
```

porcupine/default_keybindings.tcl
```diff
@@ -39,6 +39,9 @@ event add "<<Run:Repeat3>>" <F8>
 # gotoline plugin
 event add "<<Menubar:Edit/Go to Line>>" <$contmand-l>
 
+# google search plugin
+event add "<<Menubar:Tools/Google Search>>" <$contmand-g>
```
Akuli, line 43:
```suggestion
event add "<<Menubar:Tools/Search selected text on Google>>" <$contmand-g>
```

Akuli (commented):
Thanks! A few suggestions below.

porcupine/plugins/text_google_search.py
okankamilsen: Warnings are deleted in commit: https://github.com/Akuli/porcupine/pull/1024/commits/4ec19862ec918996f7091e86d98b887f01d5ecea

porcupine/plugins/text_google_search.py
okankamilsen: It has been simplified in commit https://github.com/Akuli/porcupine/pull/1024/commits/4ec19862ec918996f7091e86d98b887f01d5ecea

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,13 @@
+"""Search selected text on Google."""
+
+import tkinter as tk
+from porcupine import menubar, tabs
+import webbrowser
+
+def google_search(tab: tabs.FileTab) -> None:
+    selected_text = tab.textwidget.get("sel.first", "sel.last")
```
Akuli, line 8: This will error if you press Ctrl+G with no text selected. I think it should just do nothing.

porcupine/plugins/text_google_search.py
okankamilsen: Resolved in commit https://github.com/Akuli/porcupine/pull/1024/commits/0869a0e826a7438a3864e9cb4fff3b3346577310

porcupine/plugins/text_google_search.py
Akuli: It would also be good to avoid google-searching multiple lines of text, or text that consists of space characters only. Something like `if selected_text.strip() and "\n" not in selected_text:`

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,17 @@
...
+import webbrowser
+import urllib.parse
+
+def google_search(tab: tabs.FileTab) -> None:
+    try:
+        selected_text = tab.textwidget.get("sel.first", "sel.last")
+        url = "https://www.google.com/search?q={}".format(urllib.parse.quote_plus(selected_text))
+        webbrowser.open_new_tab(url)
+    except:
+        pass
```
Akuli, line 14: - It would be good to specify the exception being caught here, as in `except TclError:`.
- Avoid putting many lines of code inside try/except. Only put the first line, because that's where the error will come from.

Something like this:

```
try:
    selected_text = ...
except tk.TclError:
    return

url = ...
webbrowser.open_new_tab(url)
```

Akuli: It might be good to also add a comment like `# nothing selected` or similar. It won't be obvious for someone else looking at the code :)

porcupine/plugins/text_google_search.py
okankamilsen: resolved in commit https://github.com/Akuli/porcupine/pull/1024/commits/b7105b4ff877b0fef23a4df394a50a879c1c701e

porcupine/plugins/text_google_search.py
okankamilsen: resolved in commit https://github.com/Akuli/porcupine/pull/1024/commits/b7105b4ff877b0fef23a4df394a50a879c1c701e

porcupine/plugins/text_google_search.py
```diff
@@ -0,0 +1,23 @@
...
+
+def google_search(tab: tabs.FileTab) -> None:
+    try:
+        selected_text = tab.textwidget.get("sel.first", "sel.last")
+    except tk.TclError:
+        # nothing selected
+        return
+    
+    # Check multi line or text with only spaces
+    if selected_text.strip() and "\n" not in selected_text and not selected_text.isspace():
```
Akuli, line 17: I don't think you need both `.strip()` and `.isspace()`. One of the two is enough.

Akuli:
```suggestion
    if "\n" not in selected_text and not selected_text.isspace():
```

okankamilsen: .strip() kept .isspace() deleted.

Akuli (approved):
Thanks! I like how this works now :)

Issue #1042, opened 2022-04-09
More right-click options


Akuli:
In the directory tree, after right-clicking a directory it would be nice to have an option "Open in terminal", instead of having to open a new terminal and `cd` to the directory. It's already possible to run commands in a terminal, and we can probably reuse much of the code.

~~It would be also nice to have a "Copy full path to clipboard" option, so that I can paste the path into some other program instead of browsing or typing it manually.~~ (this part is now done)

PR #1043, opened 2022-04-09
"Copy full path to clipboard" right click option

ihammadasghar:
Resolve part of #1042

Since this is my first time contributing to this repo, after studying the code base & docs here is what I came up with.

Added the following features:
- "Copy full path to clipboard": made use of tkinter's clipboard functions, also used the is_dir() validation which was being used previously with other options.
- "Open in terminal": used the subprocess module to open a new terminal with the directory that was chosen.

I coded and tested this on my windows 11 machine.

### DIFF
--- a/porcupine/plugins/filemanager.py
+++ b/porcupine/plugins/filemanager.py
@@ -389,7 +389,7 @@ commands = [
     Command("Delete", "<<FileManager:Delete>>", is_NOT_project_root, delete),
     Command("Open in file manager", None, (lambda p: p.is_dir()), open_in_file_manager),
     Command("Open in terminal", None, (lambda p: p.is_dir()), open_in_terminal),
-    Command("Copy full path to clipboard", None, (lambda p: p.is_dir()), copy_full_path_to_clipboard),
+    Command("Copy full path to clipboard", None, (lambda p: True), copy_full_path_to_clipboard),
 ]
 
 
--- a/porcupine/plugins/filemanager.py
+++ b/porcupine/plugins/filemanager.py
@@ -316,17 +316,6 @@ def open_in_file_manager(path: Path) -> None:
     subprocess.Popen([opener_command, str(path)])
 
 
-def open_in_terminal(path: Path) -> None:
-    # Using pushd for windows to change drive aswell navigate to directory
-    windowingsystem = get_main_window().tk.call("tk", "windowingsystem")
-    if windowingsystem == "win32":
-        command = "pushd"
-    else:
-        command = "cd"
-
-    subprocess.Popen(f"start /wait {command} {path}", shell=True)
-
-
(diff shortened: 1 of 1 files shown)


porcupine/plugins/filemanager.py
```diff
@@ -371,6 +388,8 @@ def new_file_here(path: Path) -> None:
     Command(f"Move to {trash_name}", "<<FileManager:Trash>>", is_NOT_project_root, trash),
     Command("Delete", "<<FileManager:Delete>>", is_NOT_project_root, delete),
     Command("Open in file manager", None, (lambda p: p.is_dir()), open_in_file_manager),
+    Command("Open in terminal", None, (lambda p: p.is_dir()), open_in_terminal),
+    Command("Copy full path to clipboard", None, (lambda p: p.is_dir()), copy_full_path_to_clipboard),
```
Akuli, line 392: You should be able to copy any path whatsoever to clipboard, not just directories.
```suggestion
    Command("Copy full path to clipboard", None, (lambda p: True), copy_full_path_to_clipboard),
```

porcupine/plugins/filemanager.py
```diff
@@ -316,6 +316,23 @@ def open_in_file_manager(path: Path) -> None:
...
 
+def open_in_terminal(path: Path) -> None:
+    # Using pushd for windows to change drive aswell navigate to directory
+    windowingsystem = get_main_window().tk.call("tk", "windowingsystem")
+    if windowingsystem == "win32":
+        command = "pushd"
+    else:
+        command = "cd"
+
+    subprocess.Popen(f"start /wait {command} {path}", shell=True)
```
Akuli, line 327: This has a couple problems:
- I don't think it works if the path has spaces in it. For example `C:\Foo Bar` would turn into `start /wait pushd C:\Foo Bar`, which looks wrong to me.
- It doesn't work at all on macos and linux. There's no `start` command on these operating systems.

Starting a terminal is trickier than you might expect, and the `run` plugin already has code that deals with it. We should reuse it somehow, without copy/pasting anything of course, and I think that will be tricky.

You can just delete `open_in_terminal()` from this PR. You can then work on it in a separate PR, or just wait for me to eventually do it.

Akuli (commented):
Thanks!

porcupine/plugins/filemanager.py
ihammadasghar: I did test paths with spaces, there wasn't a problem on windows 11 at least.
I didn't consider the "start" command only working on windows, my bad. That does make it a bit trickier, but I'll look in to the run plugin code and see if I'm able to reuse it. Thanks!

Akuli:
You now need to run `black porcupine/plugins/filemanager.py`, and then this PR is done :)

Akuli (approved):
Thanks!

PR #1044, opened 2022-04-09
added the ability for a user to make a new directory

### LINKED ISSUE #1040: New directory here
In the file manager at left, you can right-click a file or folder and choose "New file here". But there's nothing for making new directories. I would like to have a "New directory here" button, similar to "New file here".

TyGuy54:
resolved #1040
This is my first time to contributing on open source projects. Hopefully this code is helpfully!

### DIFF
--- a/porcupine/plugins/filemanager.py
+++ b/porcupine/plugins/filemanager.py
@@ -149,36 +149,6 @@ def ask_file_name(
 
     return new_path
 
-
-    def update_dialog_state(*junk: object) -> None:
-        if overwrite_var.get():
-            ok_button.config(state="normal")
-            entry.config(state="disabled")
-            return
-
-        entry.config(state="normal")
-        name = entry.get()
-        try:
-            possible_new_path = (target_dir / "dummy").with_name(name)
-        except ValueError:
-            ok_button.config(state="disabled")
-            return
-
-        ok_button["state"] = "disabled" if possible_new_path.exists() else "normal"
-
-    file_name_var.trace_add("write", update_dialog_state)
-    overwrite_var.trace_add("write", update_dialog_state)
-    entry.bind("<Return>", (lambda event: ok_button.invoke()), add=True)
-    entry.bind("<Escape>", (lambda event: cancel_button.invoke()), add=True)
-    entry.select_range(0, "end")
-    entry.focus()
-
-    dialog.title(dialog_title)
-    dialog.resizable(False, False)
-    dialog.wait_window()
-
-    return new_path
-
 # Not necessarily same concept as Porcupine's project root
 def find_git_root(path: Path) -> Path | None:
     for parent in path.parents:
@@ -393,7 +363,7 @@ def new_directory_here(path: Path) -> None:
         path=path.parent
 
     dir_name = ask_file_name(path, "", mode=FilenameMode.NEW_DIRECTORY)
-    if dir_name :
+    if dir_name:
         dir_name.mkdir()
         get_tab_manager().event_generate("<<FileSystemChanged>>")
 
--- a/porcupine/plugins/filemanager.py
+++ b/porcupine/plugins/filemanager.py
@@ -32,6 +32,11 @@ class FilenameMode(enum.Enum):
     NEW = ("New file", "Enter a name for the new file. Remember the extension (e.g. .py).")
     PASTE = ("File conflict", "There is already a file named {name} in {parent}.\n\n{plus}")
 
+class DirectoryName(enum.Enum):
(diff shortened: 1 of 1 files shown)


porcupine/plugins/filemanager.py
```diff
@@ -149,6 +154,98 @@ def update_dialog_state(*junk: object) -> None:
 
     return new_path
 
+def ask_directory_name(
```
Akuli, line 157: I don't like the copy/paste. Is there a reason why we can't just use `ask_file_name`?

porcupine/plugins/filemanager.py
```diff
@@ -358,12 +455,22 @@ def new_file_here(path: Path) -> None:
         get_tab_manager().event_generate("<<FileSystemChanged>>")
         get_tab_manager().open_file(name)
 
+def new_directory_here(path: Path) -> None:
+    # print('New Directory Created')
+    name = ask_directory_name(path, "", mode=DirectoryName.NEW)
+    if path.is_dir():
+        path=path.parent
+        name.mkdir()
```
Akuli, line 463: If I right-click a directory, choose "New directory here", and then click Cancel, I get:
```
    return self.func(*args)
  File "/home/akuli/porcu/porcupine/plugins/filemanager.py", line 463, in new_directory_here
    name.mkdir()
AttributeError: 'NoneType' object has no attribute 'mkdir'
```

porcupine/plugins/filemanager.py
```diff
@@ -32,6 +32,11 @@ class FilenameMode(enum.Enum):
     NEW = ("New file", "Enter a name for the new file. Remember the extension (e.g. .py).")
     PASTE = ("File conflict", "There is already a file named {name} in {parent}.\n\n{plus}")
 
+class DirectoryName(enum.Enum):
+    RENAME = ("Rename", "Enter a new name for {name}:")
+    NEW = ("New directory", "Enter a name for the new directory.")
+    PASTE = ("Directory conflict", "There is already a Directory named {name} in {parent}.\n\n{plus}")
```
Akuli, line 38: These could probably be just inside `FilenameMode`. Something like this:

```python3
class FilenameMode:
    RENAME = (...)
    NEW_FILE = (...)
    NEW_DIRECTORY = (...)
    PASTE_FILE = (...)
    PASTE_DIRECTORY = (...)
```

Akuli (commented):
Thanks!

TyGuy54:
All the above issues should be fixed. 
I'm not sure what I was thinking with the whole copying and pasting, It kept me up last night haha! But for the error you got when you hit cancel when making a new directory was caused by where the `name.mkdir()` it needed to be in the `if name:` block in the `new_directory_here ` function.

porcupine/plugins/filemanager.py
```diff
@@ -150,6 +150,35 @@ def update_dialog_state(*junk: object) -> None:
...
+    entry.bind("<Escape>", (lambda event: cancel_button.invoke()), add=True)
+    entry.select_range(0, "end")
+    entry.focus()
+
+    dialog.title(dialog_title)
+    dialog.resizable(False, False)
+    dialog.wait_window()
+
+    return new_path
+
```
Akuli, line 181: You apparently forgot to delete some of the copy/paste. There's still two `return new_path`s, and the code after the first one never runs.

porcupine/plugins/filemanager.py
```diff
@@ -352,18 +381,29 @@ def new_file_here(path: Path) -> None:
...
         get_tab_manager().event_generate("<<FileSystemChanged>>")
-        get_tab_manager().open_file(name)
 
 
 commands = [
     # Doing something to an entire project is more difficult than you would think.
     # For example, if the project is renamed, venv locations don't update.
     # TODO: update venv locations when the venv is renamed
     Command("New file here", "<<FileManager:New file>>", (lambda p: True), new_file_here),
+    Command("New directory here", "<<FileManager:New directory>>", (lambda p: True), new_directory_here),
```
Akuli, line 406: If you right-click a file, there should be no "New directory here" option.
```suggestion
    Command("New directory here", "<<FileManager:New directory>>", (lambda p: p.is_dir()), new_directory_here), 
```

Akuli: Nevermind. I forgot about the whole `path=path.parent` thing.

Akuli:
You need to run `black porcupine/plugins/filemanager.py` (see the newly updated README), and then this PR is done :)

Akuli (approved):
Thanks!

PR #1045, opened 2022-04-10
README: explain mypy, black and isort

Akuli:
The README previously mentioned `mypy`, `black` and `isort`, but didn't explain what they are and that you need to run them on your pull requests. This clearly wasn't enough, because most people new to Porcupine don't use these tools.

@ihammadasghar and @TyGuy54: what do you think?


ihammadasghar:
Much more clearer now, will make sure to run them on future PRs. Thanks!

PR #1048, opened 2022-04-10
run plugin: handle lots of output e.g. print in infinite loop

### LINKED ISSUE #969: porcupine freezes if you print in an infinite loop
```python
import os
while True:
    print(os.getpid())
```

Run this with output going to Porcupine window. I recommend printing the PID so you can kill the process manually.

IDLE's solution is replacing `sys.stdout` with an object where `.write()` waits until the text has been actually displayed. Even though that's too python-specific for Porcupine, the same idea might work, if there's a way to make a blocking file descriptor.

Akuli:
Fixes #969

### DIFF
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -52,6 +52,9 @@ def open_file_with_line_number(path: Path, lineno: int) -> None:
         tab.textwidget.tag_add("sel", "insert", "insert lineend")
 
 
+MAX_SCROLLBACK = 5000
+
+
 class Executor:
     def __init__(self, cwd: Path, textwidget: tkinter.Text, link_manager: textutils.LinkManager):
         self.cwd = cwd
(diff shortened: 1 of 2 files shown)


porcupine/plugins/run/no_terminal.py
```diff
@@ -185,8 +191,22 @@ def stop(self, *, quitting: bool = False) -> None:
...
             if not quitting:
-                self._queue.put(("error", "Killed."))
-                self._flush_queue()
+                assert self._thread is not None
+
+                # Consume queue until the thread stops
+                while True:
+                    message_type, text = self._queue.get(block=True)
+                    # For killing messages, a separate "Killed." will be added below
+                    if sys.platform != "win32" and (message_type, text) != (
```
Akuli, line 200:
```suggestion
                    if sys.platform == "win32" or (message_type, text) != (
```

porcupine/plugins/run/no_terminal.py
```diff
@@ -185,8 +191,22 @@ def stop(self, *, quitting: bool = False) -> None:
 
         else:
             if not quitting:
-                self._queue.put(("error", "Killed."))
-                self._flush_queue()
+                assert self._thread is not None
```
Akuli, line 194: If the thread has stopped for some unexpected reason, it's better to error here than to freeze the GUI in `.get(block=True)` below.
```suggestion
                assert self._thread is not None and self._thread.is_alive()
```

PR #1049, opened 2022-04-11
run plugin: detect filenames from valgrind output

### LINKED ISSUE #1019: parenthesized filename:linenumber not detected
`SDL_assert` prints things like this:

```
Assertion failure at map_get_rotation (src/map.c:391), triggered 1 time:
  'fabsf(mat3_det(res) - 1) < 0.01f'
```

Would be nice if `(src/map.c:391)` was clickable.

Akuli:
Fixes #1019

### DIFF
--- a/tests/test_run_plugin.py
+++ b/tests/test_run_plugin.py
@@ -167,6 +167,10 @@ def test_bindcheck_message(filetab, tabmanager, tmp_path, wait_until):
     assert click_last_link() == "asdf.bind('<Foo>', print)"
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="commands below wouldn't work on windows even if valgrind supported windows",
+)
 @pytest.mark.skipif(shutil.which("gcc") is None, reason="C compiler needed")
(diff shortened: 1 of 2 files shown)


tests/test_run_plugin.py
```diff
@@ -167,6 +167,31 @@ def test_bindcheck_message(filetab, tabmanager, tmp_path, wait_until):
     assert click_last_link() == "asdf.bind('<Foo>', print)"
 
 
+# caplog needed to silence logging errors from langserver plugin, which tries to start clangd
+@pytest.mark.skipif(shutil.which("gcc") is None, reason="C compiler needed")
+@pytest.mark.skipif(shutil.which("valgrind") is None, reason="need valgrind")
```
Akuli, line 172:
```suggestion
@pytest.mark.skipif(shutil.which("gcc") is None, reason="C compiler needed")
@pytest.mark.skipif(shutil.which("valgrind") is None, reason="need valgrind")
# caplog needed to silence logging errors from langserver plugin, which tries to start clangd
```

PR #1062, opened 2022-05-07
filetypes: properly reset settings from previous filetype

### LINKED ISSUE #1036: css unwanted autoindent
```css
    background-color: #00ff00;
```
If I press enter at the end of this line, the next line autoindents. I think because it treats `#` as comment prefix and `:` as in Python

Akuli:
Fixes #1036

### DIFF
--- a/porcupine/settings.py
+++ b/porcupine/settings.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import atexit
+import builtins
 import copy
 import dataclasses
 import enum
@@ -90,12 +91,14 @@ class _Option:
         self.default = default
         self.type = type_
         self.converter = converter
+        self.tag: str | None = None
 
 
 @dataclasses.dataclass
 class _UnknownOption:
     value: Any
     call_converter: bool
+    tag: str | None
 
 
 def _default_converter(value: Any) -> Any:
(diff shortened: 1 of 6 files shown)


porcupine/settings.py
```diff
@@ -223,22 +227,24 @@ def set(
 
         You can specify ``call_converter`` to force the converter to be or
         to not be called.
-        """
-        # ...even though this method isn't named 'set_'. But the docstring is
-        # used in settings.rst to document a global "function".
 
+        If ``tag`` is given, the option name will be returned from
+        :meth:`get_options_by_tag`. Otherwise the tag of the option is cleared.
```
Akuli, line 232:
```suggestion
        Each option has a tag that shows up in :meth:`debug_dump` and :meth:`get_options_by_tag`.
        If the ``tag`` argument is given, the tag of the option will be changed to it.
        Otherwise the tag is cleared.
        Tagging is useful for figuring out where/why an option was set:
        if you call ``.set()`` in two places with different tags,
        the option's current tag tells which of the two places has set the current value.
```

porcupine/settings.py
```diff
@@ -260,6 +266,10 @@ def set(
         else:
             self._change_event_widget.event_generate(event_name)
 
+    def get_options_by_tag(self, tag: str) -> builtins.set[str]:
+        """Return the names of all options that were :meth:`set` with the given ``tag``."""
```
Akuli, line 270:
```suggestion
        """Return the names of all options whose current value was set with the given ``tag``."""
```

PR #1065, opened 2022-05-14
Fix resolution on Windows 10

### LINKED ISSUE #24: DPI awareness

VideoCarp:
Fixes #24 

Fixed the resolution on Windows 10 by using `ctypes.windll.shcore.SetProcessDpiAwareness(1)`.

Upon running, 5 tests failed. I'm not sure if this is induced by a problem on my end (and thus has nothing to do with the PR) or not. But it showed a higher resolution on execution on my Windows 10 machine. 
image

### DIFF
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import dataclasses
 import logging
 import os
+import sys
 import tkinter
 import types
 from typing import Any, Type
@@ -12,7 +13,7 @@ import platform
 from porcupine import images, tabs, utils
 
 # Windows resolution
-if platform.system() == "Windows":
+if sys.platform == "win32":
     from ctypes import windll
 
     windows_version = platform.release()
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -4,11 +4,11 @@ from __future__ import annotations
 import dataclasses
 import logging
 import os
+import platform
 import sys
 import tkinter
 import types
 from typing import Any, Type
-import platform
 
 from porcupine import images, tabs, utils
 
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -17,9 +17,10 @@ if sys.platform == "win32":
     from ctypes import windll
 
     windows_version = platform.release()
-    if windows_version == "10" or windows_version == "11":
+    try:
         windll.shcore.SetProcessDpiAwareness(1)
-    elif windows_version == "7" or windows_version == "Vista":
+    except (AttributeError, OSError):
+        # Windows 7 or older
         windll.user32.SetProcessDPIAware()
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


porcupine/_state.py
```diff
@@ -6,11 +6,13 @@
 import os
 import tkinter
 import types
+from ctypes import windll
```
Akuli, line 9: This doesn't work on my computer. I don't use Windows, so importing `windll` fails with `ImportError` for me. You need to use e.g. try/except or `if sys.platform == "win32"`.

Akuli (commented):
Thanks! This fixes the oldest open issue in porcupine.

Akuli:
The test failures are unrelated to this PR. I created #1066, please answer a couple questions there.

porcupine/_state.py
VideoCarp: Added.

porcupine/_state.py
```diff
@@ -10,6 +10,13 @@
 
 from porcupine import images, tabs, utils
 
+try:
+    from ctypes import windll
```
Akuli, line 14: This is a bit annoying. Mypy, the type checker that Porcupine uses, complains about this because it doesn't exist on Windows. But if I silence the mypy error by adding a `# type: ignore` comment here, it then complains that the `# type: ignore` comment is unnecessary, which it is on Windows.

The easiest fix is to switch from try/except to `if sys.platform == "win32"`.

porcupine/_state.py
```diff
@@ -7,9 +7,15 @@
 import tkinter
 import types
 from typing import Any, Type
+from sys import platform
```
Akuli, line 10: There's a mypy bug that prevents this from working :D https://github.com/python/mypy/issues/10054

You need to use `import sys` and `sys.platform`.

porcupine/_state.py
```diff
@@ -7,9 +7,15 @@
...
 
 from porcupine import images, tabs, utils
 
+if platform == "win32":
+    from ctypes import windll
+
+    windll.shcore.SetProcessDpiAwareness(1)
```
Akuli, line 17: On Windows 7, I get this error:

win7

I know that Windows 7 is no longer supported, but I use it whenever I develop Windows things for Porcupine, so I want it to still work.

```suggestion
    try:
        windll.shcore.SetProcessDpiAwareness(1)
    except FileNotFoundError:  # windows 7
        pass
```

Akuli (commented):
Thanks for the patience! We're almost done here.

Akuli:
I fixed the remaining issues myself.

VideoCarp:
I've created another solution, which might improve DPI awareness on Windows 7, too. Works on my Windows 10, it should
also work with Windows 7. Maybe Vista, too. I forgot to add `or windows_version == "8"`. So that might be important.

> I fixed the remaining issues myself.

Akuli:
It might be easiest to use try/except for this. That way we won't forget to check for a specific windows version, and it will simply use `SetProcessDpiAwareness` if it exists,  `SetProcessDPIAware` if not.

Akuli:
Thanks for fixing this!

Issue #885, opened 2021-10-10
git file locking error


Akuli:
https://github.com/Akuli/porcupine/blob/30404adf536783e115e1547f7d61843410740e9a/porcupine/plugins/git_status.py#L33-L42

Maybe the timeout isn't that huge after all. I believe it times out sometimes and that's what causes me to see errors like this:

```
(env) akuli@akuli-desktop:~/porcu$ git checkout master
fatal: Unable to create '/home/akuli/porcu/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```

Akuli:
Today it timed out after 15 seconds, lol.

Akuli:
Still keeps happening

PR #1075, opened 2022-06-01
run plugin: avoid errors for bad file paths

Akuli:
Today at work, I printed some random-ish bytes in Porcupine. That caused the run plugin to think it might be a file name and a line number, but it errored when checking if a file with that name exists, because the "filename" was really long.

### DIFF
--- a/porcupine/plugins/run/no_terminal.py
+++ b/porcupine/plugins/run/no_terminal.py
@@ -222,39 +222,6 @@ class Executor:
             get_tab_manager().event_generate("<<FileSystemChanged>>")
 
 
-# Path.is_file() can fail with "OSError: [Errno 36] File name too long"
-# crasher below:
-#
-#    print(bytes.fromhex("""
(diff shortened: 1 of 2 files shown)


Moosems:
Why did the tests fail?

Akuli:
It's not caused by the changes in this PR. It seems to be #885: Porcupine runs `git status` in the background, timing out after 15sec, and apparently `git status` can take longer than 15sec when we are simultaneously running tests. A good solution would be to get rid of the timeout, at least when running tests and maybe in general: killing a currently running `git status` just doesn't seem to be a good idea.

Let me know if you want to work on porcupine. There's plenty of things waiting to be done :)

Moosems:
> Let me know if you want to work on porcupine. There's plenty of things waiting to be done :)

I would be quite interested in this. I need to find time to set it up on my computer but I'll let you know when that's done :)

PR #1078, opened 2022-07-18
Fix annoying "Do you want to close" behavior

### LINKED ISSUE #605: When responding to many "do you want to close" things, I can't see what's already closed
1. Open 3 new files
2. Type something to each
3. Close porcupine
4. Click no 3 times while nothing happens in gui. Would like to see closed tabs go away

rdbende:
Fixes #605

Today I got annoyed by this behavior, so trying to fix it.
However, placing the `event_generate` before the `can_be_closed` stuff breaks the restart plugin. It doesn't know if the tab was saved or not. And it doesn't necessarily quit after the event, so it's dumb this way. Crap! 😕

### DIFF
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -126,7 +126,7 @@ def quit() -> None:
     destroyed.
     """
     get_main_window().event_generate("<<PorcupineQuit>>")  # TODO: still needed?
-    
+
     for tab in reversed(get_tab_manager().tabs()):
         if not tab.can_be_closed():
             return
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -125,12 +125,11 @@ def quit() -> None:
     with :meth:`~porcupine.tabs.TabManager.close_tab` and widgets are
     destroyed.
     """
-    for tab in get_tab_manager().tabs():
+    get_main_window().event_generate("<<PorcupineQuit>>")  # TODO: still needed?
+    
+    for tab in reversed(get_tab_manager().tabs()):
         if not tab.can_be_closed():
             return
-        # the tabs must not be closed here, otherwise some of them
-        # are closed if not all tabs can be closed
-    get_main_window().event_generate("<<PorcupineQuit>>")  # TODO: still needed?
-    for tab in get_tab_manager().tabs():
         get_tab_manager().close_tab(tab)
+
     get_main_window().destroy()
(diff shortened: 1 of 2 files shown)


porcupine/_state.py
```diff
@@ -125,12 +125,11 @@ def quit() -> None:
     with :meth:`~porcupine.tabs.TabManager.close_tab` and widgets are
     destroyed.
     """
-    for tab in get_tab_manager().tabs():
+    get_main_window().event_generate("<<PorcupineQuit>>")  # TODO: still needed?
+
+    for tab in reversed(get_tab_manager().tabs()):
```
rdbende, line 130: Idk, it just makes more sense to me to start closing the tabs from the end of the list.

Akuli:
"Do you want to close" only makes sense when you click the close button or press Ctrl+W with unsaved changes, or when the restart plugin is disabled and your changes will actually be gone. So just asking it when the editor is quitting is a bad idea. It worked well when I wrote that code, but that was before the restart plugin existed.

Here's how I would want it to be solved. We would add a new checkbox in the settings dialog, "Remember opened tabs when quitting", that would allow for two different ways to use Porcupine:
- Tab states are saved when you quit Porcupine. No questions asked. When you open Porcupine, you continue from where you left off last time. I know at least one user who loves using Porcupine this way.
- When you open Porcupine, no tabs are opened automatically: your programming experience basically gets a new fresh start. I would like to do this. I think I disabled the restart plugin to get something closer to this :)

The new checkbox should probably be added inside `porcupine/settings.py`, not in a plugin, so that the quitting code can access its value.

rdbende:
> I know at least one user who loves using Porcupine this way.

Now you know two. Or at least one and a half :)
I like that recently used files remain open, so I can continue my work where I left off, but I hate when every it asks, whether I want to save the new unsaved file, I click no, and next time it reopens it.

I assume we can close this as of #1079 (?)

Akuli:
This PR is useful even with #1079. When I, with my "don't save tab states" settings, have many tabs opened when I close Porcupine, I want to see the file that a yes/no/cancel dialog comes from.

Akuli:
I have now included the relevant couple lines of this PR into #1079. I didn't include the reversing, because it somehow doesn't feel as natural when not all files contain unsaved changes, but feel free to add it back if you want :)

PR #1079, opened 2022-07-23
Let user decide what to do with open tabs when closing Porcupine

Akuli:
This makes Porcupine work better for the "two different ways to use Porcupine" I explained in #1078. Will conflict with #1078.

### DIFF
--- a/porcupine/settings.py
+++ b/porcupine/settings.py
@@ -414,6 +414,7 @@ def _init_global_gui_settings() -> None:
     global_settings.add_option(
         "default_line_ending", LineEnding(os.linesep), converter=LineEnding.__getitem__
     )
+    global_settings.add_option("remember_tabs_on_restart", default=True)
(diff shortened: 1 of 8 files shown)


porcupine/settings.py
```diff
@@ -414,6 +414,7 @@ def _init_global_gui_settings() -> None:
     global_settings.add_option(
         "default_line_ending", LineEnding(os.linesep), converter=LineEnding.__getitem__
     )
+    global_settings.add_option("remember_tabs_on_restart", default=True)
```
Akuli, line 417: Defining this setting should move into the `restart` plugin

Akuli: @rdbende When you review this PR, feel free to fix this if I haven't fixed it yet :)

rdbende (approved):
I like it!

PR #1084, opened 2022-07-24
Upgrade dependencies

### LINKED ISSUE #409: update mypy
so that don't need to have weird github release link in requirements-dev.txt

### LINKED ISSUE #1080: mypy is old
https://github.com/Akuli/porcupine/blob/001809509927a1d41c2e3cdd284f0d81f1255c86/requirements-dev.txt#L12

Latest mypy version is 0.971. Might as well upgrade all `types-foo` packages while we're at it.

Akuli:
Fixes #1080

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,7 +11,6 @@ requests>=2.24.0, <3.0.0
 # type checking, exact versions to avoid "works on my computer" problems
 mypy==0.971
 types-Pygments==2.12.1
-types-toml==0.10.8
 types-docutils==0.19.0
 types-Send2Trash==1.8.2
 types-setuptools==63.2.1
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -1,22 +1,22 @@
...
-types-Send2Trash==1.8.0
-types-setuptools==57.4.0
-types-ttkthemes==3.2.1
-types-colorama==0.4.3
-types-appdirs==1.4.0
-types-toposort==1.6.0
-types-psutil==5.8.12
+mypy==0.971
+types-Pygments==2.12.1
+types-toml==0.10.8
```
Akuli, line 14: Should no longer be needed. I have switched to tomli.

Akuli:
```suggestion
```

PR #1088, opened 2022-07-24
Add delay to hover popups

rdbende:
@Akuli Is there any particular reason to create a new Label at every popup?
I removed it, because it just didn't seem right, and simplified the delay stuff. But if there are, I'll put it back.


rdbende:
Lol, sh*t, forgot to pull first

porcupine/plugins/hover.py
```diff
@@ -16,24 +16,27 @@ class Response(utils.EventDataclass):
...
+        self._label.place_forget()
+
+    def _show_label(self, message: str) -> None:
+        self.hide_label()
+
+        if message != self._label["text"]:
+            self._label.configure(
+                text=message, wraplength=1000  # place_popup will adjust the wraplength
+            )
+            self._label.after(500, None)
```
Akuli, line 38: This is equivalent to `time.sleep(0.5)`: it freezes the GUI for 0.5 seconds. This is easy to see if you type something while you're hovering, for example, but I think I would occasionally notice it when using the editor normally. You need to make a callback function and pass it to the `.after()` method.

This way to use the `.after()` method is just dumb and weird. Here's its source code:

```python3
    def after(self, ms, func=None, *args):
        """..."""
        if not func:
            # I'd rather use time.sleep(ms*0.001)
            self.tk.call('after', ms)
            return None
        else:
            ...
```

Whoever wrote this a long time ago wanted to mirror what Tcl's `after` command does (there's no separate `sleep` in Tcl), and left a comment about how it felt weird in Python.

porcupine/plugins/hover.py
```diff
@@ -16,24 +16,27 @@ class Response(utils.EventDataclass):
...
         )
+        self._location = textwidget.index("insert")
+
+    def hide_label(self, junk: object = None) -> None:
+        self._label.place_forget()
+
+    def _show_label(self, message: str) -> None:
+        self.hide_label()
+
+        if message != self._label["text"]:
```
Akuli, line 34: Maybe we should get everything under the same `if` statement? If I hover the `i` of `list` in Python code and I move my mouse over the `s`, I currently get a newly placed popup even though the text doesn't change.

Akuli (commented):
There's no particular reason to not reuse the same label. I think both ways are fine.

porcupine/plugins/hover.py
Akuli: Presumably you should use `after_cancel()` too: if another hover appears or `hide_label()` is called during the 500 milliseconds, we don't want to do anything for this hover.

Akuli:
After thinking about this a bit more, I think the fix is in the wrong place.

At a high level, the hover works like this:

1. Mouse or cursor movement
2. `<<HoverRequest>>`
3. langserver
4. `<<HoverResponse>>`
5. Show label

If you are moving your mouse all over Porcupine and never stopping for 500 milliseconds, we don't want to spam the langserver with hover requests. We only want to send a hover request **after** the mouse (or cursor) has been not moving for 500 milliseconds. This way we don't consume as much CPU power, so laptop users will get a longer battery life and people who look at CPU usage graphs won't wonder what Porcupine is doing.

This means that we should insert the 500ms delay between steps 1 and 2, not between steps 4 and 5. It could also go between steps 2 and 3, but then it would be a langserver-specific fix even though this should apply to hovers in general.

porcupine/plugins/hover.py
rdbende: Yea, lol I'm super stupid, and I even remembered that comment. Tcl is weird.

porcupine/plugins/hover.py
rdbende: `_request_hover` hides the label first anyways, and moving the `after` essentially makes this `if` redundant.

Akuli (approved):
I like it! Porcupine now feels much more calm.

PR #1089, opened 2022-07-25
Add pause button

### LINKED ISSUE #971: pause button
On non-Windows it is possible to stop pretty much anything by sending SIGSTOP.

### LINKED ISSUE #1056: pause button
On unix-like systems, it's possible to stop the execution of a process at any point by sending SIGSTOP to it and continue by sending SIGCONT. For example, I have made a game, but it doesn't have a pause feature yet. I can still pause it with `kill -STOP 25781` and `kill -CONT 25781` on terminal.

It would be nice if I could stop and continue playing by simply clicking a button in Porcupine. Other editors I have seen don't have this feature, maybe because people aren't aware of `SIGSTOP` and `SIGCONT`, but I would find it useful.

rdbende:
Fixes #1056, fixes #971

Very basic, needs tests


porcupine/plugins/run/no_terminal.py
```diff
@@ -258,6 +278,19 @@ def stop_executor(self, junk_event: object = None, *, quitting: bool = False) ->
         if self.executor is not None:
             self.executor.stop(quitting=quitting)
 
+    def pause_resume_executor(self, junk: object = None) -> None:
+        if sys.platform == "win32" or self.executor is None:
+            return
+
+        if self.executor.paused:
+            self.executor.send_signal(signal.SIGCONT)
```
Akuli, line 286: I believe it is enough to send `SIGSTOP` and `SIGCONT` to the parent process. For example, let's create a `bash` process with a child process `sleep 10`, and stop only the parent:

```
>>> has_sleeping_child = subprocess.Popen(['bash', '-c', 'sleep 10'])
>>> has_sleeping_child.send_signal(signal.SIGSTOP)
```

If I now wait for more than 10 seconds, I can still see the child process in the process list, meaning it stopped too:

```
>>> subprocess.call('ps aux | grep sleep', shell=True)
akuli     7675  0.0  0.0   5308   580 pts/0    T+   23:29   0:00 sleep 10
akuli     7699  0.0  0.0   6200   720 pts/0    S+   23:30   0:00 grep sleep
0
```

porcupine/plugins/run/no_terminal.py
```diff
@@ -316,7 +349,11 @@ def toggle_visible(junk_event: object = ...) -> None:
 
     runner.hide_button.bind("<Button-1>", toggle_visible, add=True)
     runner.stop_button.bind("<Button-1>", runner.stop_executor, add=True)
+    runner.pause_button.bind("<Button-1>", runner.pause_resume_executor, add=True)
     menubar.get_menu("Run").add_command(label="Show/hide output", command=toggle_visible)
+    menubar.get_menu("Run").add_command(
+        label="Pause/resume process", command=runner.pause_resume_executor
```
Akuli, line 355: On Windows, this should be either grayed out or not exist at all. The button is already invisible because it isn't packed.

Akuli (commented):
Awesome!

The test could be super simple:
- Create a Python script into a temporary directory that prints something, sleeps 0.5 seconds, and print something else.
- Run it and wait for the first print.
- Pause, wait one second, assert that second print does not show up.
- Resume and wait for the second print.

porcupine/plugins/run/no_terminal.py
rdbende: I tested it with Tkinter code, and it only paused the parent process (`sh`), but not Tkinter (`python3`)

Akuli: This works (freezes the GUI):

```
akuli@akuli-desktop:~$ bash -c 'python3 -c "import tkinter; tkinter.Tk().mainloop()"' &
[1] 9741
akuli@akuli-desktop:~$ kill -STOP 9741
```

This doesn't work:

```
akuli@akuli-desktop:~$ sh -c 'python3 -c "import tkinter; tkinter.Tk().mainloop()"' &
[1] 9753
akuli@akuli-desktop:~$ kill -STOP 9753
```

Apparently bash does some magic that makes `SIGSTOP` apply to its child processes too, and `sh` doesn't do it. A few options:
- Figure out what exactly the magic is (with e.g. `strace`, looking around in `/proc`, or googling)
- Change the shell we use from sh to bash
- Keep the complicated child process logic (although I'd rather not mess with the user's child processes if we can avoid that)

Akuli: In either case, I think we can improve that in a separate PR :)

Akuli: I figured out what the "magic" is by looking at the output of `pstree -a`. Here I started `bash -c 'python3 -c "..."' &` from `sh`:

```
`-sh
    |-less
    |-pstree -a
    `-python3 -c import tkinter; tkinter.Tk().mainloop()
        `-{python3}
```

There's no `bash` process! When I ran `bash`, it replaced itself with (using `exec`) Python 3, apparently because that's the only thing it was supposed to run. I get a different result if I tell bash to also `sleep 1`:

```
`-sh
    |-bash -c python3 -c "import tkinter; tkinter.Tk().mainloop()"; sleep 1
    |   `-python3 -c import tkinter; tkinter.Tk().mainloop()
    |       `-{python3}
    |-less
    `-pstree -a
```

And if I now kill with `SIGSTOP`, the GUI doesn't freeze. So the "magic" that bash did was replacing itself with the Python or `sleep` process, so I was really stopping the Python or `sleep` directly.

Ideally there would be a way to disable this behavior so that no matter what commands the user runs, our only-one-level-deep killing logic will not recurse into child processes created by the user's programs.

porcupine/plugins/run/no_terminal.py
```diff
@@ -258,6 +278,19 @@ def stop_executor(self, junk_event: object = None, *, quitting: bool = False) ->
         if self.executor is not None:
             self.executor.stop(quitting=quitting)
 
+    def pause_resume_executor(self, junk: object = None) -> None:
+        if sys.platform == "win32" or self.executor is None:
+            return
```
Akuli, line 283: Mypy's unreachable code logic doesn't understand early-`return` very well: even though `mypy --platform win32` understands that the if statement will always execute, it doesn't realize that the code after the `return` is still reachable on other platforms.

To work around this, we can just avoid early returns:

```
if sys.platform != "win32" and self.executor is not None:
    if self.executor.paused:
        ...
```

tests/test_run_plugin.py
```diff
@@ -429,6 +429,17 @@ def test_smashing_f5(tmp_path, wait_until, use_after_idle):
     assert rest == "Hello\nThe process completed successfully."
 
 
+def test_pause_resume_button(tmp_path, wait_until):
+    (tmp_path / "sleeper.py").write_text("import time; print('before'); time.sleep(0.5); print('after')")
+    no_terminal.run_command(f"{utils.quote(sys.executable)} sleeper.py", tmp_path)
+    wait_until(lambda: "before" in get_output())
+    no_terminal.runner.pause_button.event_generate("<Button-1>")
+    assert "after" not in get_output()
+    time.sleep(1)
```
Akuli, line 438: - Maybe these two should be in reverse? Even if the pause didn't work, it would be very unlikely that `"after"` appears immediately.
- You need to replace `time.sleep(1)` with something that lets the event loop run, so that if the button doesn't work, text will actually get inserted into the output area. For example:
    ```
    sleep_end = time.time() + 1
    wait_until(lambda: time.time() > sleep_end)
    ```

rdbende: 👍

By the way, I assume this is just a mistake in `test_stop_button`:

```python
wait_until(lambda: "started\nKilled.")
```

porcupine/plugins/run/no_terminal.py
```diff
@@ -258,6 +278,17 @@ def stop_executor(self, junk_event: object = None, *, quitting: bool = False) ->
         if self.executor is not None:
             self.executor.stop(quitting=quitting)
 
+    def pause_resume_executor(self, junk: object = None) -> None:
+        if sys.platform != "win32" or self.executor is not None:
```
Akuli, line 282:
```suggestion
        if sys.platform != "win32" and self.executor is not None:
```

tests/test_run_plugin.py
Akuli: Woah, yes that is a mistake :) This is the problem with writing asserts, they aren't supposed to fail and you can't test them failing easily.

rdbende:
Todos:
- [ ] Disable pause/resume, when no process is running
- [ ] Skip test on Windoze
- [ ] Cancel timeout on pause, start again on resume (?)

Akuli:
It's fine to have a timeout running even when paused. It simplifies the code and isn't noticeable otherwise.

porcupine/plugins/run/no_terminal.py
```diff
@@ -122,6 +123,7 @@ def _thread_target(self, command: str, env: dict[str, str]) -> None:
         else:
             self._queue.put(("error", f"The process failed with status {status}."))
         self._queue.put(("end", ""))
+        self.running = False
```
Akuli, line 126: I don't like this attribute: if you forget to set it in one of the many places, we suddenly have a bug that happens rarely and is hard to notice when it does happen. Maybe instead make a method (or property) that checks whether `self._shell_process` is non-None and alive? This way we don't store the same information in two places, in the process object and in a separate boolean.

Akuli (approved):
I like it!

Issue #998, opened 2021-11-30
typo


Akuli:
![anhuge](https://user-images.githubusercontent.com/18505570/144142891-6a9a4305-f836-4a6b-be18-35737b38d8bb.png)

an huge

Moosems:
Location of code?

rdbende:
It's here:

https://github.com/Akuli/porcupine/blob/e94f8e6fc5a5330ddb6c1653acd10067b984329c/porcupine/tabs.py#L173

rdbende:
By the way, if you use Git, you can simply search for such things with `git grep`.

```bash
git grep "an huge"
```

Moosems:
Since I use a Mac it's difficult to set up porcupine (it won't run) so I've been doing my edits and searching in the GitHub editor and viewer

Akuli:
Porcupine should work on mac. I know at least one mac user who has contributed to Porcupine before. If you get an error, can you post it here?

Moosems:
Yeah. It's just import errors that I can't figure out

Moosems:
I'll post em in a few

Moosems:
@Akuli 
```yaml
Traceback (most recent call last):
  File "/Users/Moosems/Downloads/porcupine-master/porcupine/__main__.py", line 6, in <module>
    from porcupine import __version__ as porcupine_version
ModuleNotFoundError: No module named 'porcupine'
```

Akuli:
- What command are you using to run porcupine? You should use `python3 -m porcupine` inside `/Users/Moosems/Downloads/porcupine-master`.
- If you didn't already, you need to create and activate a venv.
- You might want to use git instead of downloading Porcupine as a zip file.

There's commands for doing all this in the README, in the "Developing Porcupine" section.

Moosems:
Oh my goodness this is epic

PR #1091, opened 2022-07-25
Fix typo in #998

Moosems:


rdbende (approved):
@Moosems surely knows English better than I do, so approved :)

PR #1100, opened 2022-07-28
Refactor dialog creating code into util functions

rdbende:
Dialogs are inconsistent

### DIFF
--- a/porcupine/plugins/pluginmanager.py
+++ b/porcupine/plugins/pluginmanager.py
@@ -245,8 +245,7 @@ class PluginDialogContent:
 
 
 def create_dialog() -> PluginDialogContent | None:
-    dialog = utils.make_dialog("plugin_manager")
-
+    dialog = utils.make_dialog_if_not_exists("plugin_manager")
     if dialog is None:
         return None
 
--- a/porcupine/plugins/pluginmanager.py
+++ b/porcupine/plugins/pluginmanager.py
@@ -10,12 +10,10 @@ from functools import partial
 from tkinter import messagebox, ttk
 from typing import List
 
-from porcupine import get_main_window, menubar, pluginloader, settings, textutils, utils
+from porcupine import menubar, pluginloader, settings, textutils, utils
 from porcupine.settings import global_settings
 
 log = logging.getLogger(__name__)
-
-dialog: tkinter.Toplevel | None = None
 
 
 def get_docstring(module_name: str) -> str:
@@ -242,26 +240,26 @@ class PluginDialogContent:
         self._on_select()
         self._update_plz_restart_label()
 
-
-def show_dialog() -> None:
-    global dialog
-    if dialog is not None and dialog.winfo_exists():
-        dialog.lift()
-        dialog.focus()
-    else:
-        dialog = create_dialog()[0]
+    def destroy(self) -> None:
+        self.content_frame.master.destroy()
 
 
-def create_dialog() -> tuple[tkinter.Toplevel, PluginDialogContent]:
-    dialog = tkinter.Toplevel()
+def create_dialog() -> PluginDialogContent | None:
+    dialog = utils.make_dialog("plugin_manager")
+
+    if dialog is None:
+        return None
+
     dialog.title("Porcupine Plugin Manager")
+
     content = PluginDialogContent(dialog)
     content.content_frame.pack(fill="both", expand=True)
-    dialog.transient(get_main_window())
+
     dialog.geometry(f"{DIALOG_WIDTH}x{DIALOG_HEIGHT}")
     dialog.minsize(DIALOG_WIDTH, DIALOG_HEIGHT)
-    return (dialog, content)  # content returned for tests
+
+    return content  # content returned for tests
 
 
 def setup() -> None:
-    menubar.get_menu("Settings").add_command(label="Plugin Manager", command=show_dialog)
+    menubar.get_menu("Settings").add_command(label="Plugin Manager", command=create_dialog)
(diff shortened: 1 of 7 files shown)


porcupine/plugins/run/dialog.py
```diff
@@ -58,7 +58,7 @@ def validate(self, *junk_from_var_trace: object) -> None:
 
 class _CommandAsker:
     def __init__(self, ctx: common.Context):
-        self.window = tkinter.Toplevel()
+        self.window = tkinter.Toplevel(name="command_asker")
```
Akuli, line 61: I think `run_command_asker` would be a better name, just in case another plugin decides to also make a widget named `command_asker`.

porcupine/utils.py
```diff
@@ -645,14 +645,18 @@ def callback(shifted: bool, event: tkinter.Event[tkinter.Misc]) -> Any:
 
 # TODO: document this?
 def ask_encoding(text: str, old_encoding: str) -> str | None:
-    label_width = 400
+    if porcupine.get_main_window().tk.call("winfo", "exists", ".choose_encoding"):
+        porcupine.get_main_window().tk.call("focus", ".choose_encoding")
```
Akuli, line 649: The existing encoding dialog could be completely unrelated, e.g. you could click the encoding choosing buttons of two different tabs without closing the encoding dialog in between.

```suggestion
        porcupine.get_main_window().tk.call("destroy", ".choose_encoding")
```

Akuli (commented):
This code is a bit repetitive. Maybe we could move it into a util function that gets called in many places? It could take in a `name` and return a `Toplevel` (showing or creating it, depending on whether or not it existed).

porcupine/utils.py
rdbende: Wow, you always know some corner case :D
I would have never thought of this one.

rdbende: Or perhaps not so much a corner-case.

rdbende:
@Akuli what should I do with None-ness stuff in `run/dialog.py`?

Akuli:
I split `make_dialog()` into two functions, one that can return `None` and another that can't. Mypy understands this better. Another option would be to use `@overload`, but I think it's overkill for this use case, and useful mostly when you need to type something that was designed before typing was a thing.

porcupine/utils.py
```diff
@@ -678,6 +706,8 @@ def ask_encoding(text: str, old_encoding: str) -> str | None:
     def select_encoding() -> None:
         nonlocal selected_encoding
         selected_encoding = combobox.get()
+
+        assert dialog is not None  # mypy grrr...
```
Akuli, line 710:
```suggestion
```

tests/test_pluginmanager.py
```diff
@@ -6,9 +6,9 @@
 
 @pytest.fixture
 def dialog_content(mocker):
-    dialog, content = create_dialog()
+    content = create_dialog()
     yield content
-    dialog.destroy()
+    content.destroy()
```
Akuli, line 11:
```suggestion
    content.winfo_toplevel().destroy()
```

porcupine/plugins/pluginmanager.py
```diff
@@ -242,25 +240,25 @@ def _set_enabled(self, they_become_enabled: bool) -> None:
         self._on_select()
         self._update_plz_restart_label()
 
+    def destroy(self) -> None:
```
Akuli, line 243:
```suggestion
    def destroy_window(self) -> None:
```

tests/test_pluginmanager.py
Akuli:
```suggestion
    content.destroy_dialog()
```

Akuli:
After working on this a bit more, I'm no longer sure if refactoring the checks into separate functions is a good idea. I will make a new PR without this refactoring so that we can review the refactoring more easily.

porcupine/utils.py
```diff
@@ -645,12 +671,11 @@ def callback(shifted: bool, event: tkinter.Event[tkinter.Misc]) -> Any:
 
 # TODO: document this?
 def ask_encoding(text: str, old_encoding: str) -> str | None:
-    if porcupine.get_main_window().tk.call("winfo", "exists", ".choose_encoding"):
-        porcupine.get_main_window().tk.call("destroy", ".choose_encoding")
+    dialog = make_dialog_if_not_exists("choose_encoding")
+    if dialog is None:
+        # Another (possibly unrelated) encoding chooser is already running
+        return None
```
rdbende, line 677:
```suggestion
    dialog = make_dialog_destroying_if_exists("choose_encoding")
    if dialog is None:
        # Another (possibly unrelated) encoding chooser is already running
        return None
```


Maybe we should remove this check, so if I open "choose encoding", then switch to another tab, and click the encoding button again, it will open a new dialog for that tab, and I don't have to close the existing window, and reopen it.

Akuli:
I don't really like this refactoring, even though it was my idea. It works, but I think everything that goes to utils should be:
1. Difficult, ugly or lot of code to get right
2. Used in multiple places
3. Be significantly easier to use than just doing what the util function does internally
4. Be shorter than just copy/pasting the contents of the util file everywhere

It's fine if not all of these apply, but most of them should. Let's look at the first few utils we have now:
- `running_pythonw` and `python_executable`have properties 1, 2 and 3, but not 4
- `quote()` has all properties
- `copy_type()` has properties 1 and 3, but since it's used only twice and both inside `utils.py` itself, not really properties 2 and 4. Maybe it should be non-public, as in `_copy_type()`?
- `format_command()` has properties 1 and 2, probably not 4, maybe 3
- `file_url_to_path()` has properties 1, 3 and 4, but not 2 (I think it should move to the langserver plugin entirely, as that's the only place where it's used)

The functions in this PR only have property 2. They also kinda have property 1, but not really: it wasn't hard to get right as soon as we realized what the problem was.

tl;dr: I have chosen carefully what to put into utils and these functions just don't fit well.

PR #1105, opened 2022-07-30
windows-build: Try to get a C compiler to work again

Akuli:


Akuli:
It compiles. I will test it once I get access to a Windows VM again.

PR #1106, opened 2022-07-31
Clean up unnecessary utils and document what utils should be like

Akuli:
Came up in #1100

### DIFF
--- a/porcupine/utils.py
+++ b/porcupine/utils.py
@@ -645,13 +645,16 @@ _list_of_encodings = [
 
 # TODO: document this?
 def ask_encoding(text: str, old_encoding: str) -> str | None:
-    label_width = 400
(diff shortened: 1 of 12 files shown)


porcupine/utils.py
```diff
@@ -1,4 +1,18 @@
-"""Handy utility functions."""
+"""A collection of carefully selected utility functions.
+
+Not everything should be an util: for example, if it's only used in one file,
+then it should be defined in that file. Specifically, I believe everything in
+this file should have most of these properties:
+
+* Compared to the rest of Porcupine, the util does something difficult, ugly or tricky.
```
Akuli, line 7:
```suggestion
* The util is a lot of code or it does something that is hard to get right.
```

PR #1114, opened 2022-08-05
New syntax highlighter

### LINKED ISSUE #216: highlighting after reloading is slow, needs some kind of spinner
problem "introduced" in #207: was less noticable before but it's not worth going back to subprocess highlighter just because of this

### LINKED ISSUE #434: javascript highlighter multiline comment bug
1. Open javascript file
2. Write `/*` to line 10
3. Write `*/` to line 20. Nothing happens!
4. Move to end of line 10 and press enter, then it highlights correctly

### LINKED ISSUE #612: typing 'con' really quickly can result in 'cno'
happens when writing c function

`git bisect` says it's #207

### LINKED ISSUE #995: Porcupine becomes very unresponsive when a long list is added
Adding a list of many string values makes Porcupine unresponsive. It accepts the commands but executes them with a long delay and Windows eventually kills it for unresponsiveness.

### LINKED ISSUE #1032: Sometimes responds slowly to keystrokes in C++ code
There are a couple reasons why editing C++ code could be slower than editing e.g. Python code:
- Syntax highlighter (to check this, I will try switching filetype to "Plain text" next time I notice this problem)
- Langserver (to check this, I will edit my `filetypes.toml` to disable the langserver next time I notice this problem

### LINKED ISSUE #1038: Weird syntax highlight inside #if 0 at scrolling
Syntax highlighting is weird inside `#if 0`

https://user-images.githubusercontent.com/77941087/158164580-85837412-f259-4267-82fe-0617e5b3ec06.mp4

I think it's because of the visible-area-tracker-thingy in `highlight.py`

Akuli:
The two syntax highlighters will co-exist. I think they will complement each other nicely:
- Old pygments-based highlighter supports many languages, but has known bugs and is slow in large files.
- New tree-sitter based highlighter is fast and not buggy, but adding support for a new language is quite a bit of manual work.

Todo list:
- [x] Add a sanity check to the binary generating script: it should be possible to load a language (say `python`) from the resulting binary.
- [x] Rewrite the logic for choosing the highlighter to use. It is currently so complicated that I don't remember how it works. It should be one filetype-specific setting in `filetypes.toml`.
- [x] Document `filetypes.toml` things. Presumably the most technical details should be left undocumented, with a mention to just ask me to add support for whatever language you need.
- [ ] Test this on Windows.
- [x] Add a message box that shows up if there is no appropriate binary in the zip. Message box should have a button to make Porcupine issue here on github, and another button to just continue without syntax highlighting. Arm mac users will run into this, unless their Python happens to be running on x86_64 (emulated with an emulator that arm macs come with). Not easy to fix because github actions doesn't feature macos on arm. (Maybe I will find an emulator...)
- [ ] Make sure at least these languages are supported (because I personally need a better highlighter for them lol):
    - [x] Python
        - [ ] Do something about `f"foo {bar}"` highlighting very similarly to `"foo {bar}"`. This is important, because it avoids forgetting the `f` in front of a formatted string.
    - [x] Python stub file `.pyi` (same syntax as `.py`, make sure that it is configured in the same way)
    - [ ] C
    - [ ] C++
    - [x] Rust
    - [ ] JavaScript
    - [ ] Markdown
- [x] Is the `debounce()` stuff needed? The pygments highlighter does need it, because it's slow. Delete it or refactor the function to a place where it isn't copy/pasted.
- [x] Make sure that my refactoring didn't break the pygments highlighter. In particular, what happens to root state marks when a file is reloaded and many big changes happen?

Fixes #1038
Fixes #1032 
Fixes #612 (not really, but makes the problem much less noticeable because there's no longer a noticeable freeze after each key press)
Fixes #434
Fixes #216 (not really, you could still have huge file in a pygments-highlighted language, but it is unlikely because files written in exotic languages tend to be short in my experience)


porcupine/plugins/highlight/tree-sitter-data/token-mappings/python.yml
```diff
@@ -0,0 +1,204 @@
+dont_recurse_inside:
+  - string  # TODO: recurse inside f-strings
+  # TODO: distinguish between variables and attributes?
+  # Can't just add "attribute" to dont_recurse_inside because then the object
+  # whose attribute it is won't highlight, e.g. open("foo", "rb").read()
+
```
rdbende, line 6: Yess, this is what I wanted to suggest. With all this stuff I found the filetype config file quite messy.


My other issue is this:
`pygments`:
image
`tree-sitter`:
image

Syntax highlighting looks quite obscure without punctuation and operator highlighting, so I'd like to fix that.
@Akuli do you think this PR is appropriate for that, or should I open a new one, after this is completed?

Akuli: I'm fine with it either way. I recently moved the config to a yaml file. If you can figure out a way to get this to work, just push your commits to this pr :)

Akuli: lol, this PR has gotten crazy huge: it adds more than 1k lines already. I think I'll get this to pass CI and merged pretty soon, and add all the remaining things in separate PRs.

porcupine/default_filetypes.toml
```diff
@@ -5,7 +5,7 @@
 #
 #    https://github.com/Akuli/porcupine/wiki/Getting-Porcupine-to-work-with-a-programming-language
 #
-# Most commonly used options (plugins may add more):
+# Most commonly used options that Porcupine comes with (plugins may add more):
```
Akuli, line 8:
```suggestion
# Most commonly used options (plugins may add more):
```

porcupine/default_filetypes.toml
```diff
@@ -30,13 +30,48 @@
...
+#        setting chooses from:
+#
+#            "pygments" (default)
+#                This syntax highlighter supports many different languages and
+#                it is easy to configure, but it is slow with large files. It
+#                also has known bugs that are basically impossible to fix:
+#                sometimes editing the code and reverting your edits changes how
+#                the code is highlighted.
+#
+#                Recommended for any language with small files.
```
Akuli, line 44:
```suggestion
#                Recommended for most languages with small files.
```

porcupine/default_filetypes.toml
```diff
@@ -30,13 +30,48 @@
...
+#        them to work with other highlighting libraries. The theme can be
+#        changed in Porcupine Settings (in the Settings menu).
+#
 #    pygments_lexer (default: "pygments.lexers.TextLexer")
-#        Name of a Pygments lexer class that will be used for syntax
-#        highlighting. See the Pygments documentation for lexers that Pygments
-#        comes with:
+#        If syntax_highlighter is set to "pygments", this should be the name of
+#        a Pygments lexer class that will be used for syntax highlighting. See
+#        the Pygments documentation for lexers that Pygments comes with:
```
Akuli, line 65: Old explanation and new explanation are both bad. You should set `pygments_lexer` even if `syntax_highlighting` is set to `"tree_sitter"`, because the pastebin plugin uses the pygments lexer for dpaste.com's syntax highlighting.

porcupine/default_filetypes.toml
```diff
@@ -30,13 +30,48 @@
...
 #
 #            https://pygments.org/docs/lexers/
 #
+#    tree_sitter_language_id
```
Akuli, line 69: Presumably this should be renamed to `tree_sitter_language_name`. In py-tree-sitter's source code, they use a field named `language_id` to hold a pointer to a ctypes function, which is very different from what this is.

porcupine/default_filetypes.toml
```diff
@@ -30,13 +30,48 @@
...
+#        Configuration for the "tree_sitter" syntax highlighter. The following
+#        file in Porcupine's source code explains this setting in more detail:
+#
+#            porcupine/plugins/highlight/tree-sitter-data/token-mappings/readme.txt
```
Akuli, line 73: I don't like how the documentation is scattered in several places. There should be one file that explains the whole story: why I chose pygments initially, what issues did it have and how I solved them, how the tree-sitter highlighter works.

porcupine/default_filetypes.toml
```diff
@@ -313,6 +344,20 @@ tabs2spaces = true
 filename_patterns = ["*.toml"]
 pygments_lexer = "pygments.lexers.TOMLLexer"
 comment_prefix = '#'
+[TOML.tree_sitter]
```
Akuli, line 347: This no longer works. I have switched to putting token mappings into a separate yml file.

porcupine/plugins/highlight/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+    tab.settings.add_option("syntax_highlighter", default="pygments")
+    tab.settings.add_option("tree_sitter_language_id", default="<missing>")
+
+    manager = HighlighterManager(tab, tree_sitter_binary_path)
+    tab.bind("<<TabSettingChanged:pygments_lexer>>", manager.on_config_changed, add=True)
+    tab.bind("<<TabSettingChanged:syntax_highlighter>>", manager.on_config_changed, add=True)
+    tab.bind("<<TabSettingChanged:tree_sitter>>", manager.on_config_changed, add=True)
+    manager.on_config_changed()
+
+    # These need lambdas because the highlighter variable can be reassigned.
```
Akuli, line 116: outdated comment

```suggestion
```

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -0,0 +1,326 @@
...
+        binary_path = Path(dirs.user_cache_dir) / binary_filename
+        try:
+            crc = compute_crc32(binary_path)
+        except FileNotFoundError:
+            log.warning(f"binary has not been extracted yet, extracting now: {binary_path}")
+            zipfile.extract(info, binary_path.parent)
+        else:
+            if crc != info.CRC:
+                log.warning(
+                    f"binary has changed after extracting (CRC mismatch), extracting again: {binary_filename}"
```
Akuli, line 137:
```suggestion
                    f"binary has changed after extracting (CRC mismatch), extracting again: {binary_path}"
```

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -0,0 +1,326 @@
...
+
+    # only returns nodes that overlap the start,end range
+    def _get_all_nodes(
+        self, cursor: TreeCursor, start_point: Point, end_point: Point
+    ) -> Iterator[Node]:
+        assert self._config is not None
+        overlap_start = max(cursor.node.start_point, start_point)
+        overlap_end = min(cursor.node.end_point, end_point)
+        if overlap_start >= overlap_end:
+            return
```
Akuli, line 225:
```suggestion
            # No overlap with the range we care about. Skip subnodes.
            return
```

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -0,0 +1,326 @@
...
+    def update_tags_of_visible_area_from_tree(self) -> None:
+        start, end = self.get_visible_part()
+        start_row, start_col = map(int, start.split("."))
+        end_row, end_col = map(int, end.split("."))
+        start_point = (start_row - 1, start_col)
+        end_point = (end_row - 1, end_col)
+
+        self.delete_tags(start, end)
+
+        for node in list(self._get_all_nodes(self._tree.walk(), start_point, end_point)):
```
Akuli, line 248: why is it `list()`ed?

setup.py
```diff
@@ -43,7 +43,7 @@ def find_metadata() -> dict[Any, Any]:
     install_requires=list(get_requirements()),
     packages=find_packages(),
     package_data={
-        "": ["*.txt", "*.gif", "*.sh", "*.toml", "*.tcl"],
+        "": ["*.txt", "*.gif", "*.sh", "*.toml", "*.tcl", "*.zip"],
```
Akuli, line 46:
```suggestion
        "": ["*.txt", "*.gif", "*.sh", "*.toml", "*.tcl", "*.zip", "*.yml"],
```

porcupine/plugins/highlight/tree_sitter_highlighter.py
Akuli:
```suggestion
        for node in self._get_all_nodes(self._tree.walk(), start_point, end_point):
```

Akuli:
This PR turns out to have a couple obstacles:
- Getting tests to work on windows. The tests try to delete a temporary directory that a `.dll` file was extracted into. It fails because the `.dll` is still in use. If I call `FreeLibrary()` (from winapi with ctypes) manually, the tests work most of the time, but occasionally crash Python.
- Licensing. Do I really need to do like https://github.com/grantjenks/py-tree-sitter-languages#license does and include a list of many projects and their licenses? Or could I instead add a dependency on py-tree-sitter-languages?

Akuli:
I think https://pypi.org/project/tree-sitter-languages/ solves both problems.

Akuli:
...and introduces one new problem: it doesn't support windows at all lol. The tree-sitter highlighter would currently be a pain on windows anyway, because you need a C compiler to install it and installing a C compiler on Windows is really really difficult. Last time I tried, I couldn't get it to work in 3 days.

I added some if statements that make it always use the pygments highlighter on Windows. I might fix this later.

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -0,0 +1,213 @@
...
+from porcupine import textutils
+
+from .base_highlighter import BaseHighlighter
+
+# TODO: how to install tree-sitter and tree-sitter-languages on Windows?
+if sys.platform != "win32" or TYPE_CHECKING:
+    import tree_sitter_languages  # type: ignore[import]
+    from tree_sitter import Node, TreeCursor  # type: ignore[import]
+
+Point = Any
```
Akuli, line 22: isn't this just `tuple[int, int]`?

Akuli:
finally its merged! :D

PR #1115, opened 2022-08-06
Allow passing --verbose-logger multiple times

Akuli:
While working on #1114, I tried:

```
python3 -m porcupine --verbose-logger=porcupine.plugins.pygments_highlight --verbose-logger=porcupine.plugins.tree_sitter_highlight
```

Before this PR, only the second `--verbose-logger` does something. The output is:

```
log file: /home/akuli/.cache/porcupine/log/2022-08-06T15-47-37.txt
porcupine.plugins.tree_sitter_highlight DEBUG: ran tree_sitter_highlight.setup() in 34.375 milliseconds
```

After this PR, both work:

```
log file: /home/akuli/.cache/porcupine/log/2022-08-06T15-48-30.txt
porcupine.plugins.pygments_highlight DEBUG: ran pygments_highlight.setup() in 0.499 milliseconds
porcupine.plugins.tree_sitter_highlight DEBUG: ran tree_sitter_highlight.setup() in 37.312 milliseconds
```

### DIFF
--- a/porcupine/_logs.py
+++ b/porcupine/_logs.py
@@ -62,16 +62,7 @@ def _open_log_file() -> TextIO:
     assert False  # makes mypy happy
 
 
-class _FilterThatDoesntHideWarnings(logging.Filter):
-    def filter(self, record: logging.LogRecord) -> bool:
-        return record.levelno >= logging.WARNING or super().filter(record)
-
-
-# verbose_logger can be:
-#   - empty string (print everything)
-#   - logger name (print only messages from that logger)
-#   - None (only print errors)
-def setup(verbose_logger: str | None) -> None:
+def setup(*, all_loggers_verbose: bool, verbose_loggers: list[str]) -> None:
     handlers: list[logging.Handler] = []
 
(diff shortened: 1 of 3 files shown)


porcupine/_logs.py
```diff
@@ -87,11 +78,18 @@ def setup(verbose_logger: str | None) -> None:
...
+        print_handler.setLevel(logging.DEBUG)
+        if not all_loggers_verbose:
+            print_handler.addFilter(
+                lambda record: (
+                    # Always show warnings and errors
+                    record.levelno >= logging.WARNING
+                    # If --verbose-loggers=foo is passed on command line, show all messages
+                    # from loggers "foo", "foo.bar", "foo.bar.baz"
+                    or record.name in verbose_loggers
+                    or any(record.name.startswith(name + ".") for name in verbose_loggers)
```
Akuli, line 90: This looks like it could be slow. This runs for all log messages, and this would probably be best avoided with a `setLevel` when there are no extra-verbose loggers.

Issue #1120, opened 2022-08-06
github actions is slow


Akuli:
It takes quite a while for the CI to finish when making a pull request. Much of the time is spent in `pip install`. Maybe we should pass `cache: 'pip'` to the setup-python action? https://github.com/actions/setup-python#caching-packages-dependencies

Akuli:
Still pretty slow, but I can't think of easy ways to speed it up further.

PR #1129, opened 2022-08-08
github actions: Enable pip cache

Akuli:
Might help with #1120


Akuli:
Closing and reopening to run CI again, so I can see if it's any faster.

Akuli:
pip install still takes 1.5min on windows. I will try installing `wheel` so it can build and cache wheels

Akuli:
Now `pip install` on Windows is between 12sec and 16sec.

PR #1137, opened 2022-08-12
niceify tree-sitter python highlight

### LINKED ISSUE #1135: Highlighting operators
`pygments`:
image
`tree-sitter`:
image

Syntax highlighting looks quite obscure without punctuation and operator highlighting, so I'd like to fix that.

_Originally posted by @rdbende in https://github.com/Akuli/porcupine/pull/1114#discussion_r939728145_

rdbende:
Fixes #1135


porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -6,49 +6,109 @@ dont_recurse_inside:
 
 token_mapping:
   comment: Token.Comment
+  ERROR: Token.Error
```
Akuli, line 9: When does this show up?

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -6,49 +6,109 @@ dont_recurse_inside:
...
   for: Token.Keyword
   from: Token.Keyword
   global: Token.Keyword
   if: Token.Keyword
   import: Token.Keyword
-  in: Token.Keyword
-  is: Token.Keyword
   lambda: Token.Keyword
-  none: Token.Keyword
+  match: Token.Keyword
```
Akuli, line 45: I like how `match` changes color as you type a `match` statement, and it gets its special meaning.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -202,3 +255,132 @@ token_mapping:
     type: Token.Name.Builtin
     vars: Token.Name.Builtin
     zip: Token.Name.Builtin
+
+    # builtin.Pseudo
```
Akuli, line 259: Would this be a good category for `reveal_type()`?

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -202,3 +255,132 @@ token_mapping:
...
+    __objclass__: Token.Name.Variable.Magic
+    __package__: Token.Name.Variable.Magic
+    __path__: Token.Name.Variable.Magic
+    __qualname__: Token.Name.Variable.Magic
+    __self__: Token.Name.Variable.Magic
+    __slots__: Token.Name.Variable.Magic
+    __spec__: Token.Name.Variable.Magic
+    __weakref__: Token.Name.Variable.Magic
+
+    # Magic funcs
```
Akuli, line 292: I don't like the `__blah__` lines:
- They will easily get outdated as new things are added to Python, even more easily than other things in this file. I remember that in the past, pygments has not highlighted all of my magic variables I have used. I don't remember which magic variable that was though.
- They can confuse users: if you e.g. write `__init__` somewhere in the middle of a file, not in a class, you suddenly see it colored and think you're doing it right.
- When defining a magic method, `def __init__(...)` (which is by design more common than explicitly calling a magic method), you don't get the magic coloring anyway, as it is instead colored as a function name.

None of these are alone a big enough reason to remove the magic names, but it makes me dislike them overall. I'll let you decide though: if you want to keep them, we can keep them :)

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -150,6 +150,13 @@ def update_tags_of_visible_area_from_tree(self) -> None:
             #
             # There's a similar situation in rust: macro name is just an identifier in a macro_invocation
             if (
+                self._language_name == "python"
+                and node.type == "identifier"
+                and node.prev_sibling is not None
+                and node.prev_sibling.type in ("def", "class")
+            ):
+                type_name = f"{node.prev_sibling.type}_name"
```
Akuli, line 158:
```suggestion
                # class Foo:  --> highlight Foo as a class_name
                type_name = f"{node.prev_sibling.type}_name"
```

I don't like this part of the highlighter in general. Apparently tree-sitter has some queries that seem to be like CSS selectors. I'll try them for this eventually.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -6,49 +6,109 @@ dont_recurse_inside:
...
+  '^': Token.Operator
+  '|': Token.Operator
+  '~': Token.Operator
+  '=': Token.Operator
+  ':=': Token.Operator
+  and: Token.Operator.Word
+  in: Token.Operator.Word
+  is: Token.Operator.Word
+  not: Token.Operator.Word
+  or: Token.Operator.Word
```
Akuli, line 95: I don't like `Token.Operator.Word`. Not all themes define a color for it, and that makes the themes quite unusable on pygments: try `assert x is not None` with the `fruity` theme. I also have never understood why some keywords should be highlighted separately from others: I think of them as keywords, not as operators, even if they technically are operators.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -6,49 +6,109 @@ dont_recurse_inside:
...
+  '%': Token.Operator
+  '**': Token.Operator
+  '<<': Token.Operator
+  '>>': Token.Operator
+  '&': Token.Operator
+  '^': Token.Operator
+  '|': Token.Operator
+  '~': Token.Operator
+  '=': Token.Operator
+  ':=': Token.Operator
```
Akuli, line 90: This list is pretty long, and very hard to maintain correctly for many languages. Maybe we should provide a reasonable default in `tree_sitter_highlighter.py`? If the node type consists of punctuation characters only, then automatically map it to a `Token.Operator`?

Akuli (commented):
Thanks!

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
rdbende: With f-strings it shows up when there's no character after the bang, I now could not find any other places where tree-sitter would yield it

image

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
rdbende: > They can confuse users: if you e.g. write `__init__` somewhere in the middle of a file, not in a class, you suddenly see it colored and think you're doing it right.

In fact Pygments does this too

> When defining a magic method, def __init__(...) (which is by design more common than explicitly calling a magic method), you don't get the magic coloring anyway, as it is instead colored as a function name.

Lol, my color scheme doesn't make difference between them, so I didn't even noticed this. I added these before the function definition checking logic, and therefore they worked.

Actually, now I don't like this either. It's almost 100 lines, and not even complete, although I copied them from Pygments

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
Akuli:
```suggestion
  ERROR: Token.Error  # happens at least for f"{asdf!}" with nothing after the '!' character
```

rdbende: After all I'm not sure whether this is actually useful. Pyflakes will mark it as syntax error anyway.

rdbende:
image

Not sure whether I like this or not. On the one hand, it reminds you not to use builtins as variable names, but on the other hand, function arguments shouldn't have different colors, and it looks rather odd in a setup.py file for example.

Akuli:
I think we shouldn't highlight `copyright` or `license`. They conflict with `setup.py` argument names, and having them isn't useful in practice:
- It's unlikely that you want to write them into an editor. They are meant to be used on the `>>>` prompt.
- I don't remember ever using them even on the `>>>` prompt.
- If you accidentally use a variable named `copyright` or `license`, it will work anyway.
- If someone doesn't want you to use a variable named `copyright` or `license`, they probably have configured a linter to complain about it in CI. Not the best user-experience, but much more rare than looking at a `setup.py`.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
rdbende:
```python
def foo: ...
```
```
type=module text='def foo: ...\n'
  type=ERROR text='def foo:'
    type=def text='def'
    type=identifier text='foo'
    type=: text=':'
  type=expression_statement text='...'
    type=ellipsis text='...'
```

Akuli: If we want to display `ERROR` tokens, we should add `ERROR` to `dont_recurse_inside`. But again, pyflakes will highlight the error anyway and also provide a description for it upon hovering, so maybe we shouldn't handle `ERROR`.

rdbende:
Shit this doesn't work for
- non-async functions
- functions without returntype annotations
- classes without superclas

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -117,26 +188,23 @@ token_mapping:
...
-    ValueError: Token.Name.Exception
     Warning: Token.Name.Builtin
-    ZeroDivisionError: Token.Name.Exception
-    __debug__: Token.Name.Builtin
-    __doc__: Token.Name.Builtin
+    
+    # Builtins
+    'True': Token.Name.Builtin
+    'False': Token.Name.Builtin
+    None: Token.Name.Builtin
```
Akuli, line 206: These are already handled by `'true'`, `'false'` and `none` above

```suggestion
```

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
rdbende: What do you mean?

Akuli: mypy has a "fake" built-in `reveal_type()`. It would be good to have it highlighted, even though it's technically not special to Python in any way. Currently there's no indication that you're doing anything right when trying to add `reveal_type()`, as you get linter errors and no sign that it's special.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
rdbende: I'll open a separate PR

Akuli:
Some comments highlight fine. Others do not.

2022-08-14-011857_1920x1080_scrot

rdbende:
tree_sitter think that those comment aren't part of the function body, lol

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -14,27 +14,56 @@ queries:
...
+    (function_definition
+      ("async")? @Token.Keyword
+      ("def") @Token.Keyword
+      name: (identifier) @Token.Name.Function
+      parameters: (parameters) @recurse
+      ("->")? @Token.Punctuation
+      return_type: (type)? @recurse
+      (":") @Token.Punctuation
+      (comment)? @Token.Comment
+      body: (block) @recurse)
```
Akuli, line 27: imagine writing an `if` statement for this :D

Akuli (approved):

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -14,27 +14,56 @@ queries:
...
   assert: Token.Keyword
   async: Token.Keyword
   await: Token.Keyword
   break: Token.Keyword
+  case: Token.Keyword
   class: Token.Keyword
   continue: Token.Keyword
-  def: Token.Keyword
   del: Token.Keyword
+  def: Token.Keyword
```
rdbende, line 63: Whoops lol

rdbende:
A few more todos for the future that  I forgot about:
- Highlight decorators properly
- Highlight attributes (?)
- Highlight constants (?)

PR #1138, opened 2022-08-12
tree-sitter markdown highlight

rdbende:


Akuli:
I would actually prefer using "programming language" things instead of e.g. `Token.General.Emph`. This way markdown will look colorful with themes that were not designed with markup languages in mind.

An example of what I mean from commented-out markdown code that no longer works:

```python3
#    # Markdown
#    if node.type == "text":
#        return "Token.Text"
#    if node.type == "emphasis":  # *italic* text in markdown
#        return "Token.Comment"
#    if node.type == "strong_emphasis":  # **bold** text in markdown
#        return "Token.Keyword"
#    if node.type in ("fenced_code_block", "code_span"):
#        return "Token.Literal.String"
```

Akuli:
I merged #1180 which is my take on doing this. If you don't like it, feel free to make another PR to improve it :)

PR #1140, opened 2022-08-12
replace bindcheck.py with flake8-tkinter

rdbende:


.github/workflows/check.yml
```diff
@@ -123,6 +123,6 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: "3.10"
-    - run: python3 scripts/bindcheck.py porcupine
-    - run: python3 scripts/bindcheck.py more_plugins
-    - run: python3 scripts/bindcheck.py tests
+        cache: pip
+    - run: pip install flake8-tkinter
+    - run: python3 -m flake8 $(git ls-files | grep -E '\.(py|pyw)$') | grep TK231 && exit 1 || exit 0
```
Akuli, line 128: - You are doing the equivalent of `False if foo else True` here. As with Python, `not foo` is simpler: just `! grep TK231` instead of `grep TK231 && exit 1 || exit 0`. That way it also doesn't make the CI green if the `python3 -m flake8` part fails, like it currently does because of how precedence works in bash (we had a CI job that silently failed this way for a long time in typeshed).
- I don't want to suppress TK errors other than 231. For example, 232 was included in bindcheck and is something I care about. It would be better to create a flake8 config file (`.flake8` works, maybe it also supports `pyproject.toml`) that ignores the errors we want to ignore, and for anything not mentioned in the file, it doesn't silence those.

.github/workflows/check.yml
```diff
@@ -123,6 +123,6 @@ jobs:
     - uses: actions/setup-python@v4
```
Akuli, line 123: Can't add a suggestion so far up, but this job is still named `bindcheck`. Maybe it should be renamed to `flake8-tkinter` or `flake8`.

.github/workflows/check.yml
```diff
@@ -123,6 +123,6 @@ jobs:
...
       with:
         python-version: "3.10"
-    - run: python3 scripts/bindcheck.py porcupine
-    - run: python3 scripts/bindcheck.py more_plugins
-    - run: python3 scripts/bindcheck.py tests
+        cache: pip
+    - run: pip install flake8-tkinter
```
Akuli, line 127: I don't want things to break if you add newer, better warnings :) This should be `flake8-tkinter==0.5.0`, either here or in `requirements-dev.txt`.

rdbende:
It's already tomorrow and I seem tired and inattentive :P

Todo:
- If the ci falis, does it show the file and the lineno?
- flake8-tkinter has `tag_bind` stuff in loop warning. Should check for that too

.github/workflows/check.yml
rdbende: > ! grep TK231

Ohh, I didn't even knew that there's a `!` operator. I used ! only for things like `sudo !!` or `!blah` (this 'blah' is infective :D)

So I can do `! ls foo; echo $?` or even `! ls foo && echo $?` to get a big zero, right? (I'm from phone, can' t try it )

Anyways, thanks for the suggestion. I think I should look at your conversation on ##learnpython more often :)

Akuli: `! ls foo; echo $?` works. I wasn't immediately sure about what `! ls foo && echo $?` does, because it depends on operator precedence: does it first run `ls foo && echo $?` which only runs `ls foo`, or does it first run `! ls foo` and because it "succeeds" then `echo $?`. But you're right, it outputs a zero: apparently `!` has higher precedence than `&&`, as it is in most other languages.

The silencing of errors I talked about is also related to operator precedence. The buggy bash code in typeshed looked like:

```
python3 scripts/something.py | tee output.txt || true
```

The purpose of `|| true` was to ignore errors from the `tee` command, but it ended up ignoring errors from **both** commands, so after a while the script was displaying a traceback in the logs but the CI was still green. This is because `||` has very low precedence, even lower than `|`. Running `tee` and `true` in a subshell would have been better:

```
python3 scripts/something.py | (tee output.txt || true)
```

This was not obvious to me when I saw the bug, so I thought I might as well warn you about it while we're on the subject :)

rdbende:
I'm stuck. The `python3` exit code will be ignored whatever I do.

Akuli:
Why are we grepping the output anyway instead of using a flake8 config file? In a config file, you could tell flake8 to ignore any other errors we currently get without the grep.

rdbende:
But will the CI fail, if there's output from flake8?

rdbende:
Apparently yes

.github/workflows/check.yml
```diff
@@ -116,13 +116,13 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v4
       with:
         python-version: "3.10"
-    - run: python3 scripts/bindcheck.py porcupine
-    - run: python3 scripts/bindcheck.py more_plugins
-    - run: python3 scripts/bindcheck.py tests
+        cache: pip
+    - run: pip install flake8-tkinter==0.5.0
```
Akuli, line 127: Let's also pin the flake8 version so that it can't break unexpectedly

```suggestion
    - run: pip install flake8==5.0.4 flake8-tkinter==0.5.0
```

Akuli (approved):
Thanks!

PR #1142, opened 2022-08-13
tree-sitter highlighter: Add support for queries and use them to handle f-strings

### LINKED ISSUE #1136: Highlighting Python f strings
f

Above is pygments highlighter, below is tree-sitter highlighter. It would be good if the tree-sitter highlighter recursed inside f-strings like the pygments highlighter does. This is important because otherwise you don't notice if you forget the `f` in front of the string.

Akuli:
Fixes #1136

### DIFF
--- a/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
+++ b/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
@@ -3,11 +3,6 @@
 # whose attribute it is won't highlight, e.g. open("foo", "rb").read()
 
 queries:
-  # Recurse into only the len(x) part of f"blah {len(x)!r}".
-  # That makes the meaning of !r obvious: it is string syntax, not a part of the expression inside.
-  #
-  # Here "expression" is a supertype node, see query syntax docs.
-  # There doesn't seem to be any way to say "anything except the !r part".
   string: |
     # Highlight all strings with Token.Literal.String
     (string) @Token.Literal.String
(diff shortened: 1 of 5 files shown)


porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -1,16 +1,30 @@
...
+# TODO: distinguish between variables and attributes?
+# Can't just add "attribute" to dont_recurse_inside because then the object
+# whose attribute it is won't highlight, e.g. open("foo", "rb").read()
+
+queries:
+  # Recurse into only the len(x) part of f"blah {len(x)!r}".
+  # That makes the meaning of !r obvious: it is string syntax, not a part of the expression inside.
+  #
+  # Here "expression" is a supertype node, see query syntax docs.
+  # There doesn't seem to be any way to say "anything except the !r part".
```
rdbende, line 10:
```suggestion
```

Duplicated below

Akuli: lolwat

scripts/tree-sitter-dump.py
```diff
@@ -3,28 +3,55 @@
 #
 #    porcupine/plugins/highlight/tree_sitter_data/token_mappings/readme.txt
```
Akuli, line 4:
```suggestion
#    porcupine/plugins/highlight/tree-sitter-token-mappings/readme.txt
```

PR #1144, opened 2022-08-13
Handle common operators in _decide_tag

rdbende:


porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -140,6 +140,9 @@ def _decide_tag(self, node: tree_sitter.Node) -> str:
         else:
             type_name = node.type
 
+        if set(type_name) <= set("+-*/%~&|^!?<>="):
+            return "Token.Operator"
```
Akuli, line 144: This `return` statement is before getting from the `.yml` file, so it's impossible to override this in the `.yml` file. Maybe something like this:

```suggestion
            default = "Token.Operator"
        else:
            default = "Token.Text"
```
And then use `default` instead of `"Token.Text"` below.

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -140,6 +140,9 @@ def _decide_tag(self, node: tree_sitter.Node) -> str:
         else:
             type_name = node.type
 
+        if set(type_name) <= set("+-*/%~&|^!?<>="):
```
Akuli, line 143: - Why not include `@` here?
- Why not include `:` and the rest of punctuation here? Does it look bad in some themes if we map `Operator` and `Punctuation` to the same token tag?

rdbende: > Why not include @ here?

Idk, how other languages use `@`, so I thought it shouldn't be part of the "reasonable defaults"

I use the material theme, and there's no difference between `Operator` and `Punctuation`, I need to check it with other themes though

Akuli: I would prefer to just use the same color, at least for now:
- We don't need to distinguish between `Operator` and `Punctuation` for every supported language. This is important, because we both use color themes that highlight them the same.
- It's unclear whether `.` in Python, Java, C etc. should be `Operator` or `Punctuation`. It is a punctuation character but also an operator for accessing attributes.

rdbende: Actually it looks pretty weird with monokai, but it's my least favorite theme, so I don't really care
image

porcupine/plugins/highlight/tree_sitter_highlighter.py
```diff
@@ -140,9 +140,14 @@ def _decide_tag(self, node: tree_sitter.Node) -> str:
         else:
             type_name = node.type
 
-        config_value = self._config.token_mapping.get(type_name, "Token.Text")
+        if set(type_name) <= set("+-*/%~&|^!?<>=@.,:;()[]{}"):
```
Akuli, line 143: The characters are currently same as string.punctuation except that they are missing ``'\`#_"$``. I think that's good.

Akuli (approved):

Issue #1145, opened 2022-08-14
can't changes filetypes using tree-sitter higlighter


rdbende:
Really weird behaviour, when I want change a Python file's type to Rust for example.

1. open a python file
2. filetypes menu -> rust
3. nothing happens
4. filetypes menu -> python
5. highlights it as rust

PR #1148, opened 2022-08-14
When applying a filetype, defer generating change events until the whole filetype set

Akuli:
Makes it impossible to have other issues similar to #1145. I think that's important, because bugs like this apparently take some time to discover (a few days for #1145) and the resulting behavior can be very weird.

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -238,6 +238,7 @@ def test_change_events(toplevel):
 
     change_events.clear()
     with settings_obj.set_many_at_once():
+        settings_obj.set("foo", "some temporary value")
         settings_obj.set("foo", "xxx")
         settings_obj.set("bar", "yyy")
         assert change_events == []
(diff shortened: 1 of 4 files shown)


tests/test_settings.py
```diff
@@ -211,10 +211,44 @@ def test_font_family_chooser():
...
+    )
+
+    settings_obj.set("foo", "x")
+    assert change_events == ["foo = x"]
+    settings_obj.set("bar", "y")
+    assert change_events == ["foo = x", "bar = y"]
+
+    change_events.clear()
+    with settings_obj.set_many_at_once():
+        settings_obj.set("foo", "xxx")
```
Akuli, line 241: Presumably we should only generate one `foo` change event in this case:
```suggestion
        settings_obj.set("foo", "some temporary value")
        settings_obj.set("foo", "xxx")
```

Akuli:
Even though this is more code than before for doing essentially the same thing, I think this is still an improvement. About half of the plugins use settings, so there shouldn't be weird corner cases in settings that have to be worked around in plugins.

PR #1150, opened 2022-08-14
Highlight decorators

### LINKED ISSUE #1149: highlighting decorators
Above pygments, below tree-sitter:
g

Pygments isn't perfect either here. Presumably the entire decorator should be highlighted in the same color: I am after all decorating with `contextlib.contextmanager`, not with `contextlib`.

If the decorator has an argument list, e.g. `@functools.wraps(blah)`, it should highlight the `(blah)` part as it would highlight it anywhere else, and only the `@functools.wraps` part should be decorator-colored (blue in this theme).

rdbende:
Fixes #1149

### DIFF
--- a/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
+++ b/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
@@ -27,6 +27,16 @@ queries:
       body: (_) @recurse
     )
 
+  class_definition: |
+    (class_definition
+      ("class") @Token.Keyword
+      name: (identifier) @Token.Name.Function
+      superclasses: (argument_list)? @recurse
+      (":") @Token.Punctuation
+      (comment)? @Token.Comment
+      body: (_) @recurse
+    )
+
   decorated_definition: |
     (decorated_definition
       (decorator
@@ -46,16 +56,6 @@ queries:
         )?
       )
       definition: (_) @recurse
-    )
-
-  class_definition: |
-    (class_definition
-      ("class") @Token.Keyword
-      name: (identifier) @Token.Name.Function
-      superclasses: (argument_list)? @recurse
-      (":") @Token.Punctuation
-      (comment)? @Token.Comment
-      body: (_) @recurse
     )
     
 
--- a/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
+++ b/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
@@ -24,7 +24,19 @@ queries:
       return_type: (type)? @recurse
       (":") @Token.Punctuation
       (comment)? @Token.Comment
-      body: (block) @recurse)
+      body: (_) @recurse
+    )
+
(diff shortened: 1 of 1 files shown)


Akuli:
Doesn't (yet) do the right thing for `@functools.wraps(asdf)`.

rdbende:
Yeah, lol now I know why I didn't open a PR for it earlier.

Akuli:
Presumably this still doesn't work for `@decorator(foo)(bar)`. But is that so rare that we'll just happily ignore it?

Akuli:
Apparently `@foo.bar[a](b)[c](d)` is a valid decoration. Picking the `foo.bar` part is difficult, because it's quite deep in the tree: it parses as a call `some_function(d)` being the call, where `some_function` is an item lookup `some_object[c]`, and so on. But Python's syntax guarantees that eventually there will be a name or a dotted name: you can't do `@(1 + 2)` for example, that's invalid syntax.

Akuli:
Actually, `@(1+2)` is not invalid syntax. It is a runtime error. I remember it being invalid syntax though, apparently it has been changed in a recent version.

So we don't have any guarantees that there even is a name eventually. I'm starting to think that as long as `@foo`, `@foo.bar`, `@foo(...)` and `@foo.bar(...)` work, nobody will notice anything :)

rdbende:
> Apparently @foo.bar[a](b)[c](d) is a valid decoration

And it can continues endlessly... I don't think we it worth to recurse that deep.

rdbende:
We could do this, but this will highlight every other function calls as well.

```
call: |
    (call
      function: (_) @Token.Name.Function
      arguments: (_) @recurse
    )

  decorated_definition: |
    (decorated_definition
      (decorator
        ("@") @Token.Name.Decorator
        (identifier)? @Token.Name.Decorator
        (attribute)? @Token.Name.Decorator
        (call
          function: (_) @Token.Name.Decorator
          function:
            (attribute)? @Token.Name.Decorator
            (call)? @recurse
          arguments: (_) @recurse
        )?
      )
      definition: (_) @recurse
    )
```

Akuli:
It now works for `@foo.bar()()`, but not `@foo.bar()()()`. I think that's reasonable.

porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -24,7 +24,29 @@ queries:
       return_type: (type)? @recurse
       (":") @Token.Punctuation
       (comment)? @Token.Comment
-      body: (block) @recurse)
+      body: (_) @recurse
+    )
+
+  decorated_definition: |
```
Akuli, line 30: Is there any reason to put this between `function_definition` and `class_definition`? I would prefer it to be either before both or after both, because `function_definition` and `class_definition` are very similar.

Akuli (approved):
If someone seriously wants to use `@foo[0]` or `@foo()()()` as a decorator, we can later extend this to support it.

Issue #1147, opened 2022-08-14
filetypes menu is unordered


Akuli:
Currently Python is near the top and Rust is near the bottom, because python support was added first and has therefore ended up in the beginning of `default_filetypes.toml`, and Rust support has been added recently. A better order would be case-insensitive alphabetical sorting.

Akuli:
Fixed in #1154 .

PR #1154, opened 2022-08-15
Ordered the Filetypes Menu

sokratisvas:
Hi, here is my attempt to order `default_filetypes.toml`. #1147


rdbende (commented):
Nice, thanks!

However there are a couple of issues with this.

1. This order would be hard to maintain. When we add a new file type, it will likely be added to the end of this file, so the order will be messed up.
2. If I added a new filetype in my personal config file (Menubar -> Settings -> Config files -> Edit filetypes.toml), it won't be ordered, because those configs are appended to the end of the default configs.

A better solution would be to simply sort the list that appears in the menubar. Porcupine already sorts the filetypes in the settings dialog's  combobox, the sorting could be similar here.
https://github.com/Akuli/porcupine/blob/7021ee75bc8db2230936459da9adde3e82ed68c2/porcupine/plugins/filetypes.py#L258-L262

sokratisvas:
Oh I see, sorry for the confusion.

Akuli:
Next steps to fix this pull request:
- `git revert 575480d` (this undoes your previous commit that rearranged `default_filetypes.toml`)
- Add the correct sorting logic to the filetypes plugin

Maybe I should also clarify how `sorted(filetypes.keys(), key=str.casefold)` works.

All file types (combined from `default_filetypes.toml` and the user-specific `filetypes.toml`) end up in the global `filetypes` dict that looks something like this:

```python3
>>> filetypes = {"Python": "asdf", "TOML": "asdf", "reStructuredText": "asdf"}
```

with some other values instead of `"asdf"` (but that doesn't matter, we only need the keys). If you just sort the keys, they end up in the wrong order:

```python3
>>> sorted(filetypes.keys())
['Python', 'TOML', 'reStructuredText']
```

The `key` function specifies how exactly to sort them, and `str.casefold` converts them to lowercase to do the sorting.

```python3
>>> str.casefold("TOML")
'toml'
>>> sorted(filetypes.keys(), key=str.casefold)
['Python', 'reStructuredText', 'TOML']
```

Actually, `str.casefold` is a method that is usually called directly on the string instance:

```python3
>>> str.casefold("TOML")
'toml'
>>> "TOML".casefold()
'toml'
```

We could also use `str.lower` instead of `str.casefold`. The only reason for using `str.casefold` is that a long time ago someone said it's better for comparing strings with non-ASCII characters, but that doesn't matter because filetype names are almost always ASCII anyway.

sokratisvas:
Thanks for the detailed answer. I restored `default_filetypes.toml` and made some changes to `../plugins/filetypes.py`. Now it seems that the Filetypes column on the Menu bar is sorted.

Akuli (approved):
Thanks!

PR #1160, opened 2022-08-19
Fix Alt key bindings on mac

Moosems:
Simple change

### DIFF
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -2,13 +2,13 @@
 
 # On Mac, Alt is used for typing characters, command/ctrl are the main buttons for special commands
 if {[tk windowingsystem] == "aqua"} {
-    set contmand Command
+    set control_ish Command
     set alt_ish Control
     event add "<<RightClick>>" <Button-2>
     event add "<<RightClick>>" <Control-Button-1>
     event add "<<WheelClick>>" <Button-3>
 } else {
-    set contmand Control
+    set control_ish Control
     set alt_ish Alt
     event add "<<RightClick>>" <Button-3>
     event add "<<WheelClick>>" <Button-2>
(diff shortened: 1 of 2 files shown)


Akuli:
It's simpler than I thought :)

A couple thoughts:
- It looks too narrow in an empty file. Maybe it should always have room for three digits, and grow only once you get to 1000, 10000 and so on?
- This breaks the anchors plugin. It's supposed to display an anchor in the line numbers when you press Alt+Shift+A, but it doesn't show up.

porcupine/default_keybindings.tcl
```diff
@@ -3,11 +3,13 @@
 # Use Command on mac, Control on other systems
 if {[tk windowingsystem] == "aqua"} {
     set contmand Command
+    set anchmand Control
     event add "<<RightClick>>" <Button-2>
     event add "<<RightClick>>" <Control-Button-1>
     event add "<<WheelClick>>" <Button-3>
 } else {
     set contmand Control
+    set anchmand $anchmand
```
Akuli, line 12: - I don't like how these are named. Maybe `control_ish` and `alt_ish`?
- `set anchmand $anchmand` is similar to `anchmand = anchmand` in Python: it errors because you never defined what `anchmand` is. Maybe you meant `set anchmand Alt`, or `set alt_ish Alt` or whatever that will be?

Moosems: Yeah, I accidentally used replaced all when changing `Alt` to `$anchmand`. I can change it to alt_ish if you like it

porcupine/default_keybindings.tcl
```diff
@@ -3,11 +3,13 @@
 # Use Command on mac, Control on other systems
 if {[tk windowingsystem] == "aqua"} {
     set contmand Command
+    set alt_ish Control
     event add "<<RightClick>>" <Button-2>
     event add "<<RightClick>>" <Control-Button-1>
     event add "<<WheelClick>>" <Button-3>
 } else {
     set contmand Control
```
Akuli, line 11: Can you also rename `contmand` to `control_ish`?

porcupine/default_keybindings.tcl
```diff
@@ -3,11 +3,13 @@
 # Use Command on mac, Control on other systems
```
Akuli, line 3:
```suggestion
# On Mac, Alt is used for typing characters, command/ctrl are the main buttons for special commands
```

porcupine/default_keybindings.tcl
Moosems: ok!

Akuli (approved):
Thanks!

PR #1161, opened 2022-08-19
Switch to py-tree-sitter-builds

### LINKED ISSUE #1133: New highlighter windows support
Currently Porcupine on Windows always uses pygments, for a couple reasons:
- Installing `tree-sitter` requires a C compiler. Once I tried to install a C compiler on Windows and gave up after 3 days.
- `tree-sitter-languages` doesn't support Windows.

It would be good to get it to work on windows though.

Akuli:
I created https://github.com/Akuli/py-tree-sitter-builds to use py-tree-sitter without having users install a C compiler. The stuff it needs in `requirements.txt` is just a huge mess right now, but I want to know what porcupine's CI says.

Fixes #1133

### DIFF
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -51,8 +51,6 @@
 #                try to configure this highlighter yourself; just ask me to do
 #                it instead by making an issue on GitHub.
 #
-#                Currently this highlighter does not work at all on Windows.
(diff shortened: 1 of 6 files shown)


Akuli:
This is hacky, but it does allow using the new highlighter on windows.

PR #1167, opened 2022-08-21
tree-sitter highlighting for C

Akuli:
Important/useful for a couple reasons:
- C files are often huge, so efficient highlighting matters
- There are known bugs with the pygments highlighter: #1038
- Pygments highlights `#include`, `#define` etc as comments which is just super dumb

### DIFF
--- a/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
+++ b/porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
@@ -30,7 +30,7 @@ queries:
   class_definition: |
     (class_definition
       ("class") @Token.Keyword
-      name: (identifier) @Token.Name.Function  # close enough lol
+      name: (identifier) @Token.Name.Class
       superclasses: (argument_list)? @recurse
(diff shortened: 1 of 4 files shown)


porcupine/plugins/highlight/tree-sitter-token-mappings/python.yml
```diff
@@ -30,7 +30,7 @@ queries:
   class_definition: |
     (class_definition
       ("class") @Token.Keyword
-      name: (identifier) @Token.Name.Function
+      name: (identifier) @Token.Name.Function  # close enough lol
```
rdbende, line 33: Actually, idk why I made this `Token.Name.Function` probably some copy-pasta, it could be `Token.Name.Class`.

Akuli:
Looks pretty good with `Token.Name.Class`

s

Akuli:
`max(8` looks wrong though. I'll add highlighting for numbers

Akuli:
beautiful :D

s

PR #1177, opened 2022-08-25
Auto-reload filetypes when filetypes.toml is saved

### LINKED ISSUE #411: filetypes.toml refresh on save
must restart porcupine :(

Akuli:
Fixes #411

### DIFF
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -100,8 +100,6 @@ def load_filetypes() -> None:
         assert "filetype_name" not in filetype
         filetype["filetype_name"] = name
 
-
-def set_filedialog_kwargs() -> None:
     filedialog_kwargs["filetypes"] = [("All Files", ["*"])] + [
         (
             name,
@@ -123,6 +121,7 @@ def get_filetype_from_matches(
     if not matches:
         return None
     if len(matches) >= 2:
+        # Last match, because it's more likely from the user's configuration rather than default.
         names = ", ".join(matches.keys())
         log.warning(
             f"{len(matches)} file types match {they_match_what}: {names}. The last match will be"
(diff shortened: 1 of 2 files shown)


porcupine/plugins/filetypes.py
```diff
@@ -228,10 +229,27 @@ def on_path_changed(tab: tabs.FileTab, junk: object = None) -> None:
     apply_filetype_to_tab(get_filetype_for_tab(tab), tab)
 
 
+def after_save(tab: tabs.FileTab, junk: object) -> None:
+    if tab.path == Path(dirs.user_config_dir) / "filetypes.toml":
+        # Reload all filetypes when the user saves filetypes.toml.
+        # If user has chosen a custom filetype, leave the tab alone.
```
Akuli, line 235: Why not just find a filetype with whatever `filetype_name` is currently used? It still needs a fallback to `get_filetype_for_tab()` in case the user deletes a filetype.

Akuli:
When ran locally, the new test succeeds on both Windows and Linux (don't have a mac). In CI it succeeds on linux but fails on windows and mac. I will probably just skip the failures, as it doesn't seem to have anything to do with whether it actually works

Akuli:
Still passes locally.

Akuli:
This will interact badly with the editorconfig plugin. It should be able to override anything that filetypes sets.

porcupine/plugins/filetypes.py
Akuli: If the user hasn't selected a custom filetype, we should call `get_filetype_for_tab()`, because it is possible that `filename_patterns` was modified.

Moosems:
I would just keep a list of internal files and reload when any of them are saved. Thats what I plan on doing in DIP but I love how VS Code somehow always updates automatically when its changed even in another editor and needs no prompt.

Moosems:
You could try using [`watchdog`](https://pypi.org/project/watchdog/) for this.

PR #1180, opened 2022-08-27
Markdown highlighting (Akuli's version)

Akuli:
Test file that contains all features I thought of:

    ```
    hello world!! this is a bit lol
    ```

    ###### Blah blah

    ```
    hello world!! this is a bit lol
    ```

    Then you do stuff

        $ echo hello

    - sTep 1
    - SteP 2

    1. foo
    2. bar
    3. baz

    [Heres link](https://example.com/)

    Heres image

    Then u say `print("hi")` and it goes **brr** *bruh*

    Blah [foo] blah. 

    [foo]: https://github.com/

I think this is the last PR needed to make a release with the new highlighter.


porcupine/plugins/highlight/tree-sitter-token-mappings/markdown.yml
```diff
@@ -0,0 +1,48 @@
...
+  code_span: Token.Literal.String           # `print("hi")`
+  fenced_code_block: Token.Literal.String   # ```
+  indented_code_block: Token.Literal.String # indented stuff is similar to ```
+  emphasis: Token.Name.Decorator            # *italic*
+  strong_emphasis: Token.Keyword            # **bold**
+
+  # - Foo
+  # - Bar
+  # - Baz
+  list_marker: Token.Operator
```
Akuli, line 40:
```suggestion
  list_marker: Token.Operator  # "-", "*", or "1." "2." "3." etc in the beginning of a list item
```

PR #1181, opened 2022-08-27
Fix crash on Mac

### LINKED ISSUE #1092: "Save as" crashes Porcupine
As I am using the editor I'm not getting any autocomplete. Is there something I need to do to get the plugin started?

Akuli:
I'm pretty confident this fixes #1092, although I have no way to test it.

### DIFF
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -5,6 +5,7 @@ import argparse
 import fnmatch
 import logging
 import re
+import sys
 from functools import partial
 from pathlib import Path
 from typing import Any, Dict, Optional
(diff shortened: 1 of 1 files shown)


porcupine/plugins/filetypes.py
```diff
@@ -116,6 +117,10 @@ def set_filedialog_kwargs() -> None:
         if name != "Plain Text"  # can just use "All Files" for this
     ]
 
+    # Causes crashes for some Mac users, but not all. See #1092
+    if sys.platform != "darwin":
+        filedialog_kwargs["filetypes"].insert(0, ("All Files", ["*"]))
```
Akuli, line 122:
```suggestion
    if sys.platform != "darwin":
        # Causes crashes for some Mac users, but not all. See #1092
        filedialog_kwargs["filetypes"].insert(0, ("All Files", ["*"]))
```

PR #1186, opened 2022-08-27
Fix "pip install" command, in a hacky way

Akuli:
Maybe I'll try to get `flit` or similar working before I merge this, because this is just bad.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,8 +30,8 @@ def get_requirements() -> Iterator[str]:
     with open("requirements.txt", "r") as file:
         for line in map(str.strip, file):
             if (not line.startswith("#")) and line:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -12,11 +13,30 @@
...
 def get_requirements() -> Iterator[str]:
     with open("requirements.txt", "r") as file:
         for line in map(str.strip, file):
             if (not line.startswith("#")) and line:
-                yield line
+                if ';' in line:
+                    requirement, condition = line.split(';')
+                    if evaluate_condition(condition):
+                        yield requirement
+                        continue
```
Akuli, line 37:
```suggestion
```

PR #1199, opened 2022-10-03
Improve "end" fields of change events, add old_text, delete old_text_len

### LINKED ISSUE #1118: textutils.Change.end is confusing
Bugs with handling change events are difficult to reproduce, so it's important that using them correctly is easy. Porcupine has a dataclass `textutils.Change` that contains:

```python3
    # These should be Tuple[int, int], but they can't be because converting to
    # json and back turns tuples to lists
    start: List[int]
    end: List[int]
    old_text_len: int
    new_text: str
```

Let's say you replace `hello` with `hi` on line 7 with a 4-space indentation before it:
- `old_text_len` is 5
- `new_text` is `"hi"`
- `start` is `[7, 4]`
- What is `end`? Is it the "old end" `[7, 9]` (after spaces and `hello`) or the "new end" `[7, 6]` (after spaces and `hi`)?

Turns out that `end` is the "old end". It kinda makes sense: the "new end" can be computed from `start` and `new_text`, but the "old end" can't when this edit erases text from multiple lines.

Instead of just renaming the field to `old_end`, I think I should have two end fields, `old_end` and `new_end`. I usually dislike redundant fields because it's easy to update one and forget to update the other, but that doesn't really apply here because `Change`s are not meant to be mutated.

Akuli:
Fixes #1118

### DIFF
--- a/porcupine/textutils.py
+++ b/porcupine/textutils.py
@@ -21,10 +21,8 @@ if TYPE_CHECKING:
 @dataclasses.dataclass
 class Change:
     r"""
-    This :mod:`dataclass <dataclasses>` represents any change in a
-    :class:`ChangeTrackingText` widget,
-    where ``old_text_len`` characters between the text widget indexes ``start``
-    and ``end`` get replaced with ``new_text``. For example, this...
+    This :mod:`dataclass <dataclasses>` represents a deletion, insertion or
+    replacing of text in a text widget. For example, this...
     ::
 
         # let's say that text widget contains 'hello world'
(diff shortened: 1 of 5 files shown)


porcupine/textutils.py
```diff
@@ -33,30 +33,19 @@ class Change:
     \...changes the ``'hello'`` to ``'toot'``, and that's represented by a
     ``Change`` like this::
 
-        Change(start=[1, 0], end=[1, 5], old_text_len=5, new_text='toot')
+        Change(start=[1, 0], old_end=[1, 5], new_end=[1, 4], old_text='hello', new_text='toot')
 
     Insertions are represented with ``Change`` objects having ``old_text_len=0``
```
Akuli, line 38: Need to `git grep old_text_len`

porcupine/textutils.py
```diff
@@ -33,30 +31,21 @@ class Change:
...
-    say that the text widget contains ``'toot world'`` and all that is
-    deleted::
-
-        Change(start=[1, 0], end=[1, 10], old_text_len=10, new_text='')
-
-    In this case, ``old_text_len`` happens to be ``10 - 0``, where ``10`` and
-    ``0`` are the column numbers of ``start`` and ``end``. But for changing
-    multiple lines, this doesn't work, because you need to know what exactly
-    was deleted.
+    For replacing, both ``old_text`` and ``new_text`` are non-empty.
```
Akuli, line 49: Already shown above, first example

```suggestion
```

porcupine/textutils.py
```diff
@@ -33,30 +31,21 @@ class Change:
...
     ``textwidget.insert('1.0', 'hello')`` corresponds to this ``Change``::
 
-        Change(start=[1, 0], end=[1, 0], old_text_len=0, new_text='hello')
+        Change(start=[1, 0], old_end=[1, 0], new_end=[1, 5], old_text='', new_text='hello')
 
     For deletions, ``start`` and ``end`` differ and ``new_text`` is empty.
-    If the first line of a text widget contains at least 5 characters, then
-    deleting the first 5 characters looks like this::
+    If the first line of a text widget contains at least 5 characters,
+    say ``'abcde''`, then deleting the first 5 characters looks like this::
```
Akuli, line 44:
```suggestion
    say ``'abcde'`, then deleting the first 5 characters looks like this::
```

porcupine/textutils.py
```diff
@@ -33,30 +31,21 @@ class Change:
     \...changes the ``'hello'`` to ``'toot'``, and that's represented by a
     ``Change`` like this::
 
-        Change(start=[1, 0], end=[1, 5], old_text_len=5, new_text='toot')
+        Change(start=[1, 0], old_end=[1, 5], new_end=[1, 4], old_text='hello', new_text='toot')
 
-    Insertions are represented with ``Change`` objects having ``old_text_len=0``
+    Insertions are represented with ``Change`` objects with empty ``old_text``
     and the same ``start`` and ``end``. For example,
```
Akuli, line 37:
```suggestion
    and the same ``start`` and ``old_end``. For example,
```

PR #1200, opened 2022-10-07
 Filetype for .pyx files

### LINKED ISSUE #1157: Filetype for .pyx files
I came across a file named `core.pyx` when working on https://github.com/grantjenks/py-tree-sitter-languages. It's apparently a Cython file, but the one I worked on only had Python syntax in it.

Porcupine highlighted it properly, but auto-indenting didn't work after typing a `:`, for example. So I had to press the tab key to type things like this into the `.pyx` file:

```
if foo:
    bar
```

This is because `default_filetypes.toml` defines nothing for `.pyx` files, so the `filetypes` plugin falls back to defaults. But syntax highlighting still worked because Pygments still guesses an appropriate lexer:

```
>>> import pygments.lexers
>>> pygments.lexers.get_lexer_for_filename('core.pyx')
<pygments.lexers.CythonLexer>
```

We should add a new filetype for Cython files to default_filetypes.toml, with the relevant parts copied from the `[Python]` section.

lawson89:
Added basics settings for pyx files copies from python

Fixes #1157

### DIFF
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -228,10 +228,9 @@ comment_prefix = '#'
 autoindent_regexes = {indent = '.*:'}
 
 [Cython]
+# mostly copied from [Python]
 filename_patterns = ["*.pyx"]
-pygments_lexer = "pygments.lexers.Python3Lexer"
-syntax_highlighter = "tree_sitter"
-tree_sitter_language_name = "python"
+pygments_lexer = "pygments.lexers.python.CythonLexer"
 tabs2spaces = true
 indent_size = 4
 max_line_length = 79   # pep8 says so, lol
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -227,6 +227,17 @@ autocomplete_chars = ["."]
 comment_prefix = '#'
 autoindent_regexes = {indent = '.*:'}
 
+[Cython]
+filename_patterns = ["*.pyx"]
+pygments_lexer = "pygments.lexers.Python3Lexer"
+syntax_highlighter = "tree_sitter"
+tree_sitter_language_name = "python"
+tabs2spaces = true
+indent_size = 4
+max_line_length = 79   # pep8 says so, lol
+comment_prefix = '#'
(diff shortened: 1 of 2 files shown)


porcupine/default_filetypes.toml
```diff
@@ -227,6 +227,17 @@ autocomplete_chars = ["."]
 comment_prefix = '#'
 autoindent_regexes = {indent = '.*:'}
 
+[Cython]
+filename_patterns = ["*.pyx"]
+pygments_lexer = "pygments.lexers.Python3Lexer"
+syntax_highlighter = "tree_sitter"
```
Akuli, line 233: We should use the pygments highlighter because the tree_sitter highlighter doesn't support `.pyx` files. See comments in the beginning of this file for an explanation, especially the syntax_highlighter part. You also need to change `pygments_lexer`, because `.pyx` files are not just regular Python files.

porcupine/default_filetypes.toml
```diff
@@ -227,6 +227,17 @@ autocomplete_chars = ["."]
 comment_prefix = '#'
 autoindent_regexes = {indent = '.*:'}
 
+[Cython]
```
Akuli, line 230: Maybe add a comment about the copy/pasta, so it's less likely to get outdated?

```suggestion
[Cython]
# mostly copied from [Python]
```

porcupine/plugins/directory_tree.py
```diff
@@ -516,6 +516,7 @@ def setup() -> None:
     tree.bind("<<MenuKey>>", partial(tree._on_right_click, menu_key=True), add=True)
 
     tree.bind("<Key>", tree._cycle_through_items, add=True)
+    tree.bind("<Return>", tree.open_file_or_dir, add=True)
```
Akuli, line 519: Please make a separate pull request of this change. Let me know if you are still new to git and need help with that :)

Akuli (approved):
Thanks!

lawson89:
@Akuli - sorry I am bad at github! Let me know if you want me to open a new PR for this

Akuli:
It's ok! I was able to merge it anyway :)

PR #1202, opened 2022-10-08
Open file when pressing enter in directory tree

### LINKED ISSUE #1076: pressing enter should open file
Directory tree is the thing at left. For example, if it looks like this:

s

you can press the right arrow key to see what's in the `porcupine` folder. Or if you select a file instead, right arrow key opens it in Porcupine. Pressing Enter works too.

If you select a file instead of a folder, the arrow key opens the file in Porcupine, but pressing enter does nothing. It should do the same thing as the right arrow key.

<details><summary>Reasons why supporting the enter key would be good</summary>

- ~Today~ A while ago I disabled my arrow keys to strain my right hand less. I now have to double-click with the mouse.
- Enter is likely more discoverable for people new to porcupine.
- Enter already opens folders.

</details>

Relevant files:
- `porcupine/plugins/directory_tree.py` creates the directory tree
- (might not need this file: ) `porcupine/default_keybindings.tcl` is a Tcl file that runs when Porcupine starts, and it defines names for various key presses.

lawson89:
Add call to open_file_or_dir on Enter key

Fixes #1076

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -518,6 +518,7 @@ def setup() -> None:
     tree.bind("<Key>", tree._cycle_through_items, add=True)
     tree.bind("<Return>", tree.open_file_or_dir, add=True)
 
+
 # Used in other plugins
 def get_directory_tree() -> DirectoryTree:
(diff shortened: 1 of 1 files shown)


Akuli (approved):
Thanks!

Issue #1192, opened 2022-08-29
Add rust example commands to default_filetypes.toml


Akuli:
`default_filetypes.toml` contains, among other things, example commands for various programming languages. See the `[[Python.example_commands]]` sections for how they should look.

Currently rust has no example commands. They should be easy to add, because most rust users simply have `cargo` in their PATH and Porcupine doesn't need to do anything platform-specific to find it.

We should add at least these commands:

`cargo run`
`cargo test`
`cargo build`
`cargo clippy`
`cargo fmt`

I personally use the short versions of these commands commands (e.g. `cargo r` instead of `cargo run`), but using them as defaults in Porcupine would likely be more confusing than helpful.

aloner-pro:
I can do this. Can you assign me these please.

Akuli:
Fixed in #1205.

PR #1205, opened 2022-10-13
Added rust example_commands

aloner-pro:
Worked on #1192 
Hope it is fine. 
If anything is missed would correct it.

### DIFF
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -423,9 +423,19 @@ comment_prefix = '//'
 autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = 'this regex matches nothing^'}
 [[Rust.example_commands]]
 command = "cargo run"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo test"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo build"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo clippy"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo fmt"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo check"
 working_directory = "{project_path}"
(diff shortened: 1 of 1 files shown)


porcupine/default_filetypes.toml
```diff
@@ -421,3 +421,11 @@ tabs2spaces = true
 indent_size = 4
 comment_prefix = '//'
 autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = 'this regex matches nothing^'}
+[[Rust.example_commands]]
+command = "cargo run"
+command = "cargo test"
```
rdbende, line 426: Thanks! This looks promising!
However, you should put every command in its own `[[Rust.example_commands]]` table (you can see above, how is it done for Python commands for example), because this way there are duplicate `command` keys in a `[[Rust.example_commands]]`, and that isn't allowed.

aloner-pro: Oh thank you I will do the same.

Akuli: To test whether it works, you can:
1. Run porcupine
2. Make a file named `asdasd.rs`
3. Press Shift+F5

The command entry that appears should already have `cargo run` in it (taken from this file). As you type into it, other commands should be available with auto-complete, e.g. typing `cargo t` should auto-complete to `cargo test`.

aloner-pro: Regarding running porcupine from where do I run it. Is it from `laucher/lauch.pyw` ? Because then it is giving errors like 
`ModuleNotFoundError: No module named 'porcupine'`

aloner-pro: Regarding running `porcupine`, how do I do that, is it from `launcher/launch.pyw` because then it is showing `ModuleNotFoundError` like 
```
ModuleNotFoundError: No module named 'porcupine'
```
 or is it from `launcher/main.c` but then to it is showing errors both on windows and linux.

Akuli: See instructions in `CONTRIBUTING.md`. The correct command is `py -m porcupine`.

aloner-pro: So I just ran the file and tested it myself. And it is working 🥳 along with being present in the present working directory.

Akuli (approved):
Thanks!

PR #1206, opened 2022-10-14
Filetypes menu: hard to see what is selected

### LINKED ISSUE #441: filetypes menu: hard to see what is selected
If I am in a Python file, there's no checkmark or any other kind of indicator next to "Python":

s

This means that there's no good way to tell what filetype you currently have, and if your file is empty, no good way to notice whether changing the filetype in the menu actually did something.

Tkinter supports radiobutton menu items, which could be used to add a checkmark next to the selected filetype name.

lawson89:
Make FileTypes menu radio buttons and sync with notebook changes

Fixes #441

### DIFF
--- a/porcupine/menubar.py
+++ b/porcupine/menubar.py
@@ -358,6 +358,11 @@ def add_filetab_command(
     set_enabled_based_on_tab(path, (lambda tab: isinstance(tab, tabs.FileTab)))
 
 
+def add_filetype_menuitem(label: str, func: Callable[[tabs.FileTab], object], tk_var: tkinter.StringVar) -> None:
+    get_menu('Filetypes').add_radiobutton(label=f"{label}",
+                                          command=lambda: func(_get_filetab()),
+                                          variable=tk_var)
+
 # TODO: pluginify?
 def _fill_menus_with_default_stuff() -> None:
     # Make sure to get the order of menus right:
--- a/porcupine/menubar.py
+++ b/porcupine/menubar.py
@@ -295,7 +295,7 @@ def set_enabled_based_on_tab(
     return update_enabledness
 
 
-def _get_filetab() -> tabs.FileTab:
+def get_filetab() -> tabs.FileTab:
     tab = get_tab_manager().select()
     assert isinstance(tab, tabs.FileTab)
     return tab
@@ -349,19 +349,14 @@ def add_filetab_command(
     but if you do, they are passed to :meth:`tkinter.Menu.add_command`.
     """
     if func is None:
-        command = lambda: _get_filetab().event_generate(f"<<FiletabCommand:{path}>>")
+        command = lambda: get_filetab().event_generate(f"<<FiletabCommand:{path}>>")
     else:
-        command = lambda: func(_get_filetab())  # type: ignore
+        command = lambda: func(get_filetab())  # type: ignore
 
     menu_path, item_text = _split_parent(path)
     get_menu(menu_path).add_command(label=item_text, command=command, **kwargs)
     set_enabled_based_on_tab(path, (lambda tab: isinstance(tab, tabs.FileTab)))
 
-
-def add_filetype_menuitem(label: str, func: Callable[[tabs.FileTab], object], tk_var: tkinter.StringVar) -> None:
-    get_menu('Filetypes').add_radiobutton(label=f"{label}",
-                                          command=lambda: func(_get_filetab()),
-                                          variable=tk_var)
 
 # TODO: pluginify?
 def _fill_menus_with_default_stuff() -> None:
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -6,6 +6,7 @@ import fnmatch
 import logging
 import re
 import sys
+import tkinter
 from functools import partial
 from pathlib import Path
 from typing import Any, Dict, Optional
@@ -141,9 +142,9 @@ def guess_filetype_from_path(filepath: Path) -> FileType | None:
             name: filetype
             for name, filetype in filetypes.items()
             if any(
-                fnmatch.fnmatch(filepath.as_posix(), "*/" + pat)
-                for pat in filetype["filename_patterns"]
-            )
+            fnmatch.fnmatch(filepath.as_posix(), "*/" + pat)
+            for pat in filetype["filename_patterns"]
+        )
         },
         str(filepath),
     )
@@ -252,6 +253,20 @@ def setup_argument_parser(parser: argparse.ArgumentParser) -> None:
     )
 
 
+filetypes_var: tkinter.StringVar | None = None
+
+
+def _sync_filetypes_menu(event: object) -> None:
+    tab = get_tab_manager().select()
+    if isinstance(tab, tabs.FileTab) and tab.path is not None:
+        try:
+            filetype_name = tab.settings.get('filetype_name', str)
+            if filetypes_var and filetype_name:
+                filetypes_var.set(filetype_name)
+        except KeyError:
+            pass
+
+
 def setup() -> None:
     global_settings.add_option("default_filetype", "Python")
 
@@ -264,13 +279,13 @@ def setup() -> None:
         values=sorted(filetypes.keys(), key=str.casefold),
     )
     set_filedialog_kwargs()
-
+    global filetypes_var
+    filetypes_var = tkinter.StringVar()
     for name in sorted(filetypes.keys(), key=str.casefold):
         escaped_name = name.replace("/", "//")  # doesn't work in all corner cases
-        menubar.add_filetab_command(
-            f"Filetypes/{escaped_name}", partial(apply_filetype_to_tab, filetypes[name])
-        )
+        menubar.add_filetype_menuitem(escaped_name, partial(apply_filetype_to_tab, filetypes[name]), filetypes_var)
 
+    get_tab_manager().bind("<<NotebookTabChanged>>", _sync_filetypes_menu, add=True)
     path = Path(dirs.user_config_dir) / "filetypes.toml"
     menubar.get_menu("Filetypes").add_separator()
     menubar.add_config_file_button(path, menu="Filetypes")
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -9,7 +9,7 @@ import sys
 import tkinter
 from functools import partial
 from pathlib import Path
-from typing import Any, Dict, Optional, Callable
+from typing import Any, Dict, Optional
 
 import tomli
 from pygments import lexers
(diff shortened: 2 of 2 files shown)


porcupine/plugins/filetypes.py
```diff
@@ -252,6 +253,20 @@ def parse_filetype_name(name: str) -> FileType:
...
+
+def _sync_filetypes_menu(event: object) -> None:
+    tab = get_tab_manager().select()
+    if isinstance(tab, tabs.FileTab) and tab.path is not None:
+        try:
+            filetype_name = tab.settings.get('filetype_name', str)
+            if filetypes_var and filetype_name:
+                filetypes_var.set(filetype_name)
+        except KeyError:
+            pass
```
Akuli, line 267: Instead of doing nothing, we should set some dummy value (e.g. empty string) to the filetypes var, so that the menu will display no item as being selected.

porcupine/plugins/filetypes.py
```diff
@@ -252,6 +253,20 @@ def parse_filetype_name(name: str) -> FileType:
     )
 
 
+filetypes_var: tkinter.StringVar | None = None
+
+
+def _sync_filetypes_menu(event: object) -> None:
+    tab = get_tab_manager().select()
+    if isinstance(tab, tabs.FileTab) and tab.path is not None:
```
Akuli, line 261: We shouldn't check `tab.path`, because not-yet-saved files can also have a filetype. We should also ensure that no menu item shows up as selected when the tab isn't a `FileTab`. For example, Porcupine debug prompt tabs are not `FileTab`s, and no filetype should show up as selected when you are in the debug prompt.

porcupine/menubar.py
```diff
@@ -358,6 +358,11 @@ def setup() -> None:
     set_enabled_based_on_tab(path, (lambda tab: isinstance(tab, tabs.FileTab)))
 
 
+def add_filetype_menuitem(label: str, func: Callable[[tabs.FileTab], object], tk_var: tkinter.StringVar) -> None:
+    get_menu('Filetypes').add_radiobutton(label=f"{label}",
```
Akuli, line 362: - You can simply do this in `plugins/filetypes.py`. It is the only file that knows anything about filetypes, and we don't need to change that.
- Nit: `f"{label}"` is unnecessary and same as just `label`.

porcupine/plugins/filetypes.py
```diff
@@ -264,13 +279,13 @@ def setup() -> None:
         values=sorted(filetypes.keys(), key=str.casefold),
     )
     set_filedialog_kwargs()
-
+    global filetypes_var
+    filetypes_var = tkinter.StringVar()
     for name in sorted(filetypes.keys(), key=str.casefold):
         escaped_name = name.replace("/", "//")  # doesn't work in all corner cases
```
Akuli, line 285: We no longer need `escaped_name`, you can just use `name`. Previously, a filetype named `"Foo/bar file"` would result in `add_filetype_command("Filetypes/Foo//bar file", ...)`, and it would eventually call `.add_command(label="Foo/bar file", ...)`. But if a filetype now contains a slash, we simply add a menu item whose label contains a slash.

porcupine/plugins/filetypes.py
```diff
@@ -252,6 +253,30 @@ def parse_filetype_name(name: str) -> FileType:
...
+
+
+def _sync_filetypes_menu(event: object) -> None:
+    tab = get_tab_manager().select()
+    if isinstance(tab, tabs.FileTab):
+        try:
+            filetype_name = tab.settings.get("filetype_name", str)
+            if filetypes_var and filetype_name:
+                filetypes_var.set(filetype_name)
+        except KeyError:
```
Akuli, line 266: This has a weird bug. The default filetype for new files is Python, but if I make a new file, the filetype doesn't show up as Python until I switch to some other tab and back.

I think what's happening is:
1. New tab is created
2. Tab changes, `filetypes_var` is set to empty string because `filetype_name` isn't set yet
3. Filetype is set, but we no longer sync `filetypes_var` until changing tab.

I haven't yet checked whether this is correct. If you need help with debugging, I will help :)

One possible solution is to run the sync whenever the `filetype_name` setting changes. We can do it by binding `<<TabSettingChanged:filetype_name>>` inside `on_new_filetab()`. The documentation of `<<TabSettingChanged:...>>` events (generated from docstrings) is here: https://akuli.github.io/porcupine/tabs.html#porcupine.tabs.FileTab.TabSettingChanged-foo

porcupine/plugins/filetypes.py
```diff
@@ -252,6 +253,30 @@ def parse_filetype_name(name: str) -> FileType:
...
+            filetypes_var.set("")
+    else:
+        filetypes_var.set("")
+
+
+def _add_filetype_menuitem(name: str, tk_var: tkinter.StringVar) -> None:
+    menubar.get_menu("Filetypes").add_radiobutton(
+        label=name,
+        command=lambda: apply_filetype_to_tab(filetypes[name], menubar.get_filetab()),
```
Akuli, line 275: `menubar.get_filetab()` fails if the selected tab is not a `FileTab`. Previously this wasn't a problem, because `add_filetab_command()` includes a call to `set_enabled_based_on_tab()`.

The simplest solution would be disabling the entire "Filetypes" menu, unless you are currently in a `FileTab`, but that's not ideal because then you can't click the "Edit filetypes.toml" button at the bottom. A slightly better solution would be to call `set_enabled_based_on_tab()` in this function, so that every filetype menuitem gets disabled just like before.

Akuli (commented):
It's becoming pretty good :) What remains is a couple weird bugs / corner cases.

porcupine/plugins/filetypes.py
Akuli: This seems to affect more than just new files. Try clicking the "Edit filetypes.toml" button, for example: it opens the file, but the filetypes menu doesn't show TOML as selected until you change tabs.

porcupine/plugins/filetypes.py
lawson89: I noticed this as well - when I click on Filetypes -> Edit filetypes.toml:
I do get a callback to _sync_filetypes_menu and can lookup the Tab but the settings does not contain a filetype_name (or much else)

After I select another tab and then select the tab again it does contain the filetype_name. 
So in between my first callback and my second callback for filetypes.toml the settings data gets populated.
I'm just not sure where that is happening. 
Ideally, I'd get my first callback after whatever is populating the tab settings has finished it's work. Is that possible?

porcupine/plugins/filetypes.py
```diff
@@ -252,6 +253,33 @@ def parse_filetype_name(name: str) -> FileType:
...
+
+
+def _sync_filetypes_menu(event: object) -> None:
+    tab = get_tab_manager().select()
+
+    if isinstance(tab, tabs.FileTab):
+        try:
+            filetype_name = tab.settings.get("filetype_name", str)
+        except KeyError:
+            filetype_name = get_filetype_for_tab(tab).get("filetype_name")
```
Akuli, line 266: We want to keep the variable set to whatever is in the `filetype_name` setting of the current tab, or empty. To me, an obvious solution would be to sync when:

- Different tab is selected
- `filetype_name` setting of any `FileTab` changes (currently selected tab would be enough, but it's easier to bind on all `FileTab`s)

Now it syncs only when the tab changes, and if `filetype_name` hasn't been set yet, it predicts what it will be set to in the future. It probably works, but it feels unnecessarily fragile and weird to me.

lawson89: Ok, yeah I think I'm getting lost in what events to use. 
Tab selection is this event: "<<NotebookTabChanged>>?

And the filetype name change is the one you mentioned earlier: 
One possible solution is to run the sync whenever the filetype_name setting changes. We can do it by binding <<TabSettingChanged:filetype_name>> inside on_new_filetab().

Akuli: Yeah, the events can be confusing :)

- `<<NotebookTabChanged>>`: An event available on any `ttk.Notebook`. Runs when the current tab changes. Settings are not populated yet, because they are added in a callback that runs after a new tab is added.
- `<<TabSettingChanged:filetype_name>>`: Available on any `FileTab` (not the `TabManager`, which is a `ttk.Notebook`). Runs when the setting changes. This includes user selecting it in the menu (although that automatically updates the `StringVar` anyway), when the settings are set for the first time for a new or newly opened file, and in any other situation where the setting might change (e.g. in a future version of Porcupine).
- `get_tab_manager().add_filetab_callback()`: Not an event, but runs on newly added tabs. It currently sets their filetypes, but it could also bind so that the sync runs when the filetype changes.

lawson89: Ok so for my sync listener, should I bind to both <<NotebookTabChanged>> and <<TabSettingChanged:filetype_name>> and then take out the attempt to call get_filetype_for_tab?

Akuli: That would make the most sense to me :) Of course, you can do it in any way that works if you manage to convince me that your way makes sense or is better in some way.

lawson89: Ok, I could not get the <<TabSettingChanged:filetype_name>> to ever fire.
but I added sync call to on_new_filetab in the filetypes plugin and that seems to work
This is probably as close as I can get it so feel free to modify as needed

porcupine/plugins/filetypes.py
```diff
@@ -252,6 +254,32 @@ def parse_filetype_name(name: str) -> FileType:
...
+
+    if isinstance(tab, tabs.FileTab):
+        try:
+            filetype_name: str = tab.settings.get("filetype_name", str)
+        except KeyError:
+            filetype_name = ""
+        if filetypes_var:
+            filetypes_var.set(filetype_name)
+    else:
+        if filetypes_var:
```
Akuli, line 271: As far as I can tell, these if statements are unnecessary. The variable can be a `StringVar` or not exist at all. If it doesn't exist, this would `NameError`, and that apparently never happens. If it is a `StringVar`, the condition evaluates to true.

Akuli (commented):
Thanks for working on this! The code works now and looks clean enough. Just one nit/question.

porcupine/plugins/filetypes.py
lawson89: Yeah I had that check since I couldn't initialize it until setup() of the filetype plugin.
If I initialized it when I declared it (module level variable) then tk wasn't ready and it threw an error.
but in any case I don't think the checks really would have helped since setup should always run before the sync method is called
So I cleaned that method up - let me know what you think

Akuli (approved):
Thanks! Further bugs can be fixed in separate issues and pull requests.

lawson89:
Thanks for walking me through this one!

PR #1207, opened 2022-10-20
Tooltips for image buttons

### LINKED ISSUE #1139: Tooltips for image buttons
The run plugin creates these buttons. They appear when pressing F4 or running a command.

s

It would be nice if the pause button displayed e.g. "Pause execution" when hovered, and similarly for other buttons. A tooltip would be especially useful for the "X" button, because it isn't clear whether it deletes the output when clicked, but it would be clear if it had a tooltip saying "Hide output area" or similar.

The tooltips should match with what is in the Run menu, so:
- "Pause execution" or "Resume execution", depending on whether a process is running
- "Kill process"
- "Hide output"

To test these, just run a `.py` file that contains something like:
```python
import time
time.sleep(60)
```

Porcupine used to have code for creating tooltips, but it was a lot of code to maintain for only one tooltip in the whole Porcupine, so it was deleted in #610. We can revert that PR to get it back (and maybe edit the docstring to be slightly less opinionated towards idlelib).

lawson89:
Fixes #1139

### DIFF
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -40,6 +40,7 @@ Other Tkinter Utilities
 
 .. seealso:: :mod:`porcupine.textutils` contains ``tkinter.Text`` specific things.
 
+.. autofunction:: set_tooltip
 .. autofunction:: bind_tab_key
 .. autofunction:: add_scroll_command
 .. autofunction:: run_in_thread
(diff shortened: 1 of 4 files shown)


Akuli (approved):
Thanks!

PR #1210, opened 2022-10-25
Worked on plugin manager

### LINKED ISSUE #1208: Plugin manager shouldn't be a plugin
Disabling all plugins in the plugin manager (Settings menu at top, Plugin Manager, press Ctrl+A, click "Disable") is annoying, because it asks whether you also want to disable the plugin manager. You probably don't, because if you disable it, the only way to get it back is to edit a `.json` file by hand.

To fix this, the plugin manager shouldn't be a plugin. Specifically, we should:
- Move `porcupine/plugins/pluginmanager.py` so that it is not in the `plugins` folder
- Delete `setup()` function in that file (a function named `setup()` is called when a plugin is loaded, see https://akuli.github.io/porcupine/plugin-intro.html )
- The `setup()` function only adds the menu item to the Settings menu. We should do that somewhere else during Porcupine startup, probably `_fill_menus_with_default_stuff()` in `menubar.py`.

aloner-pro:
Worked on issue #1208 but I think it didn't work. Would like a review.

Fixes #1208

### DIFF
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -423,9 +423,19 @@ comment_prefix = '//'
 autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = 'this regex matches nothing^'}
 [[Rust.example_commands]]
 command = "cargo run"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo test"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo build"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo clippy"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo fmt"
+working_directory = "{project_path}"
+[[Rust.example_commands]]
 command = "cargo check"
 working_directory = "{project_path}"
--- a/porcupine/default_filetypes.toml
+++ b/porcupine/default_filetypes.toml
@@ -421,3 +421,11 @@ tabs2spaces = true
 indent_size = 4
 comment_prefix = '//'
 autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = 'this regex matches nothing^'}
+[[Rust.example_commands]]
+command = "cargo run"
+command = "cargo test"
+command = "cargo build"
+command = "cargo clippy"
+command = "cargo fmt"
+command = "cargo check"
+working_directory = "{project_path}"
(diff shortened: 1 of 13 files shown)


porcupine/menubar.py
```diff
@@ -449,3 +449,7 @@ def add_link(menu_path: str, label: str, url: str) -> None:
         "https://github.com/Akuli/python-tutorial/blob/master/README.md",
     )
     add_link("Help/Python", "Official documentation", "https://docs.python.org/")
+
+    def setup() -> None:
+        pluginmanager.get_menu("Settings").add_command(label="Plugin Manager", command=show_dialog)
```
Akuli, line 454: This shouldn't be in a `setup()` function anymore. It should look very similar to this line above that creates the "Porcupine Settings" dialog:

```python3
get_menu("Settings").add_command(label="Porcupine Settings", command=settings.show_dialog)
```

and probably be just after that line, so that it's nicely in the same place with other related code.

porcupine/pluginmanager.py
```diff
@@ -263,5 +263,3 @@ def create_dialog() -> tuple[tkinter.Toplevel, PluginDialogContent]:
     return (dialog, content)  # content returned for tests
```
Akuli, line 263: Earlier in this file there is code that asks "Do you really want to disable this plugin manager" if you try to disable it. We can delete it now, because disabling it is no longer possible.

porcupine/default_filetypes.toml
```diff
@@ -421,3 +421,21 @@ tabs2spaces = true
 indent_size = 4
 comment_prefix = '//'
 autoindent_regexes = {dedent = '(return( .+)?|break|continue);', indent = 'this regex matches nothing^'}
+[[Rust.example_commands]]
```
Akuli, line 424: These show up as a part of this PR, because you haven't pulled from my master branch in a while. See e.g. https://akuli.github.io/git-guide/pr.html#making-another-pr. The tl;dr is that you need to run:

```
git pull https://github.com/Akuli/porcupine
```

aloner-pro: I forgot to do it after merging this I will surely do it.

Akuli:
To get the CI green, you need to:
- Fix the failing tests (to run the tests locally: `python3 -m pytest tests/test_pluginmanager.py`)
- Run black (`black porcupine`, maybe also `black tests`)
- Fix mypy errors and pyflakes warnings

aloner-pro:
After some reformating `mypy`, `black` and `pyflakes` have no warnings.
But with `pytest` one test is failing 
```
FAILED tests/test_pluginmanager.py::test_disable_itself - _tkinter.TclError: Item pluginmanager not found
```

Akuli:
`test_disable_itself()` is about disabling the plugin manager with the plugin manager. That is no longer possible, because the plugin manager isn't a plugin, so you can delete the test and the related code.

Akuli (approved):
Thanks!

PR #1212, opened 2022-10-29
nsis: Less verbosity when deleting recursively

### LINKED ISSUE #1174: Installer removes existing installation slowly
To reproduce
1. Install Porcupine with the Windows installer
2. Install Porcupine with the Windows installer again

Presumably the deleting is slow because a log line is printed for every deleted file. We could set the log verbosity: https://nsis.sourceforge.io/Reference/SetDetailsPrint

Akuli:
Fixes #1174. Not tested yet

### DIFF
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -68,7 +68,7 @@ Section "!Porcupine" sec_app
   SectionIn RO
 
   ; see #328
-  DetailPrint "Deleting $INSTDIR\Python..."
+  DetailPrint "Deleting $INSTDIR\Python"
   SetDetailsPrint textonly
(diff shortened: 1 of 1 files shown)


scripts/installer.nsi
```diff
@@ -67,7 +67,12 @@ Section "!Porcupine" sec_app
   SetRegView 64
   SectionIn RO
 
-  RMDir /r "$INSTDIR\Python"  ; see #328
+  ; see #382
```
Akuli, line 70:
```suggestion
  ; see #328
```

Akuli (commented):
Showing "..." in UI seems like a bad idea, sometimes makes it look like there's more and it's showing only a part of a path.

Issue #92, opened 2020-08-12
different porcupine processes don't communicate together


Akuli:
it's possible to have the same file opened at the same time in two different porcupines, that can be confusing, especially because porcupine doesn't detect if another program has changed a file (related to #30 )

there's porcupine/_ipc.py but that appears to be something that i decided to not use for whatever reason

Akuli:
also creates funny behaviour with porcupine.settings

rdbende:
Yesterday I made a shell launcher script and a desktopfile for Porcupine, so I can launch it from my panel, and assign it to files to open with. However it's very annoying, that whenever I double-click on a file, it opens a new Porcupine instance.
Ideally there could be an setting to `Allow only single Porcupine instance`, but at least a `--new-tab` option.

Akuli:
Not being able to drag tabs into an existing Porcupine instance makes this even worse: once you have two files you want to work on in separate porcupines, the only way to bring them under the same porcupine is to open the file again.

PR #1215, opened 2022-11-03
Communicate between Porcupine instances

rdbende:
This breaks the `poppingtabs` plugin, so some `--new-window` command line option is needed.
Would be also good to have a checkbox in the settings window to turn off this feature.

### DIFF
--- a/porcupine/__main__.py
+++ b/porcupine/__main__.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
+
 import argparse
 import logging
 import sys
 from pathlib import Path
+from typing import Iterable
 
 from porcupine import __version__ as porcupine_version
 from porcupine import (
@@ -27,6 +30,28 @@ Examples:
   %(prog)s --no-plugins       # understand the power of plugins
   %(prog)s -v                 # produce lots of output for debugging
 """
+
+
+def open_files(files: Iterable[str]) -> None:
+    tabmanager = get_tab_manager()
+    for path_string in files:
+        if path_string == "-":
+            # don't close stdin so it's possible to do this:
+            #
+            #   $ porcu - -
+            #   bla bla bla
+            #   ^D
+            #   bla bla
+            #   ^D
+            tabmanager.add_tab(tabs.FileTab(tabmanager, content=sys.stdin.read()))
+        else:
+            tabmanager.open_file(Path(path_string))
+
+
+def open_files_from_tk_send(*files: str) -> None:
+    open_files(files)
+    get_main_window().deiconify()
+    get_main_window().lift()
 
 
 def main() -> None:
@@ -137,23 +162,13 @@ def main() -> None:
     # Prevent showing up a not-ready-yet root window to user
     get_main_window().withdraw()
 
+    get_main_window().createcommand("open_files", open_files_from_tk_send)
+
     settings.init_the_rest_after_initing_enough_for_using_disabled_plugins_list()
     menubar._init()
     pluginloader.run_setup_functions(args.shuffle_plugins)
 
-    tabmanager = get_tab_manager()
-    for path_string in args.files:
-        if path_string == "-":
-            # don't close stdin so it's possible to do this:
-            #
-            #   $ porcu - -
-            #   bla bla bla
-            #   ^D
-            #   bla bla
-            #   ^D
-            tabmanager.add_tab(tabs.FileTab(tabmanager, content=sys.stdin.read()))
-        else:
-            tabmanager.open_file(Path(path_string))
+    open_files(args.files)
 
     get_main_window().deiconify()
     try:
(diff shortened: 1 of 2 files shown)


porcupine/__main__.py
```diff
@@ -29,6 +32,28 @@
 """
 
 
+def open_files(files: Iterable[str]) -> None:
+    tabmanager = get_tab_manager()
+    for path_string in files:
+        if path_string == "-":
```
Akuli, line 38: Presumably this doesn't work when opening from `tk send`. It would be nice if we could fix it (here or in a separate PR), so that you could do

```
$ some long command | porcu -
```

to show the output of the command in the currently running Porcupine instance.

Akuli (commented):
This is awesome! I think we should make a new release before merging this, so that we can add `--new-window` or whatever the option will be in a separate PR.

Akuli:
> Would be also good to have a checkbox in the settings window to turn off this feature.

I don't think this is needed, at least not for now. We can always add it later if someone needs it.

porcupine/__main__.py
rdbende: Shall we use an ipc file for it, or do you have some other idea?

rdbende:
Ohh shit. Looks like this will be an X-only feature.

image
The `send` command doesn't even exists on Windoze, and `winfo interps` returns nothing 🤦🏼‍♀️

On OSX `send` exists, although it does nothing, and `winfo interps` returns only the current Tcl interpeter, but not others.

Akuli:
You can revert #786.

rdbende:
I don't think that's how we want to do it. Especially because most of us use Porcupine on Linux.

Moosems:
What do you mean by "Communicate between Porcupine instances" then?

rdbende:
`Allow only a single Porcupine instance` and #92

Akuli:
Ideally you could still have multiple Porcupine instances running. It just wouldn't be the thing that happens when you open files by double-clicking each one individually, instead they would all go to the same instance/process/window.

porcupine/__main__.py
rdbende: Yep, it doesn't

porcupine/_state.py
```diff
@@ -44,6 +46,41 @@ def _log_tkinter_error(
...
+            #   bla bla
+            #   ^D
+            tabmanager.file_queue.put(sys.stdin.read())
+        else:
+            tabmanager.file_queue.put(Path(path_string))
+
+    tabmanager.event_generate("<<OpenFilesQueue>>")
+
+
+def listen_for_files():
```
rdbende, line 71: This could be merged with what's already in `_ipc.py`

porcupine/_state.py
```diff
@@ -44,6 +46,41 @@ def _log_tkinter_error(
...
+    with _ipc.session() as ipc_message_queue:
+        while True:
+            message = ipc_message_queue.get()
+            if message is Quit:
+                break
+            else:
+                try:
+                    open_files([message])
```
Akuli, line 79: I don't like how this invokes tkinter from a thread. The usual "pattern" is to repeatedly check every 50ms or 100ms or so with `after_idle()`.

Akuli:
Nice to see some progress on this :)

porcupine/_state.py
rdbende: Even if it's only a `event_generate`? It doesn't hurt 🥺

Akuli: My understanding is that Tcl calls (which ultimately all of tkinter boils down to) aren't thread-safe. Feel free to correct me if you have read `_tkinter.c` more carefully than I have :)

rdbende: > you have read `_tkinter.c` more carefully than I have :)

Maybe I did :) But the only interesting thing I found out, is that while a `tk.call` will marshal the call to the interpreter thread, `tk.eval` doesn't do that.

Although the only case that would make `event_generate` thread-unsafe that I can think of is a race-condition on the event queue, there is indeed nothing that would make it explicitly thread-safe (and I assume that the event queue is thread-safe if the Tcl of Tkinter is threaded - which it is mostly these days). But for example, while a `label.config()` always throws an error if not called from the main thread ("Calling Tcl from different apartment"), with `event_generate` this never happened.

rdbende: Resolved. It also simplified the code quite a bit.

rdbende:
Oh man, what I just did to all the tests

Moosems:
Uh oh man what'd you do 🙄😂?

rdbende:
Idk. Something with the cache directory being elsewhere on GH actions. It works locally.

ethical-haquer:
Any updates on this?

rdbende:
It works. The test need fixing, but otherwise it's good, AFAIR.

Issue #1074, opened 2022-05-26
navigating Porcupine without mouse


Akuli:
Porcupine has 3 main widgets where you will likely want to have your keyboard focus:
- The editing area itself
- The output of commands at bottom (press F4 to show/hide)
- The directory tree (at left)

Currently it's possible to navigate between these without clicking the mouse:
- To focus directory tree, press Alt+Shift+T.
- To focus the file you're editing, press Ctrl+PageDown followed by Ctrl+PageUp. This toggles between two tabs, and when switching a tab, the text widget is focused automatically. If you only have one tab opened, you need to open another one for this to work.
- To focus the output, press Alt+Shift+T followed by Shift+Tab.

Ideally there would be separate key bindings for all of these. For example:
- To focus the directory tree, press Alt+Shift+T.
- To focus the file you're editing, press Alt+Shift+F (F for File, could also be e.g. E for Editing).
- To focus the output of commands, press Alt+Shift+C (C for Commands, not O because it's too far away on keyboard).

Akuli:
Maybe the key bindings should be Alt+Shift+D, Alt+Shift+F and Alt+Shift+C. The location of D, F and C keys on the keyboard resembles the UI layout: editing area (F) is on the right side of directory tree (D), and command output (C) is below.

lawson89:
@Akuli are all these keybindings setup in default_keybindings.tcl?

rdbende:
Currently only the directory tree has a keybinding. It's defined here:

https://github.com/Akuli/porcupine/blob/4c52560f02c6d6eaffcbd34b196c975b1a5f39f2/porcupine/default_keybindings.tcl#L86

lawson89:
I'm happy to try to fix this  issue but would need some guidance. I don't know TCL at all but would I add two new keybindings in default_keybinds.tcl and two new menu entries for <<View/Focus active file>> and <<View/Focus command output>> and then wire up the event handling on the python side?

Akuli:
Setting this up is really quite simple. The code examples below are currently in Porcupine and can be found with `git grep "Focus directory tree"`. 

You need to create the menu item. You don't need to specify anything else than the text (label) and a callback function (command).

```python
    menubar.get_menu("View").add_command(
        label="Focus directory tree", command=partial(_focus_treeview, tree)
    )
```

(You can set `command` to whatever you need, here the purpose of `partial` is to pass the `tree` as an argument to it)

[...]

PR #1217, opened 2022-11-13
navigating Porcupine without mouse

lawson89:
#1074

### DIFF
--- a/porcupine/menubar.py
+++ b/porcupine/menubar.py
@@ -389,6 +389,11 @@ def _fill_menus_with_default_stuff() -> None:
         if tab.can_be_closed():
             get_tab_manager().close_tab(tab)
 
+    def focus_active_tab() -> None:
+        tab = get_tab_manager().select()
+        if tab is not None:
+            tab.event_generate("<<TabSelected>>")
+
     get_menu("File").add_command(label="New File", command=new_file)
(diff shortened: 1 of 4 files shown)


Akuli (commented):
Thanks, this is awesome!

In the menus, maybe we should put all focus things under the same submenu, e.g.:

View --> Focus --> Active file
View --> Focus --> Directory tree
View --> Focus --> Command output

This way each plugin can add its own thing to this submenu, and when browsing the menus, you see all the focusing options at once.

porcupine/menubar.py
```diff
@@ -422,6 +427,7 @@ def change_font_size(how: Literal["bigger", "smaller", "reset"]) -> None:
     get_menu("View").add_command(
         label="Reset Font Size", command=partial(change_font_size, "reset")
     )
+    get_menu("View/Focus").add_command(label="Active file", command=focus_active_tab)
```
Akuli, line 430: It would be good to disable this when there are no tabs. This way confused users won't think it does something if they browse the menus with no tabs open. You can use `set_enabled_based_on_tab()` to do this.

Akuli (commented):
Thanks! One more nit.

Akuli (approved):
Nice

Akuli:
Test failure is unrelated to this PR.

PR #1226, opened 2022-12-16
Run plugin: add memory limit

Akuli:
On linux, it's annoying when my dumb program allocates so much memory that the entire computer freezes. This is because linux decides that it's a good idea to use swap for more memory. Swap basically means using the hard drive as if it was RAM, and that's really really slow.

To help with this, add a setting that limits how much memory the program can allocate. The setting shows up in the run dialog, but it's saved to Porcupine's settings so that it won't change if you run a different command.

### DIFF
--- a/porcupine/plugins/run/dialog.py
+++ b/porcupine/plugins/run/dialog.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import re
 import sys
 import tkinter
 from tkinter import ttk
 from typing import Callable, TypeVar
 
 from porcupine import get_main_window, textutils, utils
+from porcupine.settings import global_settings
 
 from . import common, history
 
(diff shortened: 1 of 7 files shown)


porcupine/plugins/run/dialog.py
```diff
@@ -56,6 +58,51 @@ def validate(self, *junk_from_var_trace: object) -> None:
...
+
+    # Show 2GB instead of 2.0GB
+    if number == int(number):
+        number = int(number)
+
+    return str(number) + suffix
+
+
+def _string_to_mem_limit(string: str) -> int | None:
+    match = re.fullmatch(r"([0-9]+(?:\.[0-9]+)?)([KMG]?)B?", string)
```
Akuli, line 83:
```suggestion
    match = re.fullmatch(r"([0-9]+(?:\.[0-9]+)?)([KMG]?)B?", string.replace(" ", "").upper())
```

Akuli:
This might not be a great idea, because if I understand the `preexec_fn` warning in docs correctly, it can freeze Porcupine even when we don't run out of memory.

Akuli:
I have a better idea for implementing a memory limit, but it will require some changes.

First, we should change `no_terminal.py` so that it uses `bash_run.sh` (except on Windows, of course). Currently `bash_run.sh` is invoked only when a command is ran in an external terminal.

Once that is done we can simply add a `ulimit` command to `bash_run.sh`, although it will need to take a parameter describing what memory usage to apply. It already takes one parameter indicating whether to wait for user's enter press once the command is done.

PR #1228, opened 2022-12-21
Support user input when running commands with output in Porcupine window

### LINKED ISSUE #1225: Built in runner doesn't allow input

Akuli:
Fixes #1225


Akuli:
It is currently a working proof of concept.

Next steps (probably for me or @rdbende, whoever happens to have time and feel motivated to do this first):
- Look at `diff -u --color porcupine/plugins/run/no_terminal.py input-works-yay.py` and copy the relevant changes to `porcupine/plugins/run/no_terminal.py`. Make sure it works.
- Delete `input-works-yay.py`
- Figure out what color user input so that it looks nice (can set color on text widget, since everything except user input is now tagged with `"output"` and we can override the color there)
- Get CI green

Akuli:
I pressed Shift+F5 and typed `python3 -i`. It works! (although the colors aren't great and it gives an error if you press Enter after the process exits)

Screenshot at 2023-04-13 22-56-21

rdbende:
image

Link underlines are now colored like the input, but I like it, tbh

Akuli:
I like how this works!

The only thing I don't like is how `TerminalTextWidget` and `NoTerminalRunner` reference each other in a cyclic way: `TerminalTextWidget` needs to access the runner's executor to know whether a process is currently running, so it knows whether user can write or not.

I think I will move some of the code from `TerminalTextWidget` to `NoTerminalRunner`

porcupine/plugins/run/no_terminal.py
```diff
@@ -44,7 +46,7 @@
     # blah blah: some_function (filename.c:123)
     r"\(([^\n():]+):([0-9]+)\)",
 ]
-filename_regex = "(?i)" + "|".join(r"(?:" + part + r")" for part in filename_regex_parts)
+filename_regex = "|".join(r"(?:" + part + r")" for part in filename_regex_parts)
```
Akuli, line 49:
```suggestion
filename_regex = "(?i)" + "|".join(r"(?:" + part + r")" for part in filename_regex_parts)
```

porcupine/plugins/run/no_terminal.py
```diff
@@ -281,6 +342,79 @@ def __init__(self, master: tkinter.Misc) -> None:
...
+            if action == "tagon" and tag_or_text == "output":
+                tag_on = True
+            elif action == "tagoff" and tag_or_text == "output":
+                tag_on = False
+            elif action == "text" and not tag_on:
+                text_chunks.append(tag_or_text)
+
+        return "".join(text_chunks)
+
+    # TODO: combine this with handle_enter_press()
```
Akuli, line 399: should do this

porcupine/plugins/run/no_terminal.py
```diff
@@ -242,20 +239,84 @@ def stop(self, *, quitting: bool = False) -> None:
...
+        if self._shell_process is not None:
+            assert self._shell_process.stdin is not None
+            self._shell_process.stdin.close()
+
+
+# A python-edit-tracking text widget lets you know whether the text is currently
+# being edited through invoking a method in Python, such as .insert(). This lets
+# you distinguish between default key bindings and whatever your code does.
+class PyEditTrackingText(tkinter.Text):
+    @copy_type(tkinter.Text.__init__)
```
Akuli, line 258: I wonder which is better `copy_type` or `functools.wraps`.

Akuli (commented):
Only skimmed

Akuli:
- [ ] Add a test

porcupine/plugins/run/no_terminal.py
Akuli: `copy_type` is better because `wraps` infers the type from what it's decorating. So it becomes full of `Any`.

Akuli:
I will need to try this on Windows to understand the test failure.

Issue #1213, opened 2022-10-29
Branch renamed: master --> main


Akuli:
This is more of an announcement for developers than a bug report.

Many projects now have a `main` branch instead of a `master` branch, so I renamed Porcupine's `master` branch to `main`. If you have cloned Porcupine earlier, you can run these commands to make Git not confused about it (copied from GitHub's message popup):

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

Or you can simply delete your local clone of Porcupine and run `git clone` again.

PR #1231, opened 2022-12-29
Change master to main in a few places

Akuli:


Moosems:
Why?

Akuli:
I renamed the branch a while ago (#1213). GitHub still redirects from "master" URLs to "main" URLS (e.g. https://github.com/Akuli/porcupine/blob/master/LICENSE redirects to https://github.com/Akuli/porcupine/blob/main/LICENSE ), but I think it's nice to be consistent about it.

Moosems:
Why'd you change it in the first place?

Akuli:
To me, the main reason is that "the main branch" now makes more sense to most contributors than "the master branch", because many other projects have changed to `main` and it's the default in github. There are other reasons too, search for e.g. "why main instead of master branch".

PR #1234, opened 2023-01-17
Add a desktop launcher creating plugin

Akuli:
Installing Porcupine is a bit of a mess on Linux. To run it, according to README, you need two commands:

```
source porcupine-venv/bin/activate
porcu
```

This PR adds a way to create a launcher to your desktop that you can simply double-click.

### DIFF
--- a/porcupine/plugins/desktop_launcher.py
+++ b/porcupine/plugins/desktop_launcher.py
@@ -43,7 +43,7 @@ class CreateLauncherAction(argparse.Action):
         with launcher_path.open("w") as file:
             file.write("[Desktop Entry]\n")
             file.write("Name=Porcupine\n")
-            file.write(f"Exec={command}\n")
+            file.write(f"Exec=bash -c {shlex.quote(command)}\n")
             file.write("Terminal=false\n")
             file.write("Type=Application\n")
             file.write(f"Icon={images.images_dir}/logo-200x200.gif\n")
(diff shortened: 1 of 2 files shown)


porcupine/plugins/desktop_launcher.py
```diff
@@ -0,0 +1,69 @@
...
+
+        activate_path = Path(venv) / "bin" / "activate"
+        assert activate_path.is_file()
+
+        command = f"source {shlex.quote(str(activate_path))} && porcu"
+
+        with launcher_path.open("w") as file:
+            file.write("[Desktop Entry]\n")
+            file.write("Name=Porcupine\n")
+            file.write(f"Exec={command}\n")
```
Akuli, line 46:
```suggestion
            file.write(f"Exec=bash -c {shlex.quote(command)}\n")
```

Akuli:
@rdbende What do you think? I think you have mentioned `.desktop` files before.

rdbende:
Oh, this is a nice feature!
Though I don't use desktop icons to launch programs, but rather desktop menu items. I have a plugin that adds a checkbox in the settings window, named "Show Porcupine in the desktop menu". It uses `xdg-desktop-menu` to install and uninstall the menu item. A thing like this would also be a good addition imo.

Akuli:
Let's add your plugin instead :) Want to make a PR?

taahol:
I guess that would work reasonably well on non-Linux Unix-like systems, too, as long as XDG tools are available.

PR #1235, opened 2023-01-19
“Open in terminal” right click option

### LINKED ISSUE #1042: More right-click options

taahol:
Tested on Linux only.

Fixes #1042

### DIFF
--- a/porcupine/plugins/run/terminal.py
+++ b/porcupine/plugins/run/terminal.py
@@ -24,8 +24,12 @@ else:
 
 
 # getting this to work in powershell turned out to be hard :(
-def _run_in_windows_cmd(command: str, cwd: Path, env: dict[str, str]) -> None:
+def _run_in_windows_cmd(command: str | None, cwd: Path, env: dict[str, str]) -> None:
     log.debug("using Windows command prompt")
+
+    if not command:
+        subprocess.Popen(["cmd", "/K", f"cd /d {shlex.quote(str(cwd))}"], env=env)
+        return
 
     real_command = [str(utils.python_executable), str(run_script), str(cwd), command]
     if not utils.running_pythonw:
@@ -37,8 +41,13 @@ def _run_in_windows_cmd(command: str, cwd: Path, env: dict[str, str]) -> None:
     subprocess.Popen(real_command, env=env)
 
 
-def _run_in_macos_terminal_app(command: str, cwd: Path, env: dict[str, str]) -> None:
+def _run_in_macos_terminal_app(command: str | None, cwd: Path, env: dict[str, str]) -> None:
     log.debug("using MacOS terminal.app")
+
+    if not command:
+        subprocess.Popen(["open", "-a", "Terminal", str(cwd)], env=env)
+        return
+
     assert shutil.which("bash") is not None
 
     with tempfile.NamedTemporaryFile("w", delete=False, prefix="porcupine-run-") as file:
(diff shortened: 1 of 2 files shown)


porcupine/plugins/run/terminal.py
```diff
@@ -24,9 +24,13 @@
 
 
 # getting this to work in powershell turned out to be hard :(
-def _run_in_windows_cmd(command: str, cwd: Path, env: dict[str, str]) -> None:
+def _run_in_windows_cmd(command: str | None, cwd: Path, env: dict[str, str]) -> None:
     log.debug("using Windows command prompt")
 
+    if not command:
+        subprocess.Popen(["cmd", "/K", f"cd /d {shlex.quote(str(cwd))}"], env=env)
```
Akuli, line 31: `shlex.quote` uses single quotes which doesn't work on windows. But `subprocess.list2cmdline` is a Windows-specific (undocumented) function that does this:

```
>>> shlex.quote('foo bar')
"'foo bar'"
>>> subprocess.list2cmdline(['cd', '/d', 'foo bar'])
'cd /d "foo bar"'
```

```suggestion
        cd_command = subprocess.list2cmdline(["cd", "/d", str(cwd)])
        subprocess.Popen(["cmd", "/K", cd_command], env=env)
```

porcupine/plugins/run/terminal.py
```diff
@@ -37,8 +41,13 @@ def _run_in_windows_cmd(command: str, cwd: Path, env: dict[str, str]) -> None:
     subprocess.Popen(real_command, env=env)
 
 
-def _run_in_macos_terminal_app(command: str, cwd: Path, env: dict[str, str]) -> None:
+def _run_in_macos_terminal_app(command: str | None, cwd: Path, env: dict[str, str]) -> None:
     log.debug("using MacOS terminal.app")
+
+    if not command:
+        subprocess.Popen(["open", "-a", "Terminal", str(cwd)], env=env)
```
Akuli, line 48: This is how you launch a script: `open -a Terminal <scriptfile>`. Does the same command work for directories as well? I don't have a mac so I can't test this myself, and I also failed googling this.

porcupine/plugins/run/terminal.py
```diff
@@ -116,16 +125,22 @@ def _run_in_x11_like_terminal(command: str, cwd: Path, env: dict[str, str]) -> N
         )
         return
 
-    real_command = [str(run_script), str(cwd), command]
-    real_command.extend(map(str, command))
-    subprocess.Popen([terminal, "-e", " ".join(map(shlex.quote, real_command))], env=env)
+    if command:
+        real_command = [str(run_script), str(cwd), command]
+        real_command.extend(map(str, command))
```
Akuli, line 130: `real_command.extend(map(str, command))` is dumb and weird:

```
>>> command = "python3 foo.py"
>>> real_command = ["/path/to/script.sh", "/home/akuli", command]
>>> real_command.extend(map(str, command))
>>> real_command
['/path/to/script.sh', '/home/akuli', 'python3 foo.py', 'p', 'y', 't', 'h', 'o', 'n', '3', ' ', 'f', 'o', 'o', '.', 'p', 'y']
>>> " ".join(map(shlex.quote, real_command))
"/path/to/script.sh /home/akuli 'python3 foo.py' p y t h o n 3 ' ' f o o . p y"
```

We can keep it in this PR because it was there already and running commands seems to work regardless.

Akuli (commented):
Thanks!

porcupine/plugins/run/terminal.py
taahol: It went askew in https://github.com/Akuli/porcupine/commit/e44d4aae36030ffa65dbab3e9cff65e5fd6712a0

Akuli (approved):
awesome :)

PR #1237, opened 2023-01-20
Add desktop menu entry creator plugin

rdbende:


porcupine/plugins/desktop_menu.py
```diff
@@ -0,0 +1,63 @@
...
+
+def install_uninstall_desktop_file(*dontcare) -> None:
+    if global_settings.get("has_desktop_file", bool):
+        venv = os.environ.get("VIRTUAL_ENV")
+        if not venv or not (Path(venv) / "bin" / "porcu").is_file():
+            messagebox.showerror(
+                "Creating menu entry failed",
+                "Porcupine must be installed in a virtual environment in order to create a desktop menu entry.",
+            )
+            global_settings.set("has_desktop_file", False)
```
rdbende, line 44: The checkbutton still remains checked here. It would be good to uncheck it if the desktopfile can't be installed.

porcupine/plugins/desktop_menu.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+    else:
+        subprocess.call([XDG_DESKTOP_MENU, "uninstall", "--mode", "user", "Porcupine.desktop"])
+
+
+def setup() -> None:
+    if not shutil.which("xdg-desktop-menu"):
+        return
+
+    global_settings.add_option("has_desktop_file", False)
```
Akuli, line 59: IMO this doesn't belong in `global_settings`. You should instead check whether the file exists, and set the checkbutton based on that. Otherwise the user can delete the desktop file manually and then wonder why the checkbutton is still checked, for example.

To create a checkbutton that doesn't use `global_settings`, see the code of the `add_checkbutton()` function.

Akuli: lol, there's no way to check if installed...

rdbende: We can check if it exists in `~/.local/share/applications` or in `/usr/share/applications`, although I'm sure there's a DE or WM where this doesn't work.

Akuli:
@rdbende I changed a few things. What do you think?

Akuli (approved):
Thanks for working on this!

PR #1238, opened 2023-01-21
Fix encoding issues in plugins/drop_to_open.py

### LINKED ISSUE #1233: ä or ö in file name + drop_to_open plugin + linux
trying to drag file `/tmp/jatzy-harjoitustyö.py` into Porcupine it does not open. On terminal says:

```
porcupine.plugins.drop_to_open WARNING: can't open '/tmp/jatzy-harjoitustyÃ¶.py' because it is not a file
```

taahol:
fixes #1233


Akuli (approved):
Wow... So simple and elegant!

PR #1241, opened 2023-01-26
Check the BOM characters

### LINKED ISSUE #1239: UTF-8 BOM
lol

Porcupine doesn't detect a UTF-8 BOM at the start  of a file. Instead it reads it as a normal character which happens to be `U+FEFF ZERO WIDTH NO-BREAK SPACE`. Also, it is impossible to delete this character by pressing backspace (but the delete key works).

To reproduce:

```
>>> open('/tmp/asd.txt','w',encoding='utf-8-sig').write('moi')
```

taahol:
fixes #1239

### DIFF
--- a/porcupine/tabs.py
+++ b/porcupine/tabs.py
@@ -691,17 +691,18 @@ class FileTab(Tab):
         while True:
             encoding = self.settings.get("encoding", str)
             try:
-                charset_match = charset_normalizer.from_path(self.path).best()
-                if charset_match:
-                    encoding_match = charset_match.encoding.replace("_", "-")
-                    if encoding_match == "utf-8" and charset_match.byte_order_mark:
-                        encoding_match = "utf-8-sig"
-                    if encoding != encoding_match:
-                        bad_encoding = encoding
-                        encoding = encoding_match
-                        raise UnicodeDecodeError(
-                            bad_encoding, b"\x00\x00", 1, 2, "This is just a fake reason!"
-                        )
+                charset_matches = [
+                    match.encoding.replace("_", "-") + "-sig"
+                    if match.byte_order_mark
+                    else match.encoding.replace("_", "-")
+                    for match in charset_normalizer.from_path(self.path)
+                ]
+                if encoding not in charset_matches:
+                    bad_encoding = encoding
+                    encoding = charset_matches[0]
+                    raise UnicodeDecodeError(
+                        bad_encoding, b"\x00\x00", 1, 2, "This is just a fake reason!"
+                    )
                 with self.path.open("r", encoding=self.settings.get("encoding", str)) as f:
                     stat_result = os.fstat(f.fileno())
                     content = f.read()
--- a/porcupine/tabs.py
+++ b/porcupine/tabs.py
@@ -599,8 +599,9 @@ class FileTab(Tab):
         )
         self.settings.add_option("tabs2spaces", True)
         self.settings.add_option("indent_size", 4)
+        self.settings.add_option("encoding", "utf-8")
         # Invoke encoding detection early to suppress the unsaved changes warning
-        self.settings.add_option("encoding", self._detect_encoding())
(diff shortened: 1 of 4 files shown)


taahol:
It failed the Latin-1 “mörkö” test.

taahol:
Maybe I will add some tests.

porcupine/tabs.py
```diff
@@ -687,8 +688,31 @@ def reload(self, *, undoable: bool = True) -> bool:
...
+            return utils.ask_encoding(
+                f'The content of "{self.path}" is not valid '
+                + self.settings.get("encoding", str)
+                + ". Choose an encoding to use instead:",
+                encoding,
+            )
+
         while True:
             try:
+                if self.settings.get("encoding", str) == "utf-8":
```
Akuli, line 701: Why not always look at the BOM?

porcupine/tabs.py
```diff
@@ -687,8 +688,31 @@ def reload(self, *, undoable: bool = True) -> bool:
...
+                        byte_str = fb.read(4)
+                    for bytez, guessed_encoding in [
+                        (codecs.BOM_UTF8, "utf-8-sig"),
+                        (codecs.BOM_UTF32_LE, "utf-32"),
+                        (codecs.BOM_UTF32_BE, "utf-32"),
+                        (codecs.BOM_LE, "utf-16"),
+                        (codecs.BOM_BE, "utf-16"),
+                    ]:
+                        if byte_str.startswith(bytez):
+                            user_selected_encoding = confirm_encoding(guessed_encoding)
```
Akuli, line 712: IMO we don't need to ask the user here. If we have a BOM, we are pretty sure we guessed the right encoding and there's no reason to ask the user.

porcupine/tabs.py
```diff
@@ -709,12 +733,7 @@ def reload(self, *, undoable: bool = True) -> bool:
                         continue
 
             except UnicodeDecodeError:
-                bad_encoding = self.settings.get("encoding", str)
-                user_selected_encoding = utils.ask_encoding(
-                    f'The content of "{self.path}" is not valid {bad_encoding}. Choose an encoding'
-                    " to use instead:",
-                    bad_encoding,
-                )
+                user_selected_encoding = confirm_encoding("latin1")
```
Akuli, line 736: Why does this hard-code latin1, an encoding used mostly in just one part of the world?

Akuli (commented):
Thanks! This looks pretty good overall. A few questions.

porcupine/tabs.py
```diff
@@ -598,7 +599,7 @@ def __init__(self, manager: TabManager, content: str = "", path: Path | None = N
         )
         self.settings.add_option("tabs2spaces", True)
         self.settings.add_option("indent_size", 4)
-        self.settings.add_option("encoding", "utf-8")
+        self.settings.add_option("encoding", self._detect_encoding())
```
Akuli, line 602: I think we don't need to call `_detect_encoding()` here, because it will be called on the first reload anyway. (The `open_file()` method calls `reload()`)

Another thought: I found this comment, which made me think it should be possible to make the BOM detection a plugin.

```python3
        # Add tab before loading content, so that editorconfig plugin gets a
        # chance to set the encoding into tab.settings
```
That can be a separate PR: this PR is very good as is :)

Akuli (commented):
It's looking pretty good :) Just a couple more thoughts.

porcupine/tabs.py
Akuli: If we keep this line, `"utf-8"` should still probably be the default value. I think it is currently not possible to reset a setting to its default, but changing the default value of the setting might have weird consequences now or in a future version of Porcupine.

```suggestion
        self.settings.add_option("encoding", "utf-8")
        self.settings.set("encoding", self._detect_encoding())
```

taahol:
Looks pretty good!

Akuli (approved):
Awesome

PR #1242, opened 2023-01-28
Don't forget the filetype

### LINKED ISSUE #1168: Filetype isn't remembered when restarting
1. Make sure "Remember open tabs when Porcupine is closed and reopened" is checked (in menubar --> Settings --> Porcupine Settings)
2. Make a new file (Ctrl+N)
3. Write some C code, e.g. `short int x;`. It will highlight poorly with the default Python filetype.
4. Set filetype to C. This will cause the code to highlight much better.
5. Close Porcupine (do not save the file first)
6. Reopen Porcupine

Currently filetype seems to reset back to python: the newly created file highlights badly again, and you need to select C from the "Filetypes" menu again. It should remember that the filetype was C.

Fixing this needs some time spent debugging and understanding what is happening. Relevant files are:
- `porcupine/plugins/restart.py` for the logic of how opened files are remembered
- `porcupine/plugins/tabs.py` defines `FileTab.get_state()`
- `porcupine/plugins/filetypes.py` contains all filetypes related code.

To see all log messages from a specific file, you can use for example:

```
$ python3 -m porcupine --verbose-logger porcupine.plugins.filetypes
```

taahol:
fixes #1168

### DIFF
--- a/porcupine/plugins/filetypes.py
+++ b/porcupine/plugins/filetypes.py
@@ -231,7 +231,8 @@ def on_path_changed(tab: tabs.FileTab, junk: object = None) -> None:
 
 def on_new_filetab(tab: tabs.FileTab) -> None:
     tab.settings.add_option("filetype_name", None, type_=Optional[str])
-    on_path_changed(tab)
+    if not tab.settings.get("filetype_name", Optional[str]):
+        on_path_changed(tab)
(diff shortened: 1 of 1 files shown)


taahol:
https://cdn.openbsd.org/pub/OpenBSD/songs/song62.ogg

Akuli (approved):

PR #1244, opened 2023-02-13
Simple implementation of a right-click menu

### LINKED ISSUE #1223: Right-click menu
Many things (e.g. search on google, jump to definition, pastebin selected text) would make sense to do by right-clicking the file being edited, instead of going through the menubar or using a memorized key binding.

ArchKats:
Created a plugin for a very simple right-click menu. Currently has the options from the Edit menu from the 
menu bar and the Tool menu.
Kinda fixes #1223.

### DIFF
--- a/porcupine/plugins/jump_to_definition.py
+++ b/porcupine/plugins/jump_to_definition.py
@@ -117,6 +117,4 @@ def generate_jump_request(tab: tabs.FileTab) -> None:
 def setup() -> None:
     get_tab_manager().add_filetab_callback(on_new_filetab)
     menubar.add_filetab_command("Edit/Jump to definition", generate_jump_request)
-    rightclick_menu.add_rightclick_option(
-        "Jump to definition", generate_jump_request
-    )
+    rightclick_menu.add_rightclick_option("Jump to definition", generate_jump_request)
--- a/porcupine/plugins/jump_to_definition.py
+++ b/porcupine/plugins/jump_to_definition.py
@@ -118,5 +118,5 @@ def setup() -> None:
     get_tab_manager().add_filetab_callback(on_new_filetab)
     menubar.add_filetab_command("Edit/Jump to definition", generate_jump_request)
     rightclick_menu.add_rightclick_option(
-        "Jump to definition", generate_jump_request, needs_selected_text=True
+        "Jump to definition", generate_jump_request
     )
--- a/porcupine/plugins/jump_to_definition.py
+++ b/porcupine/plugins/jump_to_definition.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import List
 
 from porcupine import get_tab_manager, menubar, tabs, utils
+from porcupine.plugins import rightclick_menu
 
 log = logging.getLogger(__name__)
 
@@ -116,3 +117,4 @@ def generate_jump_request(tab: tabs.FileTab) -> None:
 def setup() -> None:
     get_tab_manager().add_filetab_callback(on_new_filetab)
     menubar.add_filetab_command("Edit/Jump to definition", generate_jump_request)
+    rightclick_menu.add_rightclick_option("Jump to definition", generate_jump_request, needs_selected_text=True)
--- a/porcupine/plugins/jump_to_definition.py
+++ b/porcupine/plugins/jump_to_definition.py
@@ -117,4 +117,6 @@ def generate_jump_request(tab: tabs.FileTab) -> None:
 def setup() -> None:
     get_tab_manager().add_filetab_callback(on_new_filetab)
     menubar.add_filetab_command("Edit/Jump to definition", generate_jump_request)
-    rightclick_menu.add_rightclick_option("Jump to definition", generate_jump_request, needs_selected_text=True)
+    rightclick_menu.add_rightclick_option(
+        "Jump to definition", generate_jump_request, needs_selected_text=True
+    )
--- a/porcupine/plugins/rightclick_menu.py
+++ b/porcupine/plugins/rightclick_menu.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
 import tkinter
-from typing import Callable, List, Dict
+from typing import Callable, Dict, List, Union
 
 from porcupine import get_main_window, get_tab_manager, tabs
 from porcupine.menubar import get_filetab
 
 _ftab_indexes: List[int] = []
-_menu_items: Dict[str, Callable[[tabs.FileTab], object] | Callable[[], object]] = {}
+_menu_items: Dict[str, Union[Callable[[tabs.FileTab], object], Callable[[], object]]] = {}
 
 
 def text_is_selected(tab: tabs.FileTab) -> bool:
@@ -36,7 +36,7 @@ def show_menu(event: tkinter.Event[tkinter.Misc]) -> None:
 
 def add_rightclick_option(
     path: str,
-    func: Callable[[tabs.FileTab], object] | Callable[[], object],
+    func: Union[Callable[[tabs.FileTab], object], Callable[[], object]],
     needs_selected_text: bool = False,
 ) -> None:
     if needs_selected_text:
--- a/porcupine/plugins/rightclick_menu.py
+++ b/porcupine/plugins/rightclick_menu.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
 import tkinter
-from typing import Callable
+from typing import Callable, List, Dict
 
 from porcupine import get_main_window, get_tab_manager, tabs
 from porcupine.menubar import get_filetab
 
-_ftab_indexes = []
-_menu_items = {}
+_ftab_indexes: List[int] = []
+_menu_items: Dict[str, Callable[[tabs.FileTab], object] | Callable[[], object]] = {}
 
 
 def text_is_selected(tab: tabs.FileTab) -> bool:
@@ -19,7 +19,7 @@ def text_is_selected(tab: tabs.FileTab) -> bool:
     return True
 
 
-def show_menu(event) -> None:
+def show_menu(event: tkinter.Event[tkinter.Misc]) -> None:
     rm = tkinter.Menu(get_main_window(), tearoff=0)
 
     for path, func in _menu_items.items():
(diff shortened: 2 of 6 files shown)


Akuli:
Thanks for working on this! It is trickier than I thought it would be :)

I don't think we should simply add all of edit/tools menus. For example, if you select some text and you right-click something, you probably want to see something like "Search for selected text in the current file" rather than a generic "Find and Replace".

To do this, ideally other plugins would decide what is added to the menu when right-clicking. This decision should happen when the user right-clicks something, so that the plugin can e.g. do something only if the user has selected some text. Maybe there should be a "Search selected text on Google" option that is grayed out when no text is selected, for example.

This is exactly how the directory tree's right-click menu works. It has a `<<PopulateContextMenu>>` virtual event that is generated when right-clicking, and other plugins bind to that to add various things to the menu. You can search for it (e.g. `git grep "PopulateContextMenu"`) to see how it works.

One challenge is passing the menu object to other plugins so that they can add things to it. I will think about what would be the best way to do this. Remind me in a couple days if I don't reply :)

ArchKats:
Changed it to be closer to your idea.
Now other plugins can add items to the right-click menu using the `add_rightclick_option` or `add_filetab_command`
if the option needs text to be selected.
If text is not selected the option is greyed out.

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,64 @@
...
+            if flag:
+                for i in _ftab_indexes_:
+                    rm.entryconfigure(i, state=tkinter.NORMAL)
+            else:
+                for i in _ftab_indexes_:
+                    rm.entryconfigure(i, state=tkinter.DISABLED)
+        rm.tk_popup(event.x_root, event.y_root)
+        rm.bind("<Unmap>", (lambda event: rm.after_idle(rm.update)), add=True)
+
+def add_rightclick_option(path: str, func):
```
Akuli, line 45: Maybe the same function could also handle menu items that are only for selected text? It could be an optional argument. Like this, for example:
```suggestion
def add_rightclick_option(path: str, func, disable_when_nothing_selected: bool = False):
```
Then we wouldn't need the `add_filetab_command()` function below. To be honest, it's really quite confusing and I would like to get rid of it :D But I like the `add_rightclick_option()` function.

porcupine/plugins/rightclick_menu_options.py
```diff
@@ -0,0 +1,6 @@
+from porcupine.plugins import rightclick_menu, text_google_search
+
+def setup()->None:
+    rightclick_menu.add_filetab_command("Search Google for text", text_google_search.google_search)
```
Akuli, line 4: Put this to the `text_google_search()` plugin. The idea is that each plugin adds its own thing into the right-click menu, similarly to how the menubar at top works.

Akuli (commented):
I like your simple approach :)

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def show_menu(event) -> None:
+    rm = tkinter.Menu(get_main_window(), tearoff=0)
+
+    for path, func in _menu_items_.items():
+        rm.add_command(label=path, command=func)
+
+
+    if rm.index("end") is not None:
```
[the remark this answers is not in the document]
ArchKats: It checks if the menu is empty

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,64 @@
...
+        flag = bool
+        try:
+            flag=text_is_selected(get_filetab())
+        except AssertionError:
```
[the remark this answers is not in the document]
ArchKats: From what I understand `get_filetab()` throws `AssertionError` when a filetab isn't open.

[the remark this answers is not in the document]
Akuli: I don't think it is possible to click something in a right-click menu while you have no tabs opened.

porcupine/plugins/rightclick_menu.py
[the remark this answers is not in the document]
Akuli: I think we don't need to check. A typical Porcupine user will have a non-empty menu, and if someone really has a super weird setup that results in an empty menu, they can easily disable the whole `rightclick_menu` plugin in the plugin manager.

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,61 @@
...
+def add_rightclick_option(path: str, func, needs_selected_text: bool = False):
+    if not needs_selected_text:
+        _menu_items[path] = func
+    else:
+        if func is None:
+            command = lambda: get_filetab().event_generate(f"<<FiletabCommand:{path}>>")
+        else:
+            command = lambda: func(get_filetab())  # type: ignore
+
+        _menu_items[path] = command
```
Akuli, line 49: We really don't need all this:
```suggestion
        _menu_items[path] = func
```

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,61 @@
...
+        _ftab_indexes.append(len(_menu_items) - 1)
+
+
+
+def setup()-> None:
+
+    w = get_main_window()
+    w.bind("<<RightClick>>", show_menu)
```
Akuli, line 57: Ah, I see now why you are doing the weird `except AssertionError` in that other place.

This should be bound only on the text widgets of each file tab, not on the entire Porcupine window. If you click somewhere else than in a file you are editing, you shouldn't get the right-click menu of this PR.

You can use `get_tab_manager().add_filetab_callback()` to run a function on each tab. This function is typically named `on_new_filetab()`, and it should do the binding. See `porcupine/plugins/autoindent.py` for an example.

ArchKats: Thank you, I was trying to find how to do exactly this :D !!!

porcupine/plugins/rightclick_menu.py
```diff
@@ -17,45 +17,33 @@ def text_is_selected(tab: tabs.FileTab) -> bool:
...
     if not needs_selected_text:
         _menu_items[path] = func
     else:
-        if func is None:
-            command = lambda: get_filetab().event_generate(f"<<FiletabCommand:{path}>>")
-        else:
-            command = lambda: func(get_filetab())  # type: ignore
-
-        _menu_items[path] = command
+        _menu_items[path] = lambda: func(get_filetab())
```
Akuli, line 40: I think we should do this in both cases. The `needs_selected_text` shouldn't affect what ends up in `_menu_items`, it should only affect whether we append to `_ftab_indexes`.

porcupine/plugins/rightclick_menu.py
```diff
@@ -17,45 +17,33 @@ def text_is_selected(tab: tabs.FileTab) -> bool:
...
+    if not flag:
+        for i in _ftab_indexes:
+            rm.entryconfigure(i, state=tkinter.DISABLED)
 
     rm.tk_popup(event.x_root, event.y_root)
     rm.bind("<Unmap>", (lambda event: rm.after_idle(rm.destroy)), add=True)
 
-def add_rightclick_option(path: str, func, needs_selected_text: bool = False):
+
+def add_rightclick_option(path: str, func, needs_selected_text: bool = False) -> None:
```
Akuli, line 36: Let's make mypy happy. It is currently complaining that there is no type annotation for the `func` argument:
```suggestion
def add_rightclick_option(path: str, func: Callable[[tabs.FileTab], object], needs_selected_text: bool = False) -> None:
```
You will need to also add `from typing import Callable`.

ArchKats: I have it like that in the case someone wants to add an option that doesn't require the filetab, e.g. someone adding an option for opening a new file.

Akuli: The right-click menu appears when you right-click a text widget of a file tab, so a file tab must exist whenever you are accessing that menu. This means that we always have a filetab that we can pass to the function. If needed, the function can simply ignore the filetab parameter it is getting.

porcupine/plugins/rightclick_menu.py
```diff
@@ -0,0 +1,54 @@
...
+        for i in _ftab_indexes:
+            rm.entryconfigure(i, state=tkinter.DISABLED)
+
+    rm.tk_popup(event.x_root, event.y_root)
+    rm.bind("<Unmap>", (lambda event: rm.after_idle(rm.destroy)), add=True)
+
+
+def add_rightclick_option(
+    path: str,
+    func: Callable[[tabs.FileTab], object] | Callable[[], object],
```
Akuli, line 39: I still don't think there is any need to support functions that don't take a `FileTab`.

Akuli: I edited the code myself to look more like how I imagined it. Let me know what you think.

Akuli:
I just tried this on my computer, and it is pretty good :)

I would like to see more actions in the right-click menu though. I don't want everything in there, just things that people are most likely looking for when they right-click the file. Maybe these are enough for now:

s

porcupine/plugins/rightclick_menu.py
ArchKats: Ok, after thinking about it I now understand how functions that don't need the filetab parameter would 
ignore it.

porcupine/plugins/jump_to_definition.py
```diff
@@ -116,3 +117,6 @@ def generate_jump_request(tab: tabs.FileTab) -> None:
 def setup() -> None:
     get_tab_manager().add_filetab_callback(on_new_filetab)
     menubar.add_filetab_command("Edit/Jump to definition", generate_jump_request)
+    rightclick_menu.add_rightclick_option(
+        "Jump to definition", generate_jump_request, needs_selected_text=True
```
Akuli, line 121: This does not need you to select the entire thing:

```suggestion
        "Jump to definition", generate_jump_request
```

Akuli (commented):
One last nit...

Akuli:
I changed how the "Find and Replace" right-click option works. Adding it every time you create a tab turned out to not be a great idea: if you create two tabs, it would always show the finder in the second tab, even if you right-clicked the first tab. Instead I used the `tab` argument passed to the callback.

Thanks for working on this! I think I will merge this soon :)

ArchKats:
Thanks for all the support and helpful feedback!

PR #1263, opened 2023-03-20
Fix underlines plugin description

### LINKED ISSUE #1172: underlines plugin description is bad
```
[00:56]	Moosems	underlines.py just confused me
[00:57]	Akuli	underlines.py does underlines that show a tooltip when hovered
[00:57]	Akuli	if you just want an underline, it's more than you need
[00:57]	Moosems	oh
```

I think something like "underlines with tooltip" would be much better than the current "underlines in code to indicate different things"

Akuli:
Fixes #1172

### DIFF
--- a/porcupine/plugins/underlines.py
+++ b/porcupine/plugins/underlines.py
@@ -1,7 +1,16 @@
-"""Show underlines in code to indicate different things.
+"""Shows underlines in code that can be hovered to display a tooltip.
 
-Currently the langserver plugin displays errors and warnings with this plugin
-and the urls plugin uses this to create control-clickable links.
+This plugin does nothing by itself, and needs some other plugin to essentially
+tell it what to underline and what the tooltips should say.
+
+For example, if you import something that you don't need in a Python file, the
+langserver plugin will tell this plugin to underline the unused import with a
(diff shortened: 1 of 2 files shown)


porcupine/plugins/underlines.py
```diff
@@ -1,7 +1,16 @@
-"""Show underlines in code to indicate different things.
+"""Shows underlines in code that can be hovered to display a tooltip.
 
-Currently the langserver plugin displays errors and warnings with this plugin
-and the urls plugin uses this to create control-clickable links.
+This plugin does nothing by itself, and needs some other plugin to essentially
+tell it what to underline and what the tooltips should say.
```
Akuli, line 4:
```suggestion
This plugin does nothing by itself, and needs some other plugin to tell it what
to underline and what the tooltips should say.
```

porcupine/plugins/underlines.py
```diff
@@ -1,7 +1,16 @@
...
+
+For example, if you import something that you don't need in a Python file, the
+langserver plugin will tell this plugin to underline the unused import with a
```
Akuli, line 7:
```suggestion
langserver plugin tells this plugin to underline the unused import with a
```

PR #1276, opened 2023-04-15
Forward files list from Porcupine.desktop

rdbende:
This allows files to be opened with `xdg-open`, when Porcupine is set as default app for that MIME type

### DIFF
--- a/porcupine/plugins/desktop_menu.py
+++ b/porcupine/plugins/desktop_menu.py
@@ -41,8 +41,10 @@ def install_desktop_file() -> None:
         file.write("GenericName=Text Editor\n")
         # Must activate the venv, otherwise various things don't work
         # (e.g. os.environ.get("VIRTUAL_ENV") in this plugin)
-        bash_command = f"source {shlex.quote(str(activate_path))} && porcu %F"
-        file.write(f"Exec=bash -c {shlex.quote(bash_command)}\n")
+        # %F is a list of file names that are already quoted, so we cannot place that inside quotes.
+        # Instead we need https://unix.stackexchange.com/a/144519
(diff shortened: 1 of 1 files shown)


Akuli:
This doesn't work when the file name contains a space or a quote, because of how `%F` ends up inside quotes. I'll try to figure out how to fix this.

rdbende:
Indeed. Fortunately, I don't have files with spaces in their names that Porcupine can open.

porcupine/plugins/desktop_menu.py
```diff
@@ -41,7 +41,7 @@ def install_desktop_file() -> None:
         file.write("GenericName=Text Editor\n")
         # Must activate the venv, otherwise various things don't work
         # (e.g. os.environ.get("VIRTUAL_ENV") in this plugin)
-        bash_command = f"source {shlex.quote(str(activate_path))} && porcu"
+        bash_command = f"source {shlex.quote(str(activate_path))} && porcu %F"
         file.write(f"Exec=bash -c {shlex.quote(bash_command)}\n")
```
Akuli, line 45:
```suggestion
        # %F is a list of file names that are already quoted, so we cannot place that inside quotes.
        # Instead we need https://unix.stackexchange.com/a/144519
        bash_command = f'source {shlex.quote(str(activate_path))} && porcu "$@"'
        file.write(f"Exec=bash -c {shlex.quote(bash_command)} bash %F\n")
```

Akuli:
It now works on my system. It opens a file `/home/akuli/Työpöytä/asd "asd" 'asd'.py` successfully.

rdbende:
Works for me as well

PR #1295, opened 2023-05-14
Add sv-ttk plugin

rdbende:
Unfortunately this plugin doesn't really work well if the `ttk_themes` plugin is enabled, because of how  I implemented some stuff in the Sun Valley theme ://


Akuli:
We could just delete the old `ttk_themes` plugin. I think nobody will miss it.

Moosems:
I only ever the others once and it was to look at what they looked like. I prefer clam :).

PR #1301, opened 2023-06-27
Rewrite ask_encoding()

Akuli:
This makes it more similar to `ask_line_ending()`, i.e. there is an obvious list of choices and a short explanation of what each one does. People new to encodings won't need to guess what they should write, or choose randomly from a dropdown with way too many choices.

To see the dialog, click the button in status bar (it usually says `utf-8`).

Old:

o

New:

n


rdbende (approved):
LGTM :))

PR #1302, opened 2023-06-28
Add git blame plugin

### LINKED ISSUE #1289: git blame plugin
Many IDEs show "You (x minutes ago): Blah blah blah" or similar in code, even if you don't ask for this. It's great, because not many people are aware of `git blame`.

Adding this to Porcupine means that there will be 3 ways to do the same thing:
- Run `git blame`
- Press `b` in GitHub when looking at the file
- Look at the line in an editor

But I think it makes sense, each of these ways has their advantages and disadvantages.

rdbende:
Resolves #1289

### DIFF
--- a/porcupine/plugins/git_blame.py
+++ b/porcupine/plugins/git_blame.py
@@ -1,0 +1,91 @@
+from __future__ import annotations
+
+import logging
+import re
+import subprocess
+import sys
+import time
+from datetime import datetime
+from functools import partial
+from pathlib import Path
+from typing import NamedTuple
+
+from porcupine import get_tab_manager, utils
+
+log = logging.getLogger(__name__)
+
+GIT_BLAME_REGEX = r"([0-9a-fA-F]+)\s\((.+?)\s([0-9]*)\s.{5}\s[0-9]+\)"
+
+
+class Commit(NamedTuple):
+    revision: int
+    message: str
+    author: str
+    date: datetime
+
+
+def run_git(*args, cwd: Path) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        ["git"] + list(args),
+        cwd=cwd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        encoding=sys.getfilesystemencoding(),
+        timeout=(60 * 10),  # 10min. Must be huge to avoid unnecessary killing (#885)
+        **utils.subprocess_kwargs,
+    )
+
+
+def is_in_git_repo(path: Path) -> bool:
+    for parent in path.parents:
+        if (parent / ".git").is_dir():
+            return True
+    return False
+
+
+def git_blame_get_commit(path: Path, line: int) -> Commit | None:
+    if not path or not is_in_git_repo(path):
+        # TODO: Do we have to run this every time?
+        return None
+
+    try:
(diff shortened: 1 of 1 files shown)


Akuli:
Maybe the blame functionality should be behind a right-click, so that it doesn't consume CPU power and e.g. make your laptop battery run out quicker.

Moosems:
Most other editors wait for a hover on a line to render or the first line selected/the insert location.

rdbende:
Currently this doesn't do anything, simply prints the info, but yeah, I was wondering when it should actually show it, and where.

Moosems:
I would use the selection/insert and hover (if hover is possible with tkinter Text (maybe a tag_add but that adds more processing time).

Akuli:
> and where.

Found this screenshot of vscode online, IMO it's a pretty good option:

image

It assumes that the editor is wide enough to fit stuff to the side, but `git blame` needs a wide window in general, even on the terminal.

Maybe the best way to show the stuff would be to not actually insert something to the text widget, but add a label with the same font and background color as the text widget.

Akuli:
Or maybe show the commits on the left side, similar to github (press `b` when looking at a file):

b

IMO it would be enough if copying the commit hashes to clipboard was easy, so that you can take the commit hash to terminal for `git show` or `git checkout` or `git revert`, for example.

rdbende:
Inserting it to the text widget screws up the highlighting, so then I'll go with the label.

rdbende:
And what about putting it in the status bar, next to the line and column numbers?

```ruby
Line 12, column 34; changed by rdbende five months ago • this is a descriptive commit message
```

Then we don't have to worry about scrolling, and more text would fit on the screen. Although this way it would be kinda weird to have a timeout before the label appears.

porcupine/plugins/git_blame.py
```diff
@@ -0,0 +1,91 @@
...
+    result = re.search(GIT_BLAME_REGEX, git_blame_result.stdout)
+    if not result:
+        return None
+
+    revision, author, timestamp = result.groups()
+    date = datetime.fromtimestamp(int(timestamp))
+
+    git_log_result = run_git("log", "-n", "1", "--pretty=format:%s", revision, cwd=path.parent)
+
+    return Commit(int(revision), git_log_result.stdout, author, date)
```
rdbende, line 72: TODO: use pydantic here and delete some code above

Moosems:
In VS Code when you click on a hit blade it will show the change on the line in a small popup, that would be cool to have.

porcupine/plugins/git_blame.py
rdbende: No

porcupine/plugins/git_blame.py
```diff
@@ -0,0 +1,133 @@
...
+
+    commit = git_blame_get_commit(path=path, line=line_num)
+    if commit is None:
+        return
+
+    if all(c == "0" for c in commit.hash):
+        formatted = f";  not committed yet"
+    else:
+        formatted = f";  changed by {commit.author} {prettify_date(commit.date)} • {commit.message}"
+    widget.configure(text=formatted)
```
Moosems, line 119: Where does this show? A statusbar button?

Issue #1308, opened 2023-06-29
Documentation is a mess


Akuli:
Porcupine currently has many kinds of documentation:
- akuli.github.io/porcupine, aka Plugin API docs. Some pages are pretty good, e.g. https://akuli.github.io/porcupine/settings.html and https://akuli.github.io/porcupine/plugin-intro.html. A problem is that contributors are usually not familiar with how sphinx works, so these don't get updated much. Also, much of these are autogenerated from docstrings, so pretty pointless if you're going to read the source code anyway.
- Porcupine Wiki. Outdated. I'm pretty convinced nobody reads this.
- CONTRIBUTING.md. Good stuff.
- Large comments in config files (see default_filetypes.toml)

There is currently no documentation about:
- Design philosophy: small core with powerful plugins. What are the advantages and disadvantages of this design and how far we take it.
- File structure: what file should you look at if you want to fix something?
- What each thing in the UI does. I'm pretty satisfied with the picture in mantaray's docs: https://github.com/Akuli/mantaray/blob/3241303/README.md#how-mantaray-works

This is overall a mess. I would suggest:
- Move the good parts of existing documentation to markdown files. Very easy to view and edit for everyone.
- Document more stuff into markdown files
- Keep existing docstrings
- Get rid of akuli.github.io/porcupine

Thoughts? (from everyone, including experienced and new Porcupine developers and)

benjamin-kirkbride:
I would like to see more patterns documented, it's pretty daunting trying to get started.

For instance, I'm looking at the `dedent` method of the `MainText` widget.

There is a lot to figure out here. If there was a "cookbook" of common patterns that would make getting started a lot less daunting.

benjamin-kirkbride:
Example of the sort of thing I found difficult to figure out: https://www.phind.com/search?cache=7a4294b3-9ca4-49cd-a21e-e37350f316d5

rdbende:
> I would like to see more patterns documented, it's pretty daunting trying to get started.

I second that. When I start working on a new plugin, I always just copy an existing one, and delete what I don't need, because I can't remember how to implement all the fun stuff, like `on_new_filetab`, etc.

I rarely look at the docstrings, but rather use the web docs. It's easier for me to find what I'm looking for, but then go usually back to the source code, to read, what it actually does :D

Also:
```
$ git grep "# TODO: document this" | wc -l
20
```

rdbende:
[...]

PR #1314, opened 2023-06-30
Markdown Plugin

### LINKED ISSUE #413: markdown unwanted indent
```
- Oneline strings start and end with `"`. The text between the `"` characters is interpreted
  just like for a multiline string, but with these differences:
    - Writing `\n`, a backslash followed by `n`, is allowed.
```

Try pressing enter after last `.`, it indents but it shouldn't

benjamin-kirkbride:
This is my start of the Markdown Plugin.

Features:
- [X] indenting and dedenting of list items
- [x] pressing enter with a list will continue list at current indentation level
    - [x] support task items
- [ ] pressing enter on an **empty** list item will delete list item

This adds quite a few tests ;)

Fixes #413

### DIFF
--- a/porcupine/plugins/run/windows_run.py
+++ b/porcupine/plugins/run/windows_run.py
@@ -9,26 +9,27 @@ import sys
 
 import colorama
 
-colorama.init()
+if __name__ == "__main__":
+    colorama.init()
 
-prog, directory, command = sys.argv
-print(colorama.Fore.BLUE + command + colorama.Fore.RESET)
-try:
-    returncode: int | None = subprocess.call(command, cwd=directory, shell=True)
-except KeyboardInterrupt:
-    # the subprocess should have already printed any traceback or
-    # whatever it might want to print
-    # TODO: try to catch the return code in this case as well?
-    returncode = None
+    prog, directory, command = sys.argv
+    print(colorama.Fore.BLUE + command + colorama.Fore.RESET)
+    try:
+        returncode: int | None = subprocess.call(command, cwd=directory, shell=True)
+    except KeyboardInterrupt:
+        # the subprocess should have already printed any traceback or
+        # whatever it might want to print
+        # TODO: try to catch the return code in this case as well?
+        returncode = None
 
-print()
-print("-----------------------------")
-if returncode == 0:
-    print("The program completed successfully.")
-elif returncode is None:
-    print("The program was interrupted.")
-else:
-    print(f"The program failed with status {returncode}.")
+    print()
+    print("-----------------------------")
+    if returncode == 0:
+        print("The program completed successfully.")
+    elif returncode is None:
+        print("The program was interrupted.")
+    else:
+        print(f"The program failed with status {returncode}.")
 
-print("Press Enter to close this window...")
-input()
+    print("Press Enter to close this window...")
+    input()
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,10 @@
 [pytest]
 # why we disable nose plugin: https://github.com/pytest-dev/pytest/issues/10825
-addopts = --doctest-modules --capture=no -p no:nose
-testpaths = porcupine/plugins/autoindent.py tests/
+#
+# The debug prompt plugin's docstring contains a ">>>" that isn't a doctest.
+# I don't want a "# doctest: +SKIP" comment because it would show up in the plugin manager.
+addopts = --doctest-modules --capture=no -p no:nose --ignore porcupine/plugins/porcupine_debug_prompt.py
+testpaths = porcupine/ tests/
 markers = pastebin_test
 
 # uncomment this if you dare... i like how pytest hides the shittyness


porcupine/plugins/markdown.py
```diff
@@ -0,0 +1,56 @@
...
+    regex = re.compile(pattern)
+    match = regex.search(line)
+    return bool(match)
+
+
+def on_tab_key(
+    tab: tabs.FileTab, event: tkinter.Event[textutils.MainText], shift_pressed: bool
+) -> str | None:
+    """Indenting and dedenting list items"""
+    if tab.settings.get("filetype_name", str) == "Markdown":
```
Akuli, line 36: Maybe also check what the character just before the cursor is. If it is not a space character, we probably don't want to indent, so that you can autocomplete annoyingly long words even if you are writing that word into a list item. Something like `tab.textwidget.get("insert - 1 char", "insert") == " "`

porcupine/plugins/markdown.py
```diff
@@ -0,0 +1,56 @@
+"""If configuration says so, insert spaces when the tab key is pressed."""
```
Akuli, line 1: Copy/pasta from another plugin. This must be correct/reasonable, because it shows up in the plugin manager.

Akuli (commented):
Nice :)

porcupine/plugins/markdown.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def _is_list_item(line: str) -> bool:
+    """Detect if the line that is passed is a markdown list item
+
+    According to:
+    - https://spec.commonmark.org/0.30/#lists
+    - https://pandoc.org/MANUAL.html#lists
+    """
+    assert len(line.splitlines()) == 1
```
Akuli, line 25: I sometimes get an error from this line:

```
    return self.func(*args)
  File "/home/akuli/porcu/porcupine/utils.py", line 548, in callback
    return on_tab(event, shifted)
  File "/home/akuli/porcu/porcupine/plugins/markdown.py", line 38, in on_tab_key
    list_item_status = _is_list_item(line)
  File "/home/akuli/porcu/porcupine/plugins/markdown.py", line 25, in _is_list_item
    assert len(line.splitlines()) == 1
AssertionError
```

reason:

```
>>> "".splitlines()
[]
```

benjamin-kirkbride: Interesting.. Do you have a way to reproduce that?

porcupine/plugins/markdown.py
benjamin-kirkbride: Makes sense, oversight on my part.

porcupine/plugins/markdown.py
benjamin-kirkbride: Makes sense, will do that

benjamin-kirkbride: I would think that dedenting would make sense even if it's isn't a space char prior, because `Shift-Tab` doesn't trigger the autocomplete?

porcupine/plugins/markdown.py
benjamin-kirkbride: Fixed with 8b3e3df0586385815e483df3cd05a04f76809e65

porcupine/plugins/markdown.py
Akuli: I got it when I tried to do something else, but I guess pressing tab on a blank line does it.

porcupine/plugins/markdown.py
benjamin-kirkbride: Fixed with a1479f6a9e0b3631fa10b946b5a12627a7fb9e71

tests/test_markdown_plugin.py
```diff
@@ -0,0 +1,187 @@
...
+    filetab.textwidget.event_generate("<Tab>")
+    filetab.update()
+    # no change to text, should open autocomplete menu
+    assert filetab.textwidget.get("1.0", "insert") == li
+    filetab.textwidget.event_generate("<Escape>")  # close the autocomplete
+
+    # add a space
+    filetab.textwidget.insert("insert", " ")
+    filetab.textwidget.event_generate("<Tab>")
+    filetab.update()
```
Akuli, line 151: I find that you often need to `.update()` before `event_generate()`, so changing the order might help.

benjamin-kirkbride: Should we consider adding some helper utilities for these sorts of things?

Akuli: I don't think I want to hide the `.update()` + `.event_generate()` combination, since it's just two lines and many people already know what `.update()` does. It should be documented much better though (#1308)

Akuli:
I fixed a couple git things, so you will probably need `git pull` before your next push. I think I will have time to review this tomorrow or the day after tomorrow. You can @ me if I forget :)

benjamin-kirkbride:
Sorry about the mess there, I'm not the best at using git

Issue #1318, opened 2023-07-02
Headless Testing


benjamin-kirkbride:
It would be great to be able to run tests without any windows being created as an option. I have no idea how feasible this might be.

Moosems:
That's.... not really possible. The tests have to create a Tk instance which creates a window. They are decently fast though so I'm not sure why this is a problem.

benjamin-kirkbride:
I figured it out, I'm able to use `xvfb-run` to do it. It seems to break in all the same places that `@pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") == "true", reason="fails CI on all platforms")` appears.

Akuli:
> The tests have to create a Tk instance which creates a window.

Not really true, you can `.withdraw()` the window before it appears (on any platform) or use xvfb (on linux).

> It seems to break in all the same places that @pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") == "true", reason="fails CI on all platforms") appears.

Maybe these should be rewritten to say "skip if running under xvfb" instead of "skip if running in github actions".

benjamin-kirkbride:
> Maybe these should be rewritten to say "skip if running under xvfb" instead of "skip if running in github actions".

Do you know how to detect this? `$DISPLAY==99`?

Akuli:
I don't. I would start with something like `diff -u --color <(env) <(xvfb-run env)`, which compares the environment variables without and with xvfb-run.

Akuli:
Apparently `env` outputs in a random order, so you also need to pipe through `sort`. With that, I get:

```diff
akuli@akuli-desktop:~$ diff -u --color <(env | sort) <(xvfb-run env | sort)
--- /dev/fd/63	2023-07-03 22:46:34.498896512 +0300
+++ /dev/fd/62	2023-07-03 22:46:34.502896511 +0300
@@ -1,7 +1,7 @@
 CLUTTER_BACKEND=x11
 COLORTERM=truecolor
 DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/dbus-2Wvuy9UV3B,guid=daa8b24ed091f58e8466651364a2c193
-DISPLAY=:0.0
+DISPLAY=:99
 GPG_AGENT_INFO=/run/user/1000/gnupg/S.gpg-agent:0:1
 GTK_MODULES=gail:atk-bridge:canberra-gtk-module
 GTK_OVERLAY_SCROLLING=0
[...]

PR #1327, opened 2023-07-03
Testing Improvements

benjamin-kirkbride:
- Python 3.11
- xvfb/headless testing improment

### DIFF
--- a/tests/test_fullscreen_plugin.py
+++ b/tests/test_fullscreen_plugin.py
@@ -5,9 +5,7 @@ import pytest
 from porcupine import get_main_window
 from porcupine.menubar import get_menu
 
-headless = (
-    os.getenv("GITHUB_ACTIONS") == "true" or "xvfb" in os.getenv("XAUTHORITY", "")
-)
+headless = os.getenv("GITHUB_ACTIONS") == "true" or "xvfb" in os.getenv("XAUTHORITY", "")
 
 
 pytestmark = pytest.mark.skipif(headless, reason="Does not work in headless environments")
--- a/tests/test_fullscreen_plugin.py
+++ b/tests/test_fullscreen_plugin.py
@@ -5,15 +5,9 @@ import pytest
 from porcupine import get_main_window
 from porcupine.menubar import get_menu
 
-github_actions = os.getenv("GITHUB_ACTIONS")
-xauthority = os.getenv("XAUTHORITY")
-
-headless = False
-
-if github_actions is not None:
-    headless = github_actions == "true"
-elif xauthority is not None:
-    headless = "xvfb" in xauthority
+headless = (
+    os.getenv("GITHUB_ACTIONS") == "true" or "xvfb" in os.getenv("XAUTHORITY", "")
+)
 
 
 pytestmark = pytest.mark.skipif(headless, reason="Does not work in headless environments")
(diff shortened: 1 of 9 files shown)


benjamin-kirkbride:
Fixes https://github.com/Akuli/porcupine/issues/1318

tests/test_jump_to_definition_plugin.py
```diff
@@ -1,11 +1,19 @@
...
 
+import pytest
 from sansio_lsp_client import ClientState
 
 from porcupine import get_main_window
 from porcupine.plugins.langserver import langservers
 
+pytestmark = pytest.mark.xfail(
+    platform.python_version_tuple()[:2] == ("3", "11"),
+    run=False,
```
Akuli, line 13: - I would prefer `sys.version_info >= (3, 11)`, just how I'm used to seeing version checks written.
- Would be IMO better to run the test anyway, so `strict=True` instead of `run=False`. This way I won't forget to update this when I fix the Python 3.11 situation.

tests/test_fullscreen_plugin.py
```diff
@@ -5,8 +5,20 @@
...
 
+github_actions = os.getenv("GITHUB_ACTIONS")
+xauthority = os.getenv("XAUTHORITY")
+
+headless = False
+
+if github_actions is not None:
+    headless = github_actions == "true"
+elif xauthority is not None:
+    headless = "xvfb" in xauthority
```
Akuli, line 16: Maybe:

```suggestion
headless = (
    os.getenv("GITHUB_ACTIONS") == "true" or "xvfb" in os.getenv("XAUTHORITY", "")
)
```

Akuli (commented):
Thanks! Two nits.

tests/test_jump_to_definition_plugin.py
benjamin-kirkbride: > I would prefer...

Will do, believe it or not I've never actually had to do that before so I just did what seemed to work.

> better to run the test anyway...

Will do, makes sense.

tests/test_fullscreen_plugin.py
benjamin-kirkbride: ...wow how did I not know about the second arg to `get`....

tests/test_jump_to_definition_plugin.py
benjamin-kirkbride: 802e38b fixes

benjamin-kirkbride:
Should adding docs be included in this?

Akuli:
This is fine without docs. We'll need to document the testing setup a lot more later anyway :)

For the last test failure, you can change 8 to 9 in the test. The point is that a thing should take less than 10 seconds, basically happen immediately, and apparently github actions can be so slow that it takes exactly 8 seconds.

benjamin-kirkbride:
Tell me how I change a 8 to a 9 and accidentally commit my venv....

benjamin-kirkbride:
@Akuli done I think

benjamin-kirkbride:
2d9188d was an accident, not sure if I was supposed to do that

Akuli (approved):
Thanks for your work on Porcupine :)

Akuli:
Merging the main branch into your PR branch is fine, but necessary only if there are conflicts.

Issue #1300, opened 2023-06-27
pyls does not work on Python 3.11


Akuli:
Porcupine uses pyls as the default langserver for Python. It does not work at all on Python 3.11.

I really should consider switching to pyright (#560)

rdbende:
Yup, besides `pyls` I can say that everything works fine on 3.11, I've been using it for about half a year when Debian unstable got it.

benjamin-kirkbride:
Do we know which tests are expected to fail as a result of this? Can we conditionally xfail and/or skip them for now?

benjamin-kirkbride:
These tests fail for me with python 3.11, but it's possible there are other factors at play:

> FAILED tests/test_indent_dedent.py::test_markdown_autoindent - AssertionError: assert '1. Lol and\n...\nEnd of list' == '1. Lol and\n...\nEnd of list'
FAILED tests/test_jump_to_definition_plugin.py::test_basic - RuntimeError: timed out waiting
FAILED tests/test_jump_to_definition_plugin.py::test_two_definitions - RuntimeError: timed out waiting
FAILED tests/test_rstrip_plugin.py::test_rstrip - assert 'print("hello")  ' == 'print("hello")\n'
FAILED tests/test_tabs.py::test_read_only_file - ValueError: too many values to unpack (expected 1)
ERROR tests/test_jump_to_definition_plugin.py::test_basic - ValueError: error in tkinter callback while running test
ERROR tests/test_jump_to_definition_plugin.py::test_two_definitions - ValueError: error in tkinter callback while running test

Akuli:
I think only the `test_jump_to_definition_plugin.py` failures are caused by this. If you xfail them, it would be nice if you made a PR about it, and also add 3.11 to version lists in `.github/workflows/` so that CI tests it with the xfails :)

benjamin-kirkbride:
Upstream issue: https://github.com/palantir/python-language-server/issues/959

benjamin-kirkbride:
I think we should consider switching to https://github.com/python-lsp/python-lsp-server

Pyright is not really a replacement for pyls; it is just a static type checker (Mypy replacement). The VSCode/Microsoft LSP that does auto-completion, refactoring, etc is `pylance` which is not open source (or usable outside of VSCode at all).

Akuli:
I am aware that pyright and pylance are not the same, but pyright also has a langserver that gives autocompletions and such. It is usable without pylance, and probably enough for Porcupine without pylance.

[...]

PR #1328, opened 2023-07-03
Markdown

benjamin-kirkbride:
Accidentally closed https://github.com/Akuli/porcupine/pull/1314 :sweat_smile:

### DIFF
--- a/porcupine/plugins/markdown.py
+++ b/porcupine/plugins/markdown.py
@@ -30,9 +30,7 @@ def _is_list_item(line: str) -> bool:
 
 
 def on_tab_key(
-    tab: tabs.FileTab,
-    event: tkinter.Event[textutils.MainText],
-    shift_pressed: bool,
+    tab: tabs.FileTab, event: tkinter.Event[textutils.MainText], shift_pressed: bool
 ) -> str | None:
     """Indenting and dedenting list items"""
     if tab.settings.get("filetype_name", str) == "Markdown":
--- a/porcupine/plugins/markdown.py
+++ b/porcupine/plugins/markdown.py
@@ -40,11 +40,16 @@ def on_tab_key(
         line = event.widget.get("insert linestart", "insert lineend")
         list_item_status = _is_list_item(line)
 
+        # shift-tab
         if shift_pressed and list_item_status:
             event.widget.dedent("insert linestart")
             return "break"
 
-        if list_item_status:
+        # if it isn't, we want tab to trigger autocomplete instead
+        char_before_cursor_is_space = tab.textwidget.get("insert - 1 char", "insert") == " "
+
+        # tab
+        if list_item_status and char_before_cursor_is_space:
             event.widget.indent("insert linestart")
             return "break"
 
--- a/porcupine/plugins/markdown.py
+++ b/porcupine/plugins/markdown.py
@@ -30,8 +30,10 @@ def _is_list_item(line: str) -> bool:
 
 
 def on_tab_key(
-    tab: tabs.FileTab, event: tkinter.Event[textutils.MainText], shift_pressed: bool
-) -> str:
+    tab: tabs.FileTab,
+    event: tkinter.Event[textutils.MainText],
+    shift_pressed: bool,
+) -> str | None:
     """Indenting and dedenting list items"""
     if tab.settings.get("filetype_name", str) == "Markdown":
         line = event.widget.get("insert linestart", "insert lineend")
@@ -44,6 +46,8 @@ def on_tab_key(
         if list_item_status:
             event.widget.indent("insert linestart")
             return "break"
+
+    return None
 
 
 def on_new_filetab(tab: tabs.FileTab) -> None:
--- a/porcupine/plugins/markdown.py
+++ b/porcupine/plugins/markdown.py
@@ -30,9 +30,7 @@ def _is_list_item(line: str) -> bool:
 
 
 def on_tab_key(
-    tab: tabs.FileTab,
-    event: tkinter.Event[textutils.MainText],
-    shift_pressed: bool,
+    tab: tabs.FileTab, event: tkinter.Event[textutils.MainText], shift_pressed: bool
 ) -> str:
     """Indenting and dedenting list items"""
     if tab.settings.get("filetype_name", str) == "Markdown":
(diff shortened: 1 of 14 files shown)


Akuli:
As for git messes: it's totally ok to make a mess with git! I use "squash and merge" in my projects, which basically means that a messy PR will appear as just one tidy commit when someone looks at the Git history.

You might also like https://akuli.github.io/git-guide/

porcupine/plugins/markdown.py
```diff
@@ -0,0 +1,109 @@
...
+import re
+import tkinter
+from functools import partial
+
+from porcupine import get_tab_manager, tabs, textutils, utils
+
+log = logging.getLogger(__name__)
+
+
+setup_before = ["tabs2spaces", "autoindent"]
```
Akuli, line 18: Maybe must be also be before autocomplete, so that `return "break"` prevents autocompletions?

You can try running Porcupine with `--shuffle-plugins` a few times to see if you have defined all necessary dependencies. (See `--help`)

tests/test_indent_dedent.py
```diff
@@ -198,28 +198,6 @@ def check(filename, input_commands, output):
     return check
 
 
-def test_markdown_autoindent(check_autoindents):
-    check_autoindents(
```
Akuli, line 202: Can we keep this, but just modify it so that it passes?

tests/test_markdown_plugin.py
```diff
@@ -0,0 +1,257 @@
...
+    ListItemCase(id="duplicate token 5", line="  -#) item -- 1 -", expected=False),
+    ListItemCase(id="duplicate token 6", line="  *-#)1. item -- 1 -", expected=False),
+]
+
+# test `#` and 0 to 99 numbered lists
+# tests ol with `.` and `)`
+IS_LIST_ITEM_CASES.extend(
+    [
+        ListItemCase(id=f"numbered {i}", line=f"{i}{sep} item 1", expected=True)
+        for i, sep in itertools.product(itertools.chain(range(100), "#"), (".", ")"))
```
Akuli, line 46: These are maybe a bit too thorough :) We really don't need more than a thousand green dots when running pytest because of one regex. But more importantly, these two lines aren't very readable. It took me a while to figure out what they do.

Maybe instead add a few specific cases, so that it's easier to understand?

```
ListItemCase(id=f"numbered 1", line=f"1. item 1", expected=True)
ListItemCase(id=f"numbered 1", line=f"1) item 1", expected=True)
ListItemCase(id=f"numbered 4", line=f"4. item 1", expected=True)
ListItemCase(id=f"numbered 4", line=f"4) item 1", expected=True)
ListItemCase(id=f"numbered 42", line=f"42. item 1", expected=True)
ListItemCase(id=f"numbered 42", line=f"42) item 1", expected=True)
ListItemCase(id=f"numbered #", line=f"#. item 1", expected=True)
ListItemCase(id=f"numbered #", line=f"#) item 1", expected=True)
```

Also applies to other `.extend()`s.

Akuli (commented):
I like it overall :)

benjamin-kirkbride:
@Akuli does the error in CI mean anything to you?

porcupine/plugins/markdown.py
benjamin-kirkbride: The tests run with `--shuffle-plugins`, so shouldn't that be covered?

This actually doesn't interfere with autocomplete, the regex doesn't match useless there is a space after the list prefix, so regardless we should be good I believe. If you think that's wrong, or can think of a test to verify that, I'm all ears :)

tests/test_indent_dedent.py
benjamin-kirkbride: I got rid of it because it felt redundant with all the markdown tests, but I can look back and see if there is a way to make that happen

tests/test_markdown_plugin.py
benjamin-kirkbride: > We really don't need more than a thousand green dots when running pytest because of one regex

But it's so fun :weary: 

> these two lines aren't very readable

That's fair. It felt readable at the time :sweat: 

> Maybe instead add a few specific cases, so that it's easier to understand?

Yeah I can do this.

On a more broad note, how would you feel about adding https://hypothesis.readthedocs.io/en/latest/ to the test suite?

Akuli:
The CI error has to do with #1300. Apparently the CI on Python 3.11 doesn't pass very reliably, even though you got it to pass most of the time in #1327.

I'd suggest we do nothing about the CI error if it disappears when tests run again. We will "soon" switch to pyright anyway.

tests/test_markdown_plugin.py
Akuli: I'm not a big fan of `hypothesis`. IMO tests should be as concrete/specific as possible, such as `ListItemCase(id="numbered 42", line="42) item 1", expected=True)`, whereas the code being tested tends to be written very generally. This ensures that at least the test is correct, because it's ideally very simple.

And yes, I know that the point of `hypothesis` is to catch bugs where a specific string that you didn't think of causes a problem. But they tend to get caught without unit tests in Porcupine. For example, the empty `line.splitlines()` bug caused errors when I opened a file.

I can see `hypothesis` being useful in other projects though. For example, maybe you have functions to escape and unescape a string in some way, and you want to know that they work in all corner cases.

benjamin-kirkbride: I would see hypothesis as purely additive, it wouldn't replace the simple `ListItemCase(id="numbered 42", line="42) item 1", expected=True)`, sanity type tests, but it would replace:

```python
IS_LIST_ITEM_CASES.extend(
    [
        ListItemCase(id=f"numbered {i}", line=f"{i}{sep} item 1", expected=True)
        for i, sep in itertools.product(itertools.chain(range(100), "#"), (".", ")"))
```

> For example, the empty line.splitlines() bug caused errors when I opened a file.

I think there is likely many more of these sorts of things, especially as more and more plugins get added ;)

I think hypothesis would likely help a lot with the flaky tests as well. I should probably move this discussion to an issue.

benjamin-kirkbride:
This PR is blocked by https://github.com/Akuli/porcupine/pull/1349

benjamin-kirkbride:
I am now going to work to finish this using Actions

Issue #1332, opened 2023-07-04
Markdown Editor Controls


benjamin-kirkbride:
![image](https://github.com/Akuli/porcupine/assets/3373830/8a54abda-1fc0-4516-890e-b8172f9cf34a)

When editing a Markdown file, I want some buttons on a bar to appear above the editor window to help with common Markdown features.

I am not aware of an existing pattern for implementing such a thing. Probably need a separate issue for creating a toggleable "button bar".

Hopefully this isn't antithetical to what people are wanting for Porcupine, but this isn't any more invasive than the side-by-side thing that @rdbende is working on, I think

benjamin-kirkbride:
So I'm not sure this makes sense to implement as a plugin. The reason for that is because I would want other plugins to be able to bind "commands" to this (in the form of buttons). Similar to how `menubar.add_filetab_command` works now. This is also pertinent to https://github.com/Akuli/porcupine/issues/1319

I'm going to pursue this further and see how it goes. Input appreciated.

Akuli:
I don't really want buttons like this. A few reasons:
- We don't have a good way to deal with the many flavors of markdown. For example, should triple-backticks or indentation be used for code blocks? The preview plugin can just support both, but we have to pick one if we have a code snippet button. The button would be useless when writing for something that only supports the other kind of syntax.
- The entire purpose of markdown compared to other similar languages is that the syntax is easy to write and remember. For this reason I think most people wouldn't use the buttons.
- This would be a lot of work to implement (many buttons, each with an image) for little benefit.

I hope this makes sense. I find it's best to tell people early when I'm not interested, so that they don't work on something for a few days only to find out that I don't want to merge.

Akuli:
Related: #159

benjamin-kirkbride:
The way I have conceived of implementing this I think actually addresses all of your concerns. Also I'm intending to implement this as a plugin, so it can be turned off without problem :)

> We don't have a good way to deal with the many flavors of markdown

https://github.com/Akuli/porcupine/issues/1322 ;)

> should triple-backticks or indentation be used for code blocks

There is a lot of prior art here.

Github:
![image](https://github.com/Akuli/porcupine/assets/3373830/4072aaa1-cb29-4a89-9bb3-50917e12fe28)

Joplin:
![image](https://github.com/Akuli/porcupine/assets/3373830/d6716103-10d7-4309-9b50-fb9adb2dc54d)

StackEdit:
[...]

Issue #1319, opened 2023-07-02
New Feature: Command Palette 


benjamin-kirkbride:
In many modern editors, particularly VSCode, there is a "Command Palette". In VSCode it is accessed using `CTRL-SHIFT-P`:

![image](https://github.com/Akuli/porcupine/assets/3373830/9c812b84-73fa-4004-9d4a-b9d0a0601985)

The idea is that basically all "actions" provide hooks that can be activated by the Command Pallette.

It also doubles as a way to jump to various points.

<details><summary>Jump to File</summary>
<p>
![image](https://github.com/Akuli/porcupine/assets/3373830/3b69396a-1d6b-4561-95d2-01d852fc178e)
</p>
</details> 

<details><summary>Jump to Symbol</summary>
<p>
![image](https://github.com/Akuli/porcupine/assets/3373830/520420db-f6ea-4a19-a61b-c7299ba8edfc)
</p>
</details> 

And more.

Issue #1026, opened 2022-01-16
right-click menu in the main text widget


Akuli:
Porcupine has a few things that are applied in the main widget to selected text or at the cursor location:
- Edit --> Comment block
- Edit --> Fold
- Edit --> Jump to definition
- Edit --> Sort lines
- Edit --> Anchors --> Add or remove on this line
- Tools --> Search selected text on google

It would be nice if all these were also accessible with a simple right-click, instead of having to browse through the menus at top. This should be implemented in a way where plugins can add their own items to the right-click menu easily.

benjamin-kirkbride:
https://github.com/Akuli/porcupine/issues/1319 helps with this, as it makes it so the options aren't either "it's in the right click menu, or you have to hunt through the top menu

Akuli:
Also cut/copy/paste because sometimes I'm feeling really lazy and I already have mouse in hand

PR #1334, opened 2023-07-05
Feat: Toolbar

benjamin-kirkbride:
Solves https://github.com/Akuli/porcupine/issues/1332

This is a first pass, looking for feedback.

Please don't get caught up on the current buttons being for `black` and `isort`, I am sure you feel that having a dedicated toolbar for these are not of much value, and I agree, but they were the easiest to start with as an example.

I am seeking feedback on how I implemented the toolbar, and how it is interacted with from other plugins. Ideally we come up wiht a pattern for "registering" "commands" that can be selectively shared between the toolbar, [Command Palette](https://github.com/Akuli/porcupine/issues/1319), [right click menu](https://github.com/Akuli/porcupine/issues/1026), and probably the main menu as well?

Features I have thought of, and their status:
- [ ] better theming/appearance of the toolbar
    - [ ] button colors
    - [ ] button symbols/images
        - [ ] work through the sourcing/svg problem
    - [ ] toolbar separators
    - [ ] toolbar padding
    - [ ] toolbar outline
- [x] way to control ordering of buttons (needs some more thought maybe)
- [x] toolbar only appearing for filetypes that it has buttons registered for
- [ ] user able to disable toolbar on a per-filetype basis
- [x] implement as a plugin such that it (I believe) can be completely disabled

PS I may have gone a bit overboard with the custom classes :sweat_smile:

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -19,7 +19,7 @@ jobs:
         path: ./pr
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.11"
         cache: pip
     - run: pip install wheel
     - run: pip install -r requirements-dev.txt
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,7 +12,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.11"
         cache: pip
     - run: pip install wheel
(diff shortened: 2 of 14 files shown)


benjamin-kirkbride:
Also, I rebased all my other branches onto this one, that will need worked through at some point but till then focus on 8775d89 - that is where the changes are.

benjamin-kirkbride:
One thing in particular I would like feedback on is how to test this. The buttons are being generated dynamically, so I'm not sure how to "emulate" clicking them to verify their functionality.

PR #1349, opened 2023-07-07
Feat: Actions

### LINKED ISSUE #1341: Actions - Initial Implementation
Subtask of https://github.com/Akuli/porcupine/issues/1342

Reqs from https://github.com/Akuli/porcupine/issues/1335#issuecomment-1624329063

Requirements:
- [x] `actions.py`: the action classes, the registry and the related functions
- [x] A few producers (i.e. add an action in a few plugins)
- [x] One consumer

benjamin-kirkbride:
First pass, here is how you would register an action that is only available for Markdown files against Paths:
```python
register_path_action(
    name="foo",
    description="bar",
    callback=lambda: None,
    availability_callbacks=[filetype_availability(["Markdown"])],
)
```

I was going to have some dedicated arguments for things like filetype availability, whether or not you can use a directory or a file path, etc, but ultimately I think what makes the most sense is having a single callback that takes a list of `availability_callbacks`, idea being these callbacks will be executed in order, and if any return `False`, the action is not available. I think that creating helpers for this will be extremely flexible, composable, and dev friendly.

Thoughts on how it is so far?

Fixes #1341

### DIFF
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -1,0 +1,140 @@
+from dataclasses import dataclass
+from functools import partial
+from pathlib import Path
+from typing import Callable
+
+from porcupine.tabs import FileTab
+
+action_availability_callback = Callable[..., bool]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class Action:
+    """Action that requires no context in the callback"""
+
+    name: str
+    description: str
+    callback: Callable[..., None]
+    availability_callbacks: list[action_availability_callback] | None = None
+
+
+filetab_action_availability_callback = Callable[[FileTab], bool]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class FileTabAction:
+    """Action that requires a FileTab to be provided to the callback"""
+
+    name: str
+    description: str
+    callback: Callable[[FileTab], None]
+    availability_callbacks: list[filetab_action_availability_callback] | None = None
+
+
+path_action_availability_callback = Callable[[Path], bool]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class PathAction:
+    """Action that requires a Path to be provided to the callback"""
+
+    name: str
+    description: str
+    callback: Callable[[Path], None]
+    file_compatible: bool = True
+    directory_compatible: bool = False
+    availability_callbacks: list[path_action_availability_callback] | None = None
+
+
+ActionTypes = Action | FileTabAction | PathAction
+
+_actions: dict[str, ActionTypes] = {}
+
+
+def register_action(
+    *,
+    name: str,
+    description: str,
+    callback: Callable[..., None],
+    availability_callbacks: list[action_availability_callback] | None = None,
+) -> Action:
+    if name in _actions:
+        raise ValueError(f"Action with the name '{name}' already exists")
+    action = Action(
+        name=name,
+        description=description,
+        callback=callback,
+        availability_callbacks=availability_callbacks,
+    )
+    _actions[name] = action
+    return action
+
+
+def register_filetab_action(
+    *,
+    name: str,
+    description: str,
+    callback: Callable[[FileTab], None],
+    availability_callbacks: list[filetab_action_availability_callback] | None,
+) -> FileTabAction:
+    if name in _actions:
+        raise ValueError(f"Action with the name '{name}' already exists")
+    action = FileTabAction(
+        name=name,
+        description=description,
+        callback=callback,
+        availability_callbacks=availability_callbacks,
+    )
+    _actions[name] = action
+    return action
+
+
+def register_path_action(
+    *,
+    name: str,
+    description: str,
+    callback: Callable[[Path], None],
+    file_compatible: bool = True,
+    directory_compatible: bool = False,
+    availability_callbacks: list[path_action_availability_callback] | None,
+) -> PathAction:
+    if name in _actions:
+        raise ValueError(f"Action with the name '{name}' already exists")
+    action = PathAction(
+        name=name,
+        description=description,
+        callback=callback,
+        file_compatible=file_compatible,
+        directory_compatible=directory_compatible,
+        availability_callbacks=availability_callbacks,
+    )
+    _actions[name] = action
+    return action
+
+
+def filetype_availability(filetypes: list[str]) -> Callable[[FileTab | Path], bool]:
+    def _filetype_availability(filetypes: list[str], context: FileTab | Path) -> bool:
+        if isinstance(context, FileTab):
+            tab = context
+            if tab.settings.get("filetype_name", object) in filetypes:
+                return True
+            return False
+
+        if isinstance(context, Path):
+            path = context
+
+            if not path.exists():
+                raise RuntimeError(f"{path} does not exist.")
+            if path.is_dir():
+                raise RuntimeError(
+                    f"{path} is a directory - an action consumer registered this action incorrectly"
+                )
+            if not path.is_file():
+                raise RuntimeError(f"{path} is not a file")
+
+            # return True if get_filetype_from_path(path) in filetypes else False
+            raise NotImplementedError  # TODO: there is a way to do this already right?
+
+        raise RuntimeError("wrong context passed")
+
+    return partial(_filetype_availability, filetypes)
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -50,7 +50,7 @@ Action = Union[BareAction, FileTabAction, PathAction]
 _actions: dict[str, Action] = {}
 
 
-def register_action(
+def register_bare_action(
     *,
     name: str,
     description: str,
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -8,7 +8,7 @@ from porcupine.tabs import FileTab
 action_availability_callback = Callable[[], bool]
 
 
-@dataclass(slots=True, frozen=True)
+@dataclass(frozen=True)
 class Action:
     """Action that requires no context in the callback"""
 
@@ -21,7 +21,7 @@ class Action:
 filetab_action_availability_callback = Callable[[FileTab], bool]
 
 
-@dataclass(slots=True, frozen=True)
+@dataclass(frozen=True)
 class FileTabAction:
     """Action that requires a FileTab to be provided to the callback"""
 
@@ -34,7 +34,7 @@ class FileTabAction:
 path_action_availability_callback = Callable[[Path], bool]
 
 
-@dataclass(slots=True, frozen=True)
+@dataclass(frozen=True)
 class PathAction:
     """Action that requires a Path to be provided to the callback"""
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -8,7 +8,7 @@ from porcupine.tabs import FileTab
 action_availability_callback = Callable[[], bool]
 
 
-@dataclass(kw_only=True, slots=True, frozen=True)
+@dataclass(slots=True, frozen=True)
 class Action:
     """Action that requires no context in the callback"""
 
@@ -21,7 +21,7 @@ class Action:
 filetab_action_availability_callback = Callable[[FileTab], bool]
 
 
-@dataclass(kw_only=True, slots=True, frozen=True)
+@dataclass(slots=True, frozen=True)
 class FileTabAction:
     """Action that requires a FileTab to be provided to the callback"""
 
@@ -34,7 +34,7 @@ class FileTabAction:
 path_action_availability_callback = Callable[[Path], bool]
 
 
-@dataclass(kw_only=True, slots=True, frozen=True)
+@dataclass(slots=True, frozen=True)
 class PathAction:
     """Action that requires a Path to be provided to the callback"""
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -108,29 +108,9 @@ def register_path_action(
     return action
 
 
-def filetype_availability(filetypes: list[str]) -> Callable[[FileTab | Path], bool]:
-    def _filetype_availability(filetypes: list[str], context: FileTab | Path) -> bool:
-        if isinstance(context, FileTab):
-            tab = context
-            if tab.settings.get("filetype_name", object) in filetypes:
-                return True
-            return False
+def query_actions(name: str) -> Action | None:
+    return _actions.get(name)
 
-        if isinstance(context, Path):
-            path = context
 
-            if not path.exists():
-                raise RuntimeError(f"{path} does not exist.")
-            if path.is_dir():
-                raise RuntimeError(
-                    f"{path} is a directory - an action consumer registered this action incorrectly"
-                )
-            if not path.is_file():
-                raise RuntimeError(f"{path} is not a file")
-
-            # return True if get_filetype_from_path(path) in filetypes else False
-            raise NotImplementedError  # TODO: there is a way to do this already right?
-
-        raise RuntimeError("wrong context passed")
-
-    return partial(_filetype_availability, filetypes)
+def get_all_actions() -> dict[str, Action]:
+    return _actions.copy()
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 from dataclasses import dataclass
 from functools import partial
 from pathlib import Path
-from typing import Callable
+from typing import Callable, Union
 
 from porcupine.tabs import FileTab
 
@@ -46,7 +48,7 @@ class PathAction:
     availability_callbacks: list[path_action_availability_callback] | None = None
 
 
-ActionTypes = Action | FileTabAction | PathAction
+ActionTypes = Union[Action, FileTabAction, PathAction]
 
 _actions: dict[str, ActionTypes] = {}
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -16,7 +16,7 @@ class Action:
 
     name: str
     description: str
-    callback: Callable[..., None]
+    callback: Callable[[], None]
     availability_callbacks: list[action_availability_callback] | None = None
 
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -70,7 +70,7 @@ def register_filetab_action(
     availability_callback: Callable[[FileTab], bool] = lambda tab: True,
 ) -> FileTabAction:
     if name in _actions:
-        raise ValueError(f"Action with the name '{name}' already exists")
+        raise ValueError(f"Action with the name {name!r} already exists")
     action = FileTabAction(
         name=name,
         description=description,
@@ -89,7 +89,7 @@ def register_path_action(
     availability_callback: Callable[[Path], bool] = lambda path: True,
 ) -> PathAction:
     if name in _actions:
-        raise ValueError(f"Action with the name '{name}' already exists")
+        raise ValueError(f"Action with the name {name!r} already exists")
     action = PathAction(
         name=name,
         description=description,
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -11,13 +11,13 @@ action_availability_callback = Callable[[], bool]
 
 
 @dataclass(frozen=True)
-class Action:
+class BareAction:
     """Action that requires no context in the callback"""
 
     name: str
     description: str
     callback: Callable[[], None]
-    availability_callbacks: list[action_availability_callback] | None = None
+    availability_callback: action_availability_callback
 
 
 filetab_action_availability_callback = Callable[[FileTab], bool]
@@ -30,7 +30,7 @@ class FileTabAction:
     name: str
     description: str
     callback: Callable[[FileTab], None]
-    availability_callbacks: list[filetab_action_availability_callback] | None = None
+    availability_callback: filetab_action_availability_callback
 
 
 path_action_availability_callback = Callable[[Path], bool]
@@ -43,14 +43,12 @@ class PathAction:
     name: str
     description: str
     callback: Callable[[Path], None]
-    file_compatible: bool = True
-    directory_compatible: bool = False
-    availability_callbacks: list[path_action_availability_callback] | None = None
+    availability_callback: path_action_availability_callback
 
 
-ActionTypes = Union[Action, FileTabAction, PathAction]
+Action = Union[BareAction, FileTabAction, PathAction]
 
-_actions: dict[str, ActionTypes] = {}
+_actions: dict[str, Action] = {}
 
 
 def register_action(
@@ -58,15 +56,15 @@ def register_action(
     name: str,
     description: str,
     callback: Callable[..., None],
-    availability_callbacks: list[action_availability_callback] | None = None,
-) -> Action:
+    availability_callback: action_availability_callback = lambda: True,
+) -> BareAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
-    action = Action(
+    action = BareAction(
         name=name,
         description=description,
         callback=callback,
-        availability_callbacks=availability_callbacks,
+        availability_callback=availability_callback,
     )
     _actions[name] = action
     return action
@@ -77,7 +75,7 @@ def register_filetab_action(
     name: str,
     description: str,
     callback: Callable[[FileTab], None],
-    availability_callbacks: list[filetab_action_availability_callback] | None,
+    availability_callback: filetab_action_availability_callback = lambda tab: True,
 ) -> FileTabAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
@@ -85,7 +83,7 @@ def register_filetab_action(
         name=name,
         description=description,
         callback=callback,
-        availability_callbacks=availability_callbacks,
+        availability_callback=availability_callback,
     )
     _actions[name] = action
     return action
@@ -96,9 +94,7 @@ def register_path_action(
     name: str,
     description: str,
     callback: Callable[[Path], None],
-    file_compatible: bool = True,
-    directory_compatible: bool = False,
-    availability_callbacks: list[path_action_availability_callback] | None,
+    availability_callback: path_action_availability_callback = lambda path: True,
 ) -> PathAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
@@ -106,9 +102,7 @@ def register_path_action(
         name=name,
         description=description,
         callback=callback,
-        file_compatible=file_compatible,
-        directory_compatible=directory_compatible,
-        availability_callbacks=availability_callbacks,
+        availability_callback=availability_callback,
     )
     _actions[name] = action
     return action
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -111,7 +111,7 @@ def get_all_actions() -> dict[str, Action]:
 # Availability Helpers
 
 
-def filetype_is(filetypes: Union[list[str], str]) -> Callable[[FileTab], bool]:
+def filetype_is(filetypes: str | list[str]) -> Callable[[FileTab], bool]:
     def _filetype_is(filetypes: list[str], tab: FileTab) -> bool:
         try:
             filetype = tab.settings.get("filetype_name", object)
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from functools import partial
 from pathlib import Path
 from typing import Callable, Union
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -100,7 +100,7 @@ def register_path_action(
     return action
 
 
-def query_actions(name: str) -> Action | None:
+def get_action(name: str) -> Action | None:
     return _actions.get(name)
 
 
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -6,8 +6,6 @@ from typing import Callable, Union
 
 from porcupine.tabs import FileTab
 
-action_availability_callback = Callable[[], bool]
-
 
 @dataclass(frozen=True)
 class BareAction:
@@ -16,10 +14,7 @@ class BareAction:
     name: str
     description: str
     callback: Callable[[], None]
-    availability_callback: action_availability_callback
-
-
-filetab_action_availability_callback = Callable[[FileTab], bool]
+    availability_callback: Callable[[], bool]
 
 
 @dataclass(frozen=True)
@@ -29,10 +24,7 @@ class FileTabAction:
     name: str
     description: str
     callback: Callable[[FileTab], None]
-    availability_callback: filetab_action_availability_callback
-
-
-path_action_availability_callback = Callable[[Path], bool]
+    availability_callback: Callable[[FileTab], bool]
 
 
 @dataclass(frozen=True)
@@ -42,7 +34,7 @@ class PathAction:
     name: str
     description: str
     callback: Callable[[Path], None]
-    availability_callback: path_action_availability_callback
+    availability_callback: Callable[[Path], bool]
 
 
 Action = Union[BareAction, FileTabAction, PathAction]
@@ -55,7 +47,7 @@ def register_bare_action(
     name: str,
     description: str,
     callback: Callable[..., None],
-    availability_callback: action_availability_callback = lambda: True,
+    availability_callback: Callable[[], bool] = lambda: True,
 ) -> BareAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
@@ -74,7 +66,7 @@ def register_filetab_action(
     name: str,
     description: str,
     callback: Callable[[FileTab], None],
-    availability_callback: filetab_action_availability_callback = lambda tab: True,
+    availability_callback: Callable[[FileTab], bool] = lambda tab: True,
 ) -> FileTabAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
@@ -93,7 +85,7 @@ def register_path_action(
     name: str,
     description: str,
     callback: Callable[[Path], None],
-    availability_callback: path_action_availability_callback = lambda path: True,
+    availability_callback: Callable[[Path], bool] = lambda path: True,
 ) -> PathAction:
     if name in _actions:
         raise ValueError(f"Action with the name '{name}' already exists")
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from functools import partial
 from pathlib import Path
 from typing import Callable, Union
 
@@ -105,3 +106,22 @@ def query_actions(name: str) -> Action | None:
 
 def get_all_actions() -> dict[str, Action]:
     return _actions.copy()
+
+
+# Availability Helpers
+
+
+def filetype_is(filetypes: Union[list[str], str]) -> Callable[[FileTab], bool]:
+    def _filetype_is(filetypes: list[str], tab: FileTab) -> bool:
+        try:
+            filetype = tab.settings.get("filetype_name", object)
+        except KeyError:
+            # don't ask me why a `get` method can raise a KeyError :p
+            return False
+
+        return filetype in filetypes
+
+    if isinstance(filetypes, str):
+        filetypes = [filetypes]
+
+    return partial(_filetype_is, filetypes)
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -5,7 +5,7 @@ from typing import Callable
 
 from porcupine.tabs import FileTab
 
-action_availability_callback = Callable[..., bool]
+action_availability_callback = Callable[[], bool]
 
 
 @dataclass(kw_only=True, slots=True, frozen=True)
--- a/porcupine/actions.py
+++ b/porcupine/actions.py
@@ -51,7 +51,7 @@ def register_bare_action(
     availability_callback: Callable[[], bool] = lambda: True,
 ) -> BareAction:
     if name in _actions:
-        raise ValueError(f"Action with the name '{name}' already exists")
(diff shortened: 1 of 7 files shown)


benjamin-kirkbride:
Solves https://github.com/Akuli/porcupine/issues/1341

benjamin-kirkbride:
Instead of having 3 `register_*action` commands, I could have boolean switches on a single `register_action` function. What are your thoughts?

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+action_availability_callback = Callable[[], bool]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class Action:
+    """Action that requires no context in the callback"""
+
+    name: str
+    description: str
+    callback: Callable[..., None]
```
Akuli, line 17: Maybe:
```suggestion
    callback: Callable[[], None]
```

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+    availability_callbacks: list[action_availability_callback] | None = None
```
Akuli, line 18: I would prefer just `Callable[[], bool]`, not a list. There's also no need for `None` because you can use `lambda: True`. This is how `porcupine/plugins/filemanager.py` works now. It basically defines a bunch of `PathAction`s.

```python
commands = [
    Command(
        "New directory here", "<<FileManager:New directory>>", (lambda p: True), new_directory_here
    ),
    Command("Cut", "<<Cut>>", (lambda p: not p.is_dir()), cut),
    Command("Copy", "<<Copy>>", (lambda p: not p.is_dir()), copy),
    Command("Paste", "<<Paste>>", can_paste, paste),
    Command("Rename", "<<FileManager:Rename>>", is_NOT_project_root, rename),
    Command(f"Move to {trash_name}", "<<FileManager:Trash>>", is_NOT_project_root, trash),
    Command("Delete", "<<FileManager:Delete>>", is_NOT_project_root, delete),
    Command("Open in file manager", None, (lambda p: p.is_dir()), open_in_file_manager),
    Command("Open in terminal", None, (lambda p: p.is_dir()), open_in_terminal),
    Command("Copy full path to clipboard", None, (lambda p: True), copy_full_path_to_clipboard),
]
```

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+from pathlib import Path
+from typing import Callable
+
+from porcupine.tabs import FileTab
+
+action_availability_callback = Callable[[], bool]
+
+
+@dataclass(kw_only=True, slots=True, frozen=True)
+class Action:
```
Akuli, line 12: Might be nice to name this something else than `Action`. It will likely need to be special-cased somewhere when looping over all actions, and `for action in get_actions()` + `if isinstance(action, Action)` would look just confusing.

But I'm having trouble coming up with a good name. `SimpleAction`? `NoInputsAction`?

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+
+    name: str
+    description: str
+    callback: Callable[[Path], None]
+    file_compatible: bool = True
+    directory_compatible: bool = False
+    availability_callbacks: list[path_action_availability_callback] | None = None
+
+
+ActionTypes = Action | FileTabAction | PathAction
```
Akuli, line 49: IMO this union should be named `Action`. You may need to use the `Union[Foo, Bar, Baz]` syntax here to support some of the older Python versions we test in CI.

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+                raise RuntimeError(f"{path} does not exist.")
+            if path.is_dir():
+                raise RuntimeError(
+                    f"{path} is a directory - an action consumer registered this action incorrectly"
+                )
+            if not path.is_file():
+                raise RuntimeError(f"{path} is not a file")
+
+            # return True if get_filetype_from_path(path) in filetypes else False
+            raise NotImplementedError  # TODO: there is a way to do this already right?
```
Akuli, line 136: This is indeed tricky, because `filetypes` is a plugin. Maybe the entire `filetype_availability` callback factory should be defined in the filetypes plugin, just like all other filetypes stuff? You can still import from there.

porcupine/actions.py
```diff
@@ -0,0 +1,140 @@
...
+            return False
+
+        if isinstance(context, Path):
+            path = context
+
+            if not path.exists():
+                raise RuntimeError(f"{path} does not exist.")
+            if path.is_dir():
+                raise RuntimeError(
+                    f"{path} is a directory - an action consumer registered this action incorrectly"
```
Akuli, line 130: IMO it should just return `False`. The command palette has everything "registered" anyways, and it will need to test the availability of all actions when finding the one that the user wants.

Akuli (commented):
I like the design overall :)

Akuli:
I think I want the 3 separate functions, at least for now.

Akuli:
> I spent literally 3 hours trying to figure this out

Feel free to ask me if I'm online :D

porcupine/actions.py
benjamin-kirkbride: for the `availability_callbacks`?

I think that the lists will help a lot with DX because we can create helper utilities to implement common uses, and you can compose them in the list.

Consider the different things that someone might need to check to see if a action is available:
- filetype of tab
- filetype of file
- name of file
- if a Path is the "project root"
- if a Path is `.gitignore`'d
- contents of the FileTab selection
- contents of the FileTab `insert` line
- contents of the clipboard
- whether a FileTab has unsaved changes
- contents of the Filetab undo stack
- if internet is available
- various config options
- if plugins are available

I'm sure there are more.

The thought is instead of each action plugin having to create their own availability callbacks, we could let people "compose" them in a list from the helpers we created. Obviously if something is needed that doesn't exist the plugin dev could use that too.

```python
register_path_action(
    name="Update requirements.txt",
    description="Generate requirements.txt from `pyproject.toml`",
    callback=...,
    availability_callbacks=[plugin_availability("requirements_update"), internet_availability(), file_extension_availability(["txt"]), file_name_availability(r"requirements*.txt"), filetype_availability("Python Requirements") ],
)
```

You might think that the above is contrived, but I don't think so. The idea also is that the availability checks will execute from the lowest index in the list up, so you would want to put the cheap and broad checks at the beginning to make sure it's fast.

porcupine/actions.py
benjamin-kirkbride: I agree with this, what about `RootAction` or `StateAction`, (all?) actions will need some kind of state, the non-path and non-filetab ones would just get it from `state` probably?

porcupine/actions.py
benjamin-kirkbride: does `from __future__ import Anotations` not solve that?

porcupine/actions.py
benjamin-kirkbride: I need to think about this more, but the thought I had is that passing a directory to this availability function doesn't make sense; directories don't have filetypes.

Maybe we need a  4th action type: one that takes directories.

Akuli: I imagined it would be a boolean in `PathAction`. So if you set the flag to `True`, it makes the action available for directories without even calling the user's availability callback.

porcupine/actions.py
Akuli: Indeed, we need to document the spreading.

porcupine/actions.py
Akuli: As far as the runtime is concerned, this is an assignment, `foo = bar`. The right side must be evaluated. The `__future__` thing helps only with `:` syntax, as in `foo: bar`.

porcupine/actions.py
Akuli: For now I'd suggest sticking with `get_main_window()` / `get_tab_manager()` instead of injecting. We can consider changing it later, but that would be quite a big change and well beyond the scope of the PR :)

I don't really like either naming suggestion:
- `RootAction` sounds like you can only have one of them, e.g. root file system or root user.
- `StateAction` is confusing without the context you have in mind, because it doesn't have any kind of state or access to the state that other actions lack.

porcupine/actions.py
Akuli: It sounds to me that we're reinventing callback functions. With `def requirements_update_is_available() -> bool`, it is very obvious how the thing works, and you can still reuse existing things easily.

porcupine/actions.py
benjamin-kirkbride: BareAction?

porcupine/actions.py
benjamin-kirkbride: Yeah I see that that's correct. What versions of Python does Porcupine support?

porcupine/actions.py
benjamin-kirkbride: Well, see that gets tricky. Some actions might **only** work on a directory, some might work on both files and directories.

Also there is the question of recursion, not exactly sure how that should be handled :sweat:

porcupine/actions.py
benjamin-kirkbride: Can you expand on that? Are you saying instead of:
```python
    availability_callbacks=[plugin_availability("requirements_update"), internet_availability(), file_extension_availability(["txt"]), file_name_availability(r"requirements*.txt"), filetype_availability("Python Requirements") ],
```

You would suggest:
```python
def requirements_update_is_available() -> bool:
    for callback in [plugin_availability("requirements_update"), internet_availability(), file_extension_availability(["txt"]), file_name_availability(r"requirements*.txt"), filetype_availability("Python Requirements") ]:
        available = callback()
        if not available:
            return False

    return True

...
    availability_callback=requirements_update_is_available
```
?

Obviously that does work, but I'm not really sure I'd prefer it..

porcupine/actions.py
Akuli: 3.8 and newer, though pyls is terrible on 3.11.

porcupine/actions.py
Akuli: More like:
```python3
def requirements_update_is_available() -> bool:
    return (
        plugin_is_enabled("requirements_update")
        and connected_to_internet()
        and get_file_extension(get_current_tab().path) == "txt"
        and glob.match(something r"requirements*.txt")
        and get_current_filetype() == "Python Requirements"
    )

...
    availability_callback=requirements_update_is_available
```

Or if you prefer:

```python3
...
    availability_callback=lambda: (
        plugin_is_enabled("requirements_update")
        and connected_to_internet()
        and get_file_extension(get_current_tab().path) == "txt"
        and glob.match(something r"requirements*.txt")
        and get_current_filetype() == "Python Requirements"
    )
```

One more advantage is the explicit `and` keyword. With a list, you don't know whether it's `or`ing or `and`ing the individual conditions. IMO both can be what the user wants, so it's better to be explicit.

benjamin-kirkbride: Okay, that makes sense, and it does cover the case I was trying to cover of executing from right to left only until the first falsey value:

```python
def false():
    print("false")
    return False


def true():
    print("true")
    return True


print(false() and true())
```

returns:
```
false
False
```

So if you have expensive availability callbacks, you can put them at the end and they won't run unless everything else passes :)

porcupine/actions.py
Akuli: Let's make it two booleans: does it work with files (default: True), and does it work with directories (default: False). Now any combination is possible, and all except False+False make sense, so maybe throw an error if both are False.

We should solve recursion later, but my hunch is that the best solution is to do nothing, since searching the file system recursively can be slow (e.g. accidentally right-clicking `/home` on a computer that doesn't have an SSD).

porcupine/actions.py
Akuli: Yes, this pattern is more common in some other languages (e.g. C) but also works in python :)

benjamin-kirkbride: Wait actually this doesn't quite work as you have conceived it, I think?
```python
...
    availability_callback=lambda: (
        plugin_is_enabled("requirements_update")
        and connected_to_internet()
        and get_file_extension(get_current_tab().path) == "txt"
        and glob.match(something r"requirements*.txt")
        and get_current_filetype() == "Python Requirements"
    )
```

what context do these execute? they execute at import time, right? maybe I'm confused... but taking my example of `filetype_availability("Python Requirements")`, that **returns a `Callable`**, not a bool, so wouldn't `and` not work?

In fact, I'm realizing my example was wrong based on that, it would need to be:
```python
def requirements_update_is_available() -> bool:
    for callback in [plugin_availability("requirements_update"), internet_availability(), file_extension_availability(["txt"]), file_name_availability(r"requirements*.txt"), filetype_availability("Python Requirements") ]:
        available = callback(context=context) # this is the line that changes 
        if not available:
            return False

    return True

...
    availability_callback=requirements_update_is_available
```

aka the availability callbacks get injected with the same context that the actual action callback does (FileTab, Path, or nothing as of now).

Does this work the way you have concieved?

Akuli: Because it is inside a lambda, it executes when the callback is called, i.e. when the availability needs to be determined. For actions that take context (so anything except `BareAction`), it would be `lambda context: (... and ... and ...)`. That's why the list in `filemanager.py` has `lambda p: ...` conditions.

porcupine/actions.py
Akuli: To make this easy for consumers, I'd suggest adding a new `is_available(self, path: Path)` method that considers the booleans, and if needed, invokes the callback.

To be clear, the booleans will be a convenience thing that allows us to skip calling the callback when e.g. the path is a directory but we expect a file. You could just check it in the callback, but it seems easy to forget to return `False` for directories, so booleans could help with that.

Now I'm no longer sure if we need two booleans. If you only want directories, set `accepts_directories=True` and make your callback function reject non-directories?

porcupine/actions.py
Akuli: Ah, now I see your confusion:

>filetype_availability("Python Requirements"), that returns a Callable, not a bool, so wouldn't and not work?

If it returns a callback, it would look like `... and filetype_availability("Python Requirements")(path) and ...`. Not very nice. I would just rewrite the availability function so that I can write `... and get_filetype_name_for_path(path) == "Python Requirements" and ...`

benjamin-kirkbride: so would each of them need to be partials?

```python
...
    availability_callback=lambda tab: (
        partial(plugin_is_enabled("requirements_update"), tab)
        and connected_to_internet()
        and get_file_extension(tab.path) == "txt"
        and glob.match(tab.path.filename() r"requirements*.txt")
        and partial(get_current_filetype(), tab) == "Python Requirements"
    )
```

?

One of us is missing something (probably me)

porcupine/actions.py
benjamin-kirkbride: It feels like we have basically come full circle :sweat_smile:  why not just have availability callbacks for `accepts_directories` and `accepts_files`?

porcupine/actions.py
Akuli: None of them would be partials, because `and`ing `Callable`s doesn't work. Instead we invoke all `Callable`s inside the lambda until we get `bool`s for `and`ing.

For example, consider `connected_to_internet()`. It is a plain old function that returns a boolean: `def connected_to_internet() -> bool`. You use it in the `and` chain just like anything else.

Because this is wrapped in the `lambda`, it doesn't get evaluated right away, and if the user connects to the internet after starting Porcupine, the action will become available as expected.

benjamin-kirkbride: Okay, so I guess it would actually be like this:
```python
...
    availability_callback=lambda tab: (
        plugin_is_enabled("requirements_update")
        and connected_to_internet()
        and get_file_extension(tab.path) == "txt"
        and glob.match(tab.path.filename(), r"requirements*.txt")
        and get_current_filetype(tab) == "Python Requirements"
    )
```

So no partials needed?

porcupine/actions.py
Akuli: Indeed :D

If we go with my `lambda` idea from the other comment, we don't even need to predefine availability callbacks for this: path objects already have convenient `.is_file()` and `.is_dir()` methods :)

porcupine/actions.py
Akuli: Yes, that's how it would work :)

benjamin-kirkbride: Okay thanks! I will write a couple of these proposed availability functions and see if it all starts to come together, but I think we are on the same page now :)

porcupine/actions.py
benjamin-kirkbride: Yep! Sounds good!

porcupine/actions.py
benjamin-kirkbride: this is fixed

porcupine/actions.py
benjamin-kirkbride: fixed by 9c330d3

porcupine/actions.py
benjamin-kirkbride: noted in https://github.com/Akuli/porcupine/issues/1357

porcupine/actions.py
benjamin-kirkbride: tracked in https://github.com/Akuli/porcupine/issues/1357

porcupine/actions.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Callable, Union
+
+from porcupine.tabs import FileTab
+
+action_availability_callback = Callable[[], bool]
```
Akuli, line 9: I think these type aliases are no longer needed.

porcupine/actions.py
```diff
@@ -0,0 +1,115 @@
...
+    callback: Callable[[Path], None]
+    availability_callback: path_action_availability_callback
+
+
+Action = Union[BareAction, FileTabAction, PathAction]
+
+_actions: dict[str, Action] = {}
+
+
+def register_action(
```
Akuli, line 53:
```suggestion
def register_bare_action(
```

porcupine/actions.py
```diff
@@ -0,0 +1,115 @@
...
+        name=name,
+        description=description,
+        callback=callback,
+        availability_callback=availability_callback,
+    )
+    _actions[name] = action
+    return action
+
+
+def query_actions(name: str) -> Action | None:
```
Akuli, line 110:
```suggestion
def find_action(name: str) -> Action | None:
```

Akuli (commented):
Just a few nits.

porcupine/actions.py
benjamin-kirkbride: It's used in a few places, or you mean it's just unnecessary to define them?

Akuli: It's unnecessary to define them. Understanding `Callable[[], bool]` is much easier than understanding `action_availability_callback`. The type aliases were more useful with `list[action_availability_callback]`, but we can simplify since we no longer have a list of callbacks.

benjamin-kirkbride: fair enough

benjamin-kirkbride: 253fdca

benjamin-kirkbride:
I think it's fair to call "actions.py: the action classes, the registry and the related functions" done now

porcupine/menubar.py
```diff
@@ -328,6 +330,31 @@ def setup() -> None:
...
+    * Run :class:`~porcupine.actions.FileTabAction.callback` when the
+      menu item is clicked.
+
+    The ``callback`` is called with the selected tab as the only
+    argument when the menu item is clicked.
+
+    You usually don't need to provide any keyword arguments in ``**kwargs``,
+    but if you do, they are passed to :meth:`tkinter.Menu.add_command`.
+    """
+    menu_path, item_text = _split_parent(path)
```
Akuli, line 351: Presumably we could take only the `menu_path` as an argument and infer `item_text` through the action object.

Akuli (commented):
The last nit...

porcupine/menubar.py
Akuli: Maybe I should clarify that `_split_parent()` does `"Foo/Bar/Baz"` --> `("Foo/Bar", "Baz")` conversion. Doctests would be very useful here :)

benjamin-kirkbride: so lets talk about that.

I assume you mean use the `Action.name` attribute. That does make sense at first blush, my concern is does that limit us to what we can put in the menu item, things like formatting or something. I think it's probably fine though?

benjamin-kirkbride: Other problem I see is that this changes the API, but actually it doesn't have to. If you give a "fully qualified" path, like `topmenu/submenu/itemname` then we treat it like we always have, but if you just do `topmenu/submenu/` with the trailing `/`, then we use the name.

Akuli: > limit us to what we can put in the menu item

My plan for the menubar is:
- Use this function for almost everything
- Expose the menus as-is for custom stuff (e.g. funny named items, items that are more like checkboxes like the filetype chooser, anything else custom)

So if this function covers 99% of use cases, that's enough :)

benjamin-kirkbride: I didn't read your message till I already did it, it would be easy enough to undo though: 0e77816

benjamin-kirkbride (commented):
Does this work as we expect? `tab` being optional means that all of the availability checks need to be written in a way where they can handle no tab, but I think if there is no tab... there is no availability, so we are done here.

porcupine/menubar.py
Akuli: And for whatever reason, I didn't tell you my thoughts about trailing `/` magic :)

I think people expect a trailing `/` to do nothing, like how `ls /home/akuli/` is really `ls /home/akuli`. While a little bit of trailing slash magic is fine, e.g. `ls file.txt/` does not work, this goes a bit too far IMO.

Akuli:
I like how the action consumers are responsible for "no FileTab" handling :)

porcupine/menubar.py
benjamin-kirkbride: Okay, so to be clear, the only supported way to define the label text for `add_filetab_action` will be via `Action.name`?

Akuli: I think that would be good, yes.

porcupine/menubar.py
```diff
@@ -328,6 +335,35 @@ def setup() -> None:
...
+    You usually don't need to provide any keyword arguments in ``**kwargs``,
+    but if you do, they are passed to :meth:`tkinter.Menu.add_command`.
+    """
+
+    get_menu(path).add_command(
+        label=action.name, command=lambda: action.callback(get_filetab()), **kwargs
+    )
+    set_enabled_based_on_tab(
+        path,
+        callback=lambda tab: (lambda tab: isinstance(tab, tabs.FileTab))(tab)
```
rdbende, line 362: Man, this looks cursed 😩
There is way too much one-liner lambda thingy here.

Akuli: Less cursed and does the same thing (even with the `and` on the next line):
```suggestion
        callback=lambda tab: isinstance(tab, tabs.FileTab)
```

porcupine/menubar.py
```diff
@@ -328,6 +335,34 @@ def setup() -> None:
...
 
+def add_filetab_action(path: str, action: actions.FileTabAction, **kwargs: Any) -> None:
+    """
+    This is a convenience function that does several things:
+
+    * Create a menu item at the given path with action.name as label
+    * Ensure the menu item is enabled only when the selected tab is a
+      :class:`~porcupine.tabs.FileTab` AND when
+      :class:`~porcupine.actions.FileTabAction.availability_callback`
+      evaluates to True.
```
Akuli, line 346:
```suggestion
      returns True.
```

porcupine/menubar.py
```diff
@@ -328,6 +335,34 @@ def setup() -> None:
...
+
+    You usually don't need to provide any keyword arguments in ``**kwargs``,
+    but if you do, they are passed to :meth:`tkinter.Menu.add_command`.
+    """
+
+    get_menu(path).add_command(
+        label=action.name, command=lambda: action.callback(get_filetab()), **kwargs
+    )
+    set_enabled_based_on_tab(
+        path,
```
Akuli, line 361:
```suggestion
        path + "/" + action.name,
```

Akuli, line 361:
```suggestion
        path + "/" + action.name,
```

porcupine/plugins/python_tools.py
```diff
@@ -63,6 +63,21 @@ def format_code_in_textwidget(tool: str, tab: tabs.FileTab) -> None:
             tab.textwidget.replace("1.0", "end - 1 char", after)
 
 
+black_format_tab = actions.register_filetab_action(
+    name="Black Format Tab",
+    description="Autoformat open tab using Black",
+    callback=partial(format_code_in_textwidget, "black"),
+    availability_callback=lambda tab: isinstance(tab, tabs.FileTab),
```
Akuli, line 70: `FileTabAction` is never available for tabs other than `FileTab`s:

```suggestion
    availability_callback=lambda tab: True,
```

porcupine/plugins/python_tools.py
```diff
@@ -63,6 +63,21 @@ def format_code_in_textwidget(tool: str, tab: tabs.FileTab) -> None:
...
+)
+
+isort_format_tab = actions.register_filetab_action(
+    name="isort Format Tab",
+    description="Sort Imports of open tab with isort",
+    callback=partial(format_code_in_textwidget, "isort"),
+    availability_callback=lambda tab: isinstance(tab, tabs.FileTab),
```
Akuli, line 77: Same here
```suggestion
    availability_callback=lambda tab: True,
```

porcupine/plugins/python_tools.py
```diff
@@ -63,6 +63,21 @@ def format_code_in_textwidget(tool: str, tab: tabs.FileTab) -> None:
             tab.textwidget.replace("1.0", "end - 1 char", after)
 
 
+black_format_tab = actions.register_filetab_action(
```
Akuli, line 66: These should be in `setup()`, so that they will not run when the plugin is disabled, even if someone decides to import it for some reason.

porcupine/plugins/python_tools.py
```diff
@@ -63,6 +63,21 @@ def format_code_in_textwidget(tool: str, tab: tabs.FileTab) -> None:
...
+    callback=partial(format_code_in_textwidget, "isort"),
+    availability_callback=lambda tab: isinstance(tab, tabs.FileTab),
+)
+
+
 def setup() -> None:
-    menubar.add_filetab_command("Tools/Python/Black", partial(format_code_in_textwidget, "black"))
-    menubar.add_filetab_command("Tools/Python/Isort", partial(format_code_in_textwidget, "isort"))
+    menubar.add_filetab_action("Tools/Python/", black_format_tab)
+    menubar.add_filetab_action("Tools/Python/", isort_format_tab)
```
Akuli, line 83: Does the trailing slash still work? I think we changed how it works.

porcupine/plugins/python_tools.py
benjamin-kirkbride: good call

porcupine/plugins/python_tools.py
benjamin-kirkbride: So I'm not sure I agree with this. This kind of goes back to the whole "should it be possible to access actions from plugins directly" thing.

That said, I'll go ahead and move them into there now for the initial implementation and we can debate the merits of having them available outside actions module another time ;)

benjamin-kirkbride: 1edd389

porcupine/plugins/python_tools.py
benjamin-kirkbride: You are right.

benjamin-kirkbride:
@Akuli any idea what is up with this?
```python
    isort_format_tab_action = actions.register_filetab_action(
        name="isort Format Tab",
        description="Sort Imports of open tab with isort",
        callback=partial(format_code_in_textwidget, "isort"),
        availability_callback=lambda tab: tab.settings.get("filetype_name", object) == "Python",
    )
```

Note `availability_callback`

When I open Porcupine:

```python
  File "/usr/lib/python3.11/tkinter/__init__.py", line 1948, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/home/bkirkbride/projects/porcupine/porcupine/menubar.py", line 297, in update_enabledness
    menu.entryconfig(index, state=("normal" if callback(tab) else "disabled"))
                                               ^^^^^^^^^^^^^
  File "/home/bkirkbride/projects/porcupine/porcupine/menubar.py", line 362, in <lambda>
    callback=lambda tab: isinstance(tab, tabs.FileTab) and action.availability_callback(tab),
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bkirkbride/projects/porcupine/porcupine/plugins/python_tools.py", line 78, in <lambda>
    availability_callback=lambda tab: tab.settings.get("filetype_name", object) == "Python",
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bkirkbride/projects/porcupine/porcupine/settings.py", line 332, in get
    result = self._options[option_name].value
             ~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'filetype_name'
```

And the menu items are not there.

If I switch to another tab and switch back, then it works, no more errors.

I assume this is some sort of dependency issue where the options don't get loaded before the plugins do, or something. Thoughts on how to handle this?

porcupine/menubar.py
```diff
@@ -282,15 +284,22 @@ def setup():
...
         if index is None:
             raise LookupError(f"menu item {path!r} not found")
+        if not tab:
+            menu.entryconfig(index, state="disabled")
+            return
+
         menu.entryconfig(index, state=("normal" if callback(tab) else "disabled"))
 
     update_enabledness()
+    get_tab_manager().bind("<<TabFiletypeApplied>>", update_enabledness, add=True)
```
Akuli, line 300: This makes the core Porcupine aware of filetypes. Before this PR they were entirely a `plugins/filetypes.py` concept.

But I'm not sure if this is avoidable. I'll think about this a bit and respond after a couple days. Ping me if you don't hear back :)

benjamin-kirkbride: Well, we could create some sort of a.. registry for tracking what events should trigger the `update_enabledness` callback. Honestly not a bad idea IMO, let me throw that together.

benjamin-kirkbride: Basically action consumers have a registry of virtual events that should trigger them to refresh their available actions, and action providers can register events to them.

benjamin-kirkbride: @Akuli 8128316

benjamin-kirkbride: Aren't event bindings somewhat expensive? This is `events_to_trigger_on = bindings` vs  `events_to_trigger_on*menu_items = bindings`

Akuli:
The discussion on this PR has become a bit tl;dr to follow. I will hopefully have time to look at this properly during the weekend :)

benjamin-kirkbride:
@Akuli any update on this? I think it's in a good state.

porcupine/actions.py
```diff
@@ -0,0 +1,127 @@
...
+
+def register_bare_action(
+    *,
+    name: str,
+    description: str,
+    callback: Callable[..., None],
+    availability_callback: Callable[[], bool] = lambda: True,
+) -> BareAction:
+    if name in _actions:
+        raise ValueError(f"Action with the name '{name}' already exists")
```
rdbende, line 54:
```suggestion
        raise ValueError(f"Action with the name {name!r} already exists")
```

Using `!r` is equivalent to calling `repr` on the object, which is safer to use. For example it would format `I'm stupid` as `"I'm stupid"` and not as `'I'm stupid'`.

porcupine/actions.py
rdbende: Was this resolved?

porcupine/menubar.py
```diff
@@ -280,18 +296,23 @@ def setup():
...
         parent, child = _split_parent(path)
         menu = get_menu(parent)
         index = _find_item(menu, child)
         if index is None:
             raise LookupError(f"menu item {path!r} not found")
+        if not tab:
+            menu.entryconfig(index, state="disabled")
+            return
+
         menu.entryconfig(index, state=("normal" if callback(tab) else "disabled"))
```
rdbende, line 311:
```suggestion
        if tab and callback(tab):
            menu.entryconfig(index, state="normal")
        else:
            menu.entryconfig(index, state="disabled")
```

A bit cleaner

porcupine/actions.py
```diff
@@ -0,0 +1,127 @@
...
+        name=name,
+        description=description,
+        callback=callback,
+        availability_callback=availability_callback,
+    )
+    _actions[name] = action
+    return action
+
+
+def register_path_action(
```
rdbende, line 84: These functions are a bit repetitive. I haven't kept up with your discussion of the design, but perhaps they could use a helper function to do the actual registering.

rdbende (changes requested):
Some nits, but otherwise LGTM

porcupine/actions.py
benjamin-kirkbride: TIL!

porcupine/menubar.py
```diff
@@ -11,6 +11,8 @@
 from tkinter import filedialog
 from typing import Any, Callable, Iterator
 
+from porcupine import actions
+
 if sys.version_info >= (3, 8):
```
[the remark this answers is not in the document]
benjamin-kirkbride: I wonder if there is a linter or something for catching these automatically..

porcupine/menubar.py
benjamin-kirkbride: probably will do `tab is not None and callback(tab)`, as `None` is what we are actually checking for and that is the more pythonic (I believe) way to check for that.

porcupine/actions.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def get_all_actions() -> dict[str, Action]:
+    return _actions.copy()
+
+
+# Availability Helpers
+
+
+def filetype_is(filetypes: Union[list[str], str]) -> Callable[[FileTab], bool]:
```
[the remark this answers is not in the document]
benjamin-kirkbride: oh okay, thanks for the clarification. I'll do that

porcupine/menubar.py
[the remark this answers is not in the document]
rdbende: There is: https://github.com/asottile/pyupgrade

porcupine/menubar.py
rdbende: Yes, we are actually checking for `not None`, but I think it is not necessary to explicitly check for that, because a `FileTab` widget will never be "falsy" anyway.

porcupine/menubar.py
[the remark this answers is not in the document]
Akuli: I wouldn't be surprised if ruff has something for this too.

porcupine/actions.py
Akuli: This is fine as is.

porcupine/menubar.py
Akuli: Tbh, at this point I just want this merged and I don't really care :D

porcupine/menubar.py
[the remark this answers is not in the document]
rdbende: Yeah, and my problem with `pyupgrade` is that I have to `git add -p` to select the actually useful changes.

porcupine/menubar.py
benjamin-kirkbride: `tab` can be none, I found this the hard way.

Akuli: Suggesting the explicit thing that we all know will work:

```suggestion
        if tab is not None and callback(tab):
            menu.entryconfig(index, state="normal")
        else:
            menu.entryconfig(index, state="disabled")
```

porcupine/menubar.py
```diff
@@ -246,9 +248,23 @@ def update_keyboard_shortcuts() -> None:
...
+# TODO: create type for events
+def register_enabledness_check_event(event: str) -> None:
+    """Register an event which will cause all menu items to check if they are available"""
+    get_tab_manager().bind(event, _refresh_menu_item_enabledness, add=True)
+
+
 def set_enabled_based_on_tab(
-    path: str, callback: Callable[[tabs.Tab | None], bool]
-) -> Callable[..., None]:
+    path: str, callback: Callable[[tabs.Tab], bool], filetab_only: bool = False
```
Akuli, line 266: It seems that `filetab_only` is never used.

rdbende: And the docstring has this:

```py
menubar.set_enabled_based_on_tab("Foo/Bar", (lambda tab: isinstance(tab, tabs.FileTab)))
```

porcupine/actions.py
rdbende: @benjamin-kirkbride ?

benjamin-kirkbride: Sorry, somehow missed this.

How about `get_actions` since it can return None (and in fact is just a thin wrapper around `dict.get`)?

rdbende: Fine for me.

rdbende: But `get_action` then. So not plural.

benjamin-kirkbride: Yes, of course :sweat_smile:

Moosems: It looks like a consensus has been reached 😄.

benjamin-kirkbride:
Only thing left at this point is docs IMO. Is it worth doing that before more stuff uses actions?

rdbende (approved):

porcupine/menubar.py
```diff
@@ -241,9 +241,21 @@ def update_keyboard_shortcuts() -> None:
     _update_shortcuts_for_opening_submenus()
 
 
-def set_enabled_based_on_tab(
-    path: str, callback: Callable[[tabs.Tab | None], bool]
-) -> Callable[..., None]:
+_menu_item_enabledness_callbacks: list[Callable[..., None]] = []
```
Akuli, line 244: Not super happy with this design, but I just want this merged.

Akuli (approved):

Issue #1303, opened 2023-06-28
Replace isort with ruff


benjamin-kirkbride:
Any opinion on doing this?

Moosems:
Ruff doesn't quite sorry the same though :(.

Akuli:
I don't have much experience with ruff, but I am interested in trying it out for Porcupine, not just for formatting but also for other cleanups. For example, ruff can do `Optional[str]` --> `str | None`, and I want that.

Porcupine already has a setup where it runs formatters automatically on pull requests.

Moosems:
I would use both then ;).

benjamin-kirkbride:
```~/projects/porcupine$ ruff check .
docs/conf.py:9:1: E402 Module level import not at top of file
porcupine/__init__.py:37:1: E402 Module level import not at top of file
porcupine/plugins/aboutdialog.py:72:101: E501 Line too long (101 > 100 characters)
porcupine/plugins/desktop_menu.py:29:101: E501 Line too long (108 > 100 characters)
porcupine/plugins/directory_tree.py:414:23: F841 Local variable `width` is assigned to but never used
porcupine/plugins/filetypes.py:98:101: E501 Line too long (107 > 100 characters)
porcupine/plugins/highlight/tree_sitter_highlighter.py:158:101: E501 Line too long (103 > 100 characters)
porcupine/plugins/pastebin.py:195:101: E501 Line too long (101 > 100 characters)
porcupine/plugins/run/no_terminal.py:246:101: E501 Line too long (104 > 100 characters)
porcupine/plugins/sort.py:2:101: E501 Line too long (105 > 100 characters)
porcupine/settings.py:225:101: E501 Line too long (101 > 100 characters)
porcupine/tabs.py:575:101: E501 Line too long (107 > 100 characters)
porcupine/textutils.py:460:101: E501 Line too long (117 > 100 characters)
porcupine/utils.py:308:101: E501 Line too long (106 > 100 characters)
porcupine/utils.py:664:101: E501 Line too long (112 > 100 characters)
porcupine/utils.py:699:101: E501 Line too long (102 > 100 characters)
porcupine/utils.py:796:37: F821 Undefined name `value`
scripts/build-exe-installer.py:17:1: E402 Module level import not at top of file
scripts/release.py:11:1: E402 Module level import not at top of file
tests/test_run_plugin.py:103:101: E501 Line too long (111 > 100 characters)
[...]

PR #1356, opened 2023-07-07
Add ruff config

benjamin-kirkbride:
progress towards #1303

Can we get this merged ahead of time? I'm already using ruff for linting and import sorting, but the long line errors drive me mad :weary:


Akuli:
Test failure is unrelated.

Issue #1333, opened 2023-07-04
Crash when trying to open project with CMD+O


ThePhilgrim:
Porcupine crashes on Mac when trying to open a project with CMD+O.

**System:**
Mac M2 Ventura 13.4.1.

**Error output:**
```
❯ python3 -m porcupine
log file: /Users/phil/Library/Logs/Porcupine/2023-07-04T23-19-52.txt
porcupine.plugins.drop_to_open ERROR: dragging files to Porcupine won't work because tkdnd isn't installed
2023-07-04 23:19:55.251 Python[7362:137264] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001a5567154 __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x00000001a50864d4 objc_exception_throw + 60
	2   CoreFoundation                      0x00000001a56519b8 -[__NSCFString characterAtIndex:].cold.1 + 0
	3   CoreFoundation                      0x00000001a564df80 -[__NSArrayM insertObject:atIndex:].cold.2 + 0
	4   CoreFoundation                      0x00000001a548f28c -[__NSArrayM insertObject:atIndex:] + 912
	5   libtk8.6.dylib                      0x000000010128cee4 setAllowedFileTypes + 176
	6   libtk8.6.dylib                      0x000000010128d8f0 Tk_GetOpenFileObjCmd + 1408
	7   libtcl8.6.dylib                     0x0000000100f0b068 TclNRRunCallbacks + 80
	8   _tkinter.cpython-311-darwin.so      0x0000000100d712b0 Tkapp_Call + 400
	9   Python                              0x000000010080ea0c cfunction_call + 100
	10  Python                              0x00000001007c7df8 _PyObject_Call + 124
	11  Python                              0x00000001008a105c _PyEval_EvalFrameDefault + 48368
	12  Python                              0x00000001008a38e4 _PyEval_Vector + 116
	13  Python                              0x00000001007c7a34 _PyVectorcall_Call + 116
	14  Python                              0x00000001008a105c _PyEval_EvalFrameDefault + 48368
	15  Python                              0x00000001008a38e4 _PyEval_Vector + 116
	16  Python                              0x00000001008a105c _PyEval_EvalFrameDefault + 48368
[...]

PR #1358, opened 2023-07-08
Get rid of filedialog_kwargs

Akuli:
It will no longer be possible to view e.g. only Python files in the "Open" dialog. I think this is a good thing:
- It is somewhat rare to have lots of files with mixed filetypes in the same directory, the only situation where filtering is actually useful.
- I never used the filtering, and I don't think anyone else used it either.
- Some filetypes caused problems, e.g. Makefiles are recognized by the *prefix* of the name (`Makefile` and `Makefile.foo` are typical names for them) and tkinter doesn't support this on all platforms.
- There were crashes on mac (#1092 and #1333). These are difficult to get right and difficult to test.

### DIFF
--- a/docs/porcupine.rst
+++ b/docs/porcupine.rst
@@ -39,20 +39,6 @@ Version Information
           # display an error message or do things without the new feature
 
 
-File Dialogs
-------------
-
-.. data:: filedialog_kwargs
-    :type: dict[str, Any]
-
-    Porcupine uses :mod:`tkinter.filedialog` functions similarly to this::
-
-        path = filedialog.asksaveasfilename(**porcupine.filedialog_kwargs)
-
-    The :source:`filetypes plugin <porcupine/plugins/filetypes.py>` uses this
-    for displaying known filetypes in the dialogs.
-
-
 Directories
 -----------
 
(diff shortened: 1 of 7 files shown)


benjamin-kirkbride:
Do we consider making a new filedialog at some point? The tkinter one is really not great for all kinds of reasons.

rdbende:
On Windoze and macOS it uses the native file browser, so I don't think we should change that, even if the macOS one is buggy. On Linux however it would be a good thing to use something else, because Tk's filedialog SUCKS.
If we want this, we could use the `kdialog` command line utility (KDE Plasma), and `zenity` (GNOME) to invoke the native file choosers, or simply replace the built-in filedialogs with the [`tkfilebrowser`](https://github.com/j4321/tkFileBrowser) package.

benjamin-kirkbride:
TIL that it tk uses native file dialogs on not-linux...

Yes, agree completely with everything @rdbende said. My thoughts is it should work like this:
1. if windows or mac use default filedialog as now; if linux continue
2. check config to see what user wants to do; if no config set continue
3. try `zenity`
4. try `kdialog`
5. try tkfilebrowser (optional install?)
6. fallback to default `tk.filedialog`

PR #1360, opened 2023-07-09
Add new keybinding for commenting/uncommenting current line without needing to select it

### LINKED ISSUE #1350: Comment/uncomment current line without selecting
Currently, I have to highlight at least one character on a row in order to use the function _Comment/uncomment selected lines_. I think this behavior is fine if I want to comment/uncomment several lines at once. 

**Desired behavior:**
If I want to comment out only one line, I should be able to do that simply by having the cursor set to that specific line.

### LINKED ISSUE #1355: Show keybinding for comment/uncomment row in menu
As seen in the screenshot below, there is no keybinding showing for the _comment/uncomment selected line(s)_. I had to go into `README.md`, which is not exactly obvious.

Screenshot 2023-07-07 at 21 19 52

ThePhilgrim:
Fixes #1355
Fixes #1350

### DIFF
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -60,6 +60,9 @@ event add "<<Menubar:Edit/Find and Replace>>" <$control_ish-f>
 # fold plugin
 event add "<<Menubar:Edit/Fold>>" <$alt_ish-f>
 
+# comment/uncomment plugin
+event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <$control_ish-7>
+
 # anchor plugin
 event add "<<Menubar:Edit/Anchors/Add or remove on this line>>" <$alt_ish-A>
 event add "<<Menubar:Edit/Anchors/Jump to previous>>" <$alt_ish-Shift-Up>
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -61,7 +61,7 @@ event add "<<Menubar:Edit/Find and Replace>>" <$control_ish-f>
 event add "<<Menubar:Edit/Fold>>" <$alt_ish-f>
 
 # comment/uncomment plugin
-event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <Shift-slash>
+event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <$control_ish-slash>
 
 # anchor plugin
 event add "<<Menubar:Edit/Anchors/Add or remove on this line>>" <$alt_ish-A>
--- a/porcupine/default_keybindings.tcl
+++ b/porcupine/default_keybindings.tcl
@@ -61,7 +61,7 @@ event add "<<Menubar:Edit/Find and Replace>>" <$control_ish-f>
 event add "<<Menubar:Edit/Fold>>" <$alt_ish-f>
 
 # comment/uncomment plugin
-event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <$control_ish-7>
+event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <Shift-slash>
 
 # anchor plugin
 event add "<<Menubar:Edit/Anchors/Add or remove on this line>>" <$alt_ish-A>
--- a/porcupine/plugins/comment_selected_lines.py
+++ b/porcupine/plugins/comment_selected_lines.py
@@ -18,7 +18,7 @@ def already_commented(linetext: str, comment_prefix: str) -> Optional[bool]:
     # Ignore '# blah' comments because they are likely written by hand
     # But don't ignore indented '#    blah', that is most likely by this plugin
     if linetext.startswith(comment_prefix) and not re.match(
-        r" [^ ]", linetext[len(comment_prefix)]
+        r" [^ ]", linetext[len(comment_prefix) :]
     ):
         return True
     return None
--- a/porcupine/plugins/comment_selected_lines.py
+++ b/porcupine/plugins/comment_selected_lines.py
@@ -21,6 +21,7 @@ def already_commented(linetext: str, comment_prefix: str) -> Optional[bool]:
         r" [^ ]", linetext[len(comment_prefix)]
     ):
         return True
+    return None
 
 
 def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> str | None:
@@ -32,16 +33,16 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
         start_index, end_index = map(str, tab.textwidget.tag_ranges("sel"))
     except ValueError:
         # No text selected
-        lineno = tab.textwidget.index("insert").split(".")[0]
-        line_start = lineno + ".0"
+        line_no = tab.textwidget.index("insert").split(".")[0]
+        line_start = line_no + ".0"
         line_text = tab.textwidget.get(line_start, f"{line_start} lineend")
 
         if already_commented(line_text, comment_prefix):
-            tab.textwidget.delete(line_start, f"{lineno}.{len(comment_prefix)}")
+            tab.textwidget.delete(line_start, f"{line_no}.{len(comment_prefix)}")
         else:
-            tab.textwidget.insert(f"{lineno}.0", comment_prefix)
+            tab.textwidget.insert(f"{line_no}.0", comment_prefix)
 
-        return
+        return None
 
     start = int(start_index.split(".")[0])
     end = int(end_index.split(".")[0])
--- a/porcupine/plugins/comment_selected_lines.py
+++ b/porcupine/plugins/comment_selected_lines.py
@@ -38,6 +38,7 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
         line_text = tab.textwidget.get(line_start, f"{line_start} lineend")
 
         if pressed_key is None:
+            # Toggle comments on the line where cursor is
             if already_commented(line_text, comment_prefix):
                 tab.textwidget.delete(line_start, f"{line_no}.{len(comment_prefix)}")
             else:
--- a/porcupine/plugins/comment_selected_lines.py
+++ b/porcupine/plugins/comment_selected_lines.py
@@ -19,6 +19,7 @@ def already_commented(linetext: str, comment_prefix: str) -> Optional[bool]:
     if linetext.startswith(comment_prefix) and not re.match(r" [^ ]", linetext[len(comment_prefix)]):
         return True
 
+
 def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> str | None:
     comment_prefix = tab.settings.get("comment_prefix", Optional[str])
     if pressed_key is not None and pressed_key != comment_prefix:
@@ -33,7 +34,7 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
         line_text = tab.textwidget.get(line_start, f"{line_start} lineend")
 
         if already_commented(line_text, comment_prefix):
-           tab.textwidget.delete(line_start, f"{lineno}.{len(comment_prefix)}")
+            tab.textwidget.delete(line_start, f"{lineno}.{len(comment_prefix)}")
         else:
             tab.textwidget.insert(f"{lineno}.0", comment_prefix)
 
--- a/porcupine/plugins/comment_selected_lines.py
+++ b/porcupine/plugins/comment_selected_lines.py
@@ -13,10 +13,13 @@ from typing import Optional
 from porcupine import get_tab_manager, menubar, tabs, textutils
 from porcupine.plugins import rightclick_menu
 
+
 def already_commented(linetext: str, comment_prefix: str) -> Optional[bool]:
     # Ignore '# blah' comments because they are likely written by hand
     # But don't ignore indented '#    blah', that is most likely by this plugin
-    if linetext.startswith(comment_prefix) and not re.match(r" [^ ]", linetext[len(comment_prefix)]):
+    if linetext.startswith(comment_prefix) and not re.match(
+        r" [^ ]", linetext[len(comment_prefix)]
+    ):
         return True
 
 
@@ -51,7 +54,8 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
         lineno
         for lineno, line_text in enumerate(
             tab.textwidget.get(f"{start}.0", f"{end}.0").splitlines(), start
-        ) if already_commented(line_text, comment_prefix)
+        )
+        if already_commented(line_text, comment_prefix)
     }
 
     with textutils.change_batch(tab.textwidget):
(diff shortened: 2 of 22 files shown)


porcupine/default_keybindings.tcl
```diff
@@ -60,6 +60,9 @@ event add "<<Menubar:Edit/Find and Replace>>" <$control_ish-f>
 # fold plugin
 event add "<<Menubar:Edit/Fold>>" <$alt_ish-f>
 
+# comment/uncomment plugin
+event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <$control_ish-7>
```
Akuli, line 64: For people with american keyboards, it would be better if they could use the slash character:

```suggestion
event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <Shift-slash>
```

On my system this works the same as `<$control_ish-7>`. Does it work for you?

ThePhilgrim: `<Shift-slash>` does not work on my system. Can't find any button combination that triggers this.

ThePhilgrim: Sorry, my bad. I keep forgetting that this plugin doesn't support commenting out a single line without explicitly selecting text. `Shift-slash`does work for me.

ThePhilgrim:
@Akuli, Looks like the failed checks are unrelated to the code changes. Can we please rerun the checks?

Moosems:
What about contmand-slash?

Akuli:
> I keep forgetting that this plugin doesn't support commenting out a single line without explicitly selecting text. 

The new key binding (and probably the button in the menubar as well) should always do something, even without explicitly selecting text. But the old key binding should still behave in the old way, so that it is possible to just type a `#` character in the Python file without having it turn into a magic commenting key binding.

Akuli:
> Looks like the failed checks are unrelated to the code changes. Can we please rerun the checks?

Yes. You can trigger a re-run yourself by pushing. If you have no commits to push, you can make an empty commit with e.g. `git commit --allow-empty -m "trigger checks"`.

ThePhilgrim:
> > I keep forgetting that this plugin doesn't support commenting out a single line without explicitly selecting text.
> 
> The new key binding (and probably the button in the menubar as well) should always do something, even without explicitly selecting text. But the old key binding should still behave in the old way, so that it is possible to just type a `#` character in the Python file without having it turn into a magic commenting key binding.

Got it – This will be a part of the next PR

ThePhilgrim:
@Akuli Git doesn't seem to want to cooperate. See below (I've already run the `commit --allow-empty` before this output:

```
❯ git push
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push upstream HEAD:main

To push to the branch of the same name on the remote, use

    git push upstream HEAD

To choose either option permanently, see push.default in 'git help config'.

To avoid automatically configuring an upstream branch when its name
won't match the local branch, see option 'simple' of branch.autoSetupMerge
in 'git help config'.

❯ git lola
❯ clear
❯ git push origin comment_keybinding
To https://github.com/ThePhilgrim/porcupine-phil
 ! [rejected]          comment_keybinding -> comment_keybinding (fetch first)
error: failed to push some refs to 'https://github.com/ThePhilgrim/porcupine-phil'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
❯ git pull
Already up to date.
❯ git push
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push upstream HEAD:main

To push to the branch of the same name on the remote, use

    git push upstream HEAD

To choose either option permanently, see push.default in 'git help config'.

To avoid automatically configuring an upstream branch when its name
won't match the local branch, see option 'simple' of branch.autoSetupMerge
in 'git help config'.
```

rdbende:
`git pull origin comment_keybinding` and `git push origin comment_keybinding`

Akuli:
> This will be a part of the next PR

Uhh, no? The keybinding is quite useless without this :) If the new keybinding is confusing for you when testing it, it is also too confusing for users.

IMO pull requests should be self-contained. To me, "add a new keybinding with new behavior" is one self-contained thing and "change/improve behavior of existing key bindings" is a different thing entirely.

ThePhilgrim:
> `git pull origin comment_keybinding` and `git push origin comment_keybinding`

Many thanks!



> IMO pull requests should be self-contained. To me, "add a new keybinding with new behavior" is one self-contained thing

I completely agree, and I was implementing the keybinding and the single line commenting behavior, but you said on IRC you wanted 

- New keybinding
- Support for single line commenting
- Support for ```

in separate PRs. I don't have the IRC logs so I cannot quote you directly, but this was my interpretation of your feature list. That's why I created this PR with only the keybinding.

If you would like the single line comment in the same PR, I would be happy to implement that too in this PR.

Akuli:
Sorry about being confusing. Let's just do it in the same PR. I'm not sure what I was thinking when I split your ideas into 3 "separate" things :) Also, I don't think I realized that the key binding PR by itself would be as small as this PR is now.

You *do* have the IRC logs though. When mantaray starts, it prints the path to its log folder.

porcupine/default_keybindings.tcl
```diff
@@ -60,6 +60,9 @@ event add "<<Menubar:Edit/Find and Replace>>" <$control_ish-f>
 # fold plugin
 event add "<<Menubar:Edit/Fold>>" <$alt_ish-f>
 
+# comment/uncomment plugin
+event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <Shift-slash>
```
Akuli, line 64: No idea why I suggested Shift-slash. On my keyboard layout, slash means Shift+7 and I wanted this to be Ctrl+Shift+7, so the correct thing is:
```suggestion
event add "<<Menubar:Edit/Comment//uncomment selected lines>>" <$control_ish-slash>
```

Moosems: Thank you!!!!

ThePhilgrim:
Single line commenting should work now. Still missing tests. Maybe they could be included in a separate PR?

Moosems:
Will review tonight :).

porcupine/plugins/comment_selected_lines.py
```diff
@@ -14,6 +14,15 @@
 from porcupine.plugins import rightclick_menu
 
 
+def already_commented(linetext: str, comment_prefix: str) -> Optional[bool]:
+    # Ignore '# blah' comments because they are likely written by hand
```
Moosems, line 18: If I'm trying to uncomment something, does it matter how it was written?

porcupine/plugins/comment_selected_lines.py
```diff
@@ -22,8 +31,17 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
...
-        return None
+        # No text selected
+        lineno = tab.textwidget.index("insert").split(".")[0]
+        line_start = lineno + ".0"
+        line_text = tab.textwidget.get(line_start, f"{line_start} lineend")
+
+        if already_commented(line_text, comment_prefix):
+            tab.textwidget.delete(line_start, f"{lineno}.{len(comment_prefix)}")
+        else:
+            tab.textwidget.insert(f"{lineno}.0", comment_prefix)
```
Moosems, line 42: It may be a big ask but what if we made all the comments look like `# blah` by finding the indents?

porcupine/plugins/comment_selected_lines.py
```diff
@@ -34,12 +52,10 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
...
     commented = {
         lineno
-        for lineno, line in enumerate(
+        for lineno, line_text in enumerate(
             tab.textwidget.get(f"{start}.0", f"{end}.0").splitlines(), start
         )
-        # Ignore '# blah' comments because they are likely written by hand
-        # But don't ignore indented '#    blah', that is most likely by this plugin
-        if line.startswith(comment_prefix) and not re.match(r" [^ ]", line[len(comment_prefix) :])
+        if already_commented(line_text, comment_prefix)
```
Moosems, line 58: Why are comments getting ignored??

Moosems:
Didn't realize individual comments made whole reviews 😶.

porcupine/plugins/comment_selected_lines.py
rdbende: > \# Ignore '# blah' comments because they are likely written by hand

Or `#    blah` comments, formatted with black

ThePhilgrim: @Moosems @rdbende I completely agree. However, @Akuli wants the functionality like this. My first version of this code always did `# foo`, no matter the indentation. However, seems like Akuli wants to separate manually added comments and comments made with keybindings.

porcupine/plugins/comment_selected_lines.py
ThePhilgrim: My first version of this code did look for the indentation and put the comment syntax one char before first indentation level. Similar to how it works in VS Code. For example

```
def foo():
    lol = 1
    if lol == 1:
        print("bar")
        print("baz")
```

If I commented out line 3 & 4 of this block it would be:

```
def foo():
    lol = 1
    # if lol == 1:
    #    print("bar")
    #    print("baz")
```

This is because the following would look really messy:

```
def foo():
    lol = 1
    # if lol == 1:
       # print("bar")
       # print("baz")
```

However, Akuli requested that I keep all comment syntax pushed to the left.

rdbende: @Akuli is right :)

porcupine/plugins/comment_selected_lines.py
ThePhilgrim: Akuli wanted manually added comments to be ignored when the keybinding is used.

porcupine/plugins/comment_selected_lines.py
ThePhilgrim: @rdbende I think formatting `#        blah` with black actually does nothing. It seems to be left like that!

rdbende:
```console
$ cat t.py
def foo():
#    bar = 8
#    bar += 2
#    assert bar == 10
    ...
$ black t.py
reformatted t.py

All done! ✨ 🍰 ✨
1 file reformatted.

$ cat t.py
def foo():
    #    bar = 8
    #    bar += 2
    #    assert bar == 10
    ...
```

porcupine/plugins/comment_selected_lines.py
ThePhilgrim: @rdbende 

With the current state of this PR, the comments look like

```
def foo():
    lol = 1
    if lol == 1:
        if True:
            if True:
#                print("bar")
#                print("baz")
```

The same comment in for example VS Code would look like

```
def foo():
    lol = 1
    if lol == 1:
        if True:
            if True:
            #        print("bar")
            #        print("baz")
```

To clarify, do you think column 0 is better than the indentation level of what is commented?

rdbende: No, I think that comments should be at the indentation level of the outermost block

So this:
```py
def foo():
    lol = 1
    if lol == 1:
        # if True:
        #     if True:
        #         print("bar")
        #         print("baz")
```

instead of this:

```py
def foo():
    lol = 1
    if lol == 1:
        # if True:
            # if True:
                # print("bar")
                # print("baz")
```

rdbende: Black will change it to this anyway :)

```py
def foo():
    lol = 1
    if lol == 1:
        # if True:
        # if True:
        # print("bar")
        # print("baz")
        ...
```

ThePhilgrim: Ah, right. So to clarify

My first version of the code added comment syntax at the indentation level of the outermost block (never published PR). However, Akuli wanted it like follows:

```
def foo():
    lol = 1
    if lol == 1:
#         if True:
#             if True:
#                 print("bar")
#                 print("baz")
```

rdbende: Ah, I see. Then I don't agree woth Akuli. Black will indent that anyway, so it doesn't make sense to me.

porcupine/plugins/comment_selected_lines.py
Moosems: Ummmm ok. He does what he does ig.

ThePhilgrim: To be fair, he is also the project owner, so :)

Akuli:
The discussion is a bit tl;dr, but I think most of it is irrelevant for this PR. I wanted this PR to not touch *how* the indentations are done; that is, where the new `#` characters get placed, which is what most of the discussion seems to be about.

It can be changed in a new PR, but it really has nothing to do with adding a new "just comment something regardless of selection" key binding.

ThePhilgrim:
> The discussion is a bit tl;dr, but I think most of it is irrelevant for this PR. I wanted this PR to not touch _how_ the indentations are done; that is, where the new `#` characters get placed, which is what most of the discussion seems to be about.
> 
> It can be changed in a new PR, but it really has nothing to do with adding a new "just comment something regardless of selection" key binding.

Fair! So would you be okay if comment syntax is placed based on indentation in a future PR?

Akuli (commented):
When I press `#`, it now turns into `##`. Please fix that. As I said, this should not affect old behavior at all. Other than that, this looks good to me :)

Akuli:
>  comment syntax is placed based on indentation in a future PR?

I can't really give a yes/no answer. The current behavior isn't great for blacking, but it has the advantage of being super simple. In any case, this should be discussed somewhere else than in this PR, because this PR has nothing to do with how comment characters and indentations interact apart from a small code cleanup :)

porcupine/plugins/comment_selected_lines.py
```diff
@@ -22,7 +32,17 @@ def comment_or_uncomment(tab: tabs.FileTab, pressed_key: str | None = None) -> s
...
         start_index, end_index = map(str, tab.textwidget.tag_ranges("sel"))
     except ValueError:
-        # nothing selected, add '#' normally
+        # No text selected
+        line_no = tab.textwidget.index("insert").split(".")[0]
+        line_start = line_no + ".0"
+        line_text = tab.textwidget.get(line_start, f"{line_start} lineend")
+
+        if pressed_key is None:
+            if already_commented(line_text, comment_prefix):
```
Akuli, line 41:
```suggestion
            # Toggle comments on the line where cursor is
            if already_commented(line_text, comment_prefix):
```

Akuli (approved):
I'm happy with how it behaves now. We can merge as far as I'm concerned, but since you said you want to add more tests, let me know when you're happy with this :)

ThePhilgrim:
> I'm happy with how it behaves now. We can merge as far as I'm concerned, but since you said you want to add more tests, let me know when you're happy with this :)

I think that's enough testing for this PR. We are testing the keybinding and the single-line commenting, which this PR covers. I think any further PRs with new/changed features will contain relevant tests.

Akuli:
> Didn't realize individual comments made whole reviews :no_mouth:.

That's nothing. Once a guy from IRC "thanked" me for sending 20 emails to them, and the reason was that I had created my first Github review ever by clicking the individual comment button 20 times...

PR #1376, opened 2023-07-12
Sort plugin: use indentation when deciding lines to sort

### LINKED ISSUE #1375: Alt+S / Sort Lines: selects a bit too much
(Now there are 200 open issues, yay)

At work we have text files that look like this:

```
[foo]
    akuli
    dingdingding
    banana
    catris
```

When I press Alt+S in the middle of this, it becomes:

```
    akuli
    banana
    catris
    dingdingding
[foo]
```

because `[` comes after space in ASCII. This isn't great.

It would be better to only sort the lines that have the same indentation level as the current line, so this would instead become:

```
[foo]
    akuli
    banana
    catris
    dingdingding
```

Akuli:
Fixes #1375

### DIFF
--- a/porcupine/plugins/sort.py
+++ b/porcupine/plugins/sort.py
@@ -17,6 +17,7 @@ from porcupine import menubar, tabs, textutils
 
 
 def can_extend_default_selection(text: tkinter.Text, required_prefix: str, new_lineno: int) -> bool:
+    # This returns False when line number is beyond the start or end of the text widget
     line = text.get(f"{new_lineno}.0", f"{new_lineno + 1}.0")
     return bool(line.strip()) and line.startswith(required_prefix)
 
--- a/porcupine/plugins/sort.py
+++ b/porcupine/plugins/sort.py
@@ -1,12 +1,37 @@
 """
-Sort the selected lines alphabetically, or if there's no selection, sort between surrounding blank lines.
+Sort the selected lines alphabetically. Available in menubar → Edit → Sort Lines.
 
-Available in Edit/Sort Lines.
+To sort lines that are indented or surrounded by blank lines, you can simply
+place the cursor somewhere in the middle without selecting anything and run the
+sort. In other words, if there is nothing selected, this plugin will select
+lines above and below the cursor location, excluding blank lines or any lines
+with less indentation.
 """
 
+from __future__ import annotations
 import tkinter
-
 from porcupine import menubar, tabs, textutils
+
+
+def can_extend_default_selection(text: tkinter.Text, required_prefix: str, new_lineno: int) -> bool:
+    line = text.get(f"{new_lineno}.0", f"{new_lineno + 1}.0")
+    return bool(line.strip()) and line.startswith(required_prefix)
+
+
+def find_chunk_around_cursor(text: tkinter.Text) -> tuple[int, int]:
+    cursor_lineno = int(text.index("insert").split(".")[0])
+    cursor_line = text.get("insert linestart", "insert lineend")
+    cursor_line_indentation = cursor_line[:(len(cursor_line) - len(cursor_line.lstrip()))]
+
+    start = cursor_lineno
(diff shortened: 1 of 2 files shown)


tests/test_sort_plugin.py
```diff
@@ -52,6 +52,41 @@ def test_finding_blank_line_separated_block(filetab):
...
+    catris
+    banana
+
+blah blah
+""",
+    )
+
+    filetab.textwidget.mark_set("insert", "3.3")
+    filetab.event_generate("<<Menubar:Edit/Sort Lines>>")
+    assert filetab.textwidget.index("insert") == "3.3"  # not moved
```
Akuli, line 71: There's no code to ensure a specific cursor place, so in the future, someone might be annoyed by whatever it's doing and want to change the behavior

```suggestion
```

tests/test_sort_plugin.py
```diff
@@ -52,6 +52,41 @@ def test_finding_blank_line_separated_block(filetab):
...
+    filetab.textwidget.insert(
+        "end",
+        """\
+[foo]
+    dingdingding
+    akuli
+    catris
+    banana
+
+blah blah
```
Akuli, line 65:
```suggestion
[bar]
    blah blah
```

tests/test_sort_plugin.py
```diff
@@ -52,6 +52,41 @@ def test_finding_blank_line_separated_block(filetab):
...
+    assert (
+        filetab.textwidget.get("1.0", "end - 1 char")
+        == """\
+[foo]
+    akuli
+    banana
+    catris
+    dingdingding
+
+blah blah
```
Akuli, line 81:
```suggestion
[bar]
    blah blah
```

porcupine/plugins/sort.py
```diff
@@ -1,26 +1,49 @@
...
 
+def can_extend_default_selection(text: tkinter.Text, required_prefix: str, new_lineno: int) -> bool:
+    line = text.get(f"{new_lineno}.0", f"{new_lineno + 1}.0")
+    return bool(line.strip()) and line.startswith(required_prefix)
+
+
+def find_chunk_around_cursor(text: tkinter.Text) -> tuple[int, int]:
+    cursor_lineno = int(text.index("insert").split(".")[0])
+    cursor_line = text.get("insert linestart", "insert lineend")
+    cursor_line_indentation = cursor_line[: (len(cursor_line) - len(cursor_line.lstrip()))]
```
Akuli, line 26:
```suggestion
    cursor_line_indentation = re.match(r"\s*", cursor_line).group(0)
```

porcupine/plugins/sort.py
```diff
@@ -1,26 +1,47 @@
...
+from __future__ import annotations
+
+import re
 import tkinter
 
 from porcupine import menubar, tabs, textutils
 
 
+def can_extend_default_selection(text: tkinter.Text, required_prefix: str, new_lineno: int) -> bool:
+    line = text.get(f"{new_lineno}.0", f"{new_lineno + 1}.0")
```
Akuli, line 20:
```suggestion
    # This returns False when line number is beyond the start or end of the text widget
    line = text.get(f"{new_lineno}.0", f"{new_lineno + 1}.0")
```

tests/test_sort_plugin.py
```diff
@@ -52,6 +52,63 @@ def test_finding_blank_line_separated_block(filetab):
...
+    filetab.event_generate("<<Menubar:Edit/Sort Lines>>")
+    assert (
+        filetab.textwidget.get("1.0", "end - 1 char")
+        == """\
+my_list = [
+\t"asdf",
+\t"bruh",
+]"""
+    )
+    assert filetab.textwidget.get("sel.first", "sel.last") == '\t"asdf",\t"bruh",\n'
```
Akuli, line 109:
```suggestion
    assert filetab.textwidget.get("sel.first", "sel.last") == '\t"asdf",\n\t"bruh",\n'
```

PR #1379, opened 2023-07-12
Fix cyclic anchors bug

### LINKED ISSUE #1353: Anchor plugin bug jumping from top to bottom & vice versa
**To Reproduce:**

1. Go to Porcupine settings and enable "Jumping to previous/next anchor cycles to end/start of file".

2. Open a file, e.g. `foo.py`.

3. Make sure the file has two rows.
 
4. Set an anchor point on row 1 & one anchor point on row 2.

5. Set cursor on row 1. Remove the anchor point on row 1.

6. Use keyboard shortcut to jump to **next** anchor (down). **This works**.

7. Recreate anchor point on row 1. Make sure cursor is on row 1, then remove anchor point on row 1.

8. Use keyboard shortcut to jump to **previous** anchor (up). **The cursor does not move**.

Akuli:
Fixes #1353

The problem was that the plugin needed two anchors to be present for cycling back to the start, so it refused to do anything when there was only one. I changed it so that one anchor is enough. We obviously can't do anything if there are no anchors at all.

I also added tests.

### DIFF
--- a/tests/test_anchors_plugin.py
+++ b/tests/test_anchors_plugin.py
@@ -7,9 +7,9 @@ from porcupine.settings import global_settings
 
 
 def move_cursor(filetab, location):
+    filetab.textwidget.mark_set("insert", location)
     if sys.platform == "win32":
         filetab.update()  # no idea why windows need this
-    filetab.textwidget.mark_set("insert", location)
 
 
 def jump_5_times(filetab, how):
--- a/tests/test_anchors_plugin.py
+++ b/tests/test_anchors_plugin.py
@@ -1,4 +1,6 @@
-import pytest,sys
+import sys
+
+import pytest
 
 from porcupine.menubar import get_menu
 from porcupine.settings import global_settings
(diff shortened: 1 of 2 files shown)


Akuli:
@ThePhilgrim I think this is ready for your review :)

tests/test_anchors_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+# See #1353
+def test_single_anchor_bug(filetab, cyclic_setting_enabled):
+    filetab.textwidget.insert("end", "first row\nsecond row")
+
+    # Line 1 does not have anchor, line 2 has it.
+    filetab.textwidget.mark_set("insert", "2.0")
+    get_menu("Edit/Anchors").invoke("Add or remove on this line")
+
+    # Because cycling is enabled, jumping to previous anchor from line 1
+    # will jumps to the anchor below.
```
ThePhilgrim, line 74:
```suggestion
    # will jump to the anchor below.
```

ThePhilgrim (approved):
I think this looks great. I went through commit by commit to try to understand your attempted tries to fix the windows tests. The fix in `anchors.py` is really simple. Happy it was an easy fix!

I think this is definitely ready to merge.

PR #1386, opened 2023-07-17
Upgrade typing syntax

rdbende:
Used [`pyupgrade`](https://github.com/asottile/pyupgrade) to upgrade typing syntax. Might still have unused imports?

### DIFF
--- a/porcupine/_state.py
+++ b/porcupine/_state.py
@@ -39,7 +39,7 @@ _global_state: _State | None = None
 
 
 def _log_tkinter_error(
-    exc: Type[BaseException], val: BaseException, tb: types.TracebackType | None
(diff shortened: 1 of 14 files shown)


Moosems:
Using `ruff .` will quickly tell you if there’re any unused imports 😄.

rdbende:
The CI did that for me.

Akuli:
Is it possible to make the CI do this automatically for us? Ideally with ruff, but pyupgrade in CI is also fine if ruff can't do all the things as pyupgrade does. Of course, if there's no way to say "please keep the old syntax on this line", that is a no :)

rdbende:
`--keep-runtime-typing` seems to fix the issue with dataclasses. Will add it to the autofix

Akuli:
Replaced by #1387 and #1388.

PR #1388, opened 2023-07-17
Apply new autofixes

Akuli:
PR with empty commit, will hopefully be automagically turned into something more.

### DIFF
--- a/porcupine/pluginmanager.py
+++ b/porcupine/pluginmanager.py
@@ -32,7 +32,7 @@ def get_docstring(module_name: str) -> str:
     spec = importlib.util.find_spec(module_name)
     if spec is not None and spec.origin is not None and spec.has_location:
         try:
-            with open(spec.origin, "r") as file:
+            with open(spec.origin) as file:
                 ast_module = ast.parse(file.read())
         except (OSError, SyntaxError):
             pass
--- a/porcupine/utils.py
+++ b/porcupine/utils.py
@@ -32,14 +32,9 @@ import tkinter
 import traceback
 from pathlib import Path
 from tkinter import ttk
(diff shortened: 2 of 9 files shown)


rdbende:
`--exit-zero-even-if-changed`

Moosems:
Nice name Akuli XD.

Akuli:
> `--exit-zero-even-if-changed`

I already added a hacky `|| true` but this would be better.

porcupine/pluginmanager.py
```diff
@@ -32,7 +32,7 @@ def get_docstring(module_name: str) -> str:
     spec = importlib.util.find_spec(module_name)
     if spec is not None and spec.origin is not None and spec.has_location:
         try:
-            with open(spec.origin, "r") as file:
+            with open(spec.origin) as file:
```
Akuli, line 35: Not sure if I like this change, but it's fine.

rdbende:
Otherwise it works 😅

porcupine/utils.py
```diff
@@ -249,7 +244,7 @@ def invert_color(color: str, *, black_or_white: bool = False) -> str:
 
     # tkinter uses 16-bit colors, convert them to 8-bit
     r, g, b = (value >> 8 for value in widget.winfo_rgb(color))
-    return "#%02x%02x%02x" % (0xFF - r, 0xFF - g, 0xFF - b)
+    return "#{:02x}{:02x}{:02x}".format(0xFF - r, 0xFF - g, 0xFF - b)
```
Akuli, line 247: ewww, but whatever

Moosems: An f-string would probably be better :(.

Issue #1383, opened 2023-07-12
"Maintenance status" comment


Akuli:
Idea: In the beginning of every `.py` file, let's add a comment that explains how developers feel about that file. This is important to know when you start editing a file heavily, but hard to figure out.

Some examples:
- `plugins/find.py`:
    ```
    # Maintenance Status (2023): Behaves exactly as intended, carefully tested,
    # large changes are very unlikely to pass the review.
    ```
- `plugins/langserver.py`:
    ```
    # Maintenance Status (2023): Contains various hacks. I am happy that it works
    # and also happy that I don't have to touch this code very often.
    ```
- `plugins/blockcursor.py`:
    ```
    # Maintenance Status (2023): Not changed much after it was added initially. Does
    # not do anything by default (checkbox in settings). Unclear if anyone needs this.
    ```

benjamin-kirkbride:
Does this belong in a module level docstring? I could see an argument either way.

Akuli:
I don't have strong opinions for docstring vs comment either. Maybe it should depend on whether we want users to see the Maintenance Status of each plugin in the plugin manager, since the plugin manager shows the docstrings of all plugins.

benjamin-kirkbride:
> since the plugin manager shows the docstrings of all plugins

TIL.

My understanding is that docstrings (module level or otherwise) intended audience is users of the API, hence the `help` integration and such.

It seems to me that we have 3 distinct groups that may have need of information (with quite a bit of overlap depending on what hats a given person is wearing):
- users of Porcupine need information on how plugins work, interact with each other, etc. This role is currently covered by the Porcupine usage docs and the module level docstrings for plugins
- plugin developers (someone wanting to make a plugin, but not wanting to make changes to other plugins or Porcupine core), this is mostly served by the "Introduction to Writing Plugins" docs, and reading through the code of plugins that seem to do something similar to what you want to do (or at least that's how I did it :sweat_smile:)
- maintainers who are looking to do something that will modify multiple plugins, or the core, for the purposes of upstreaming/contributing to the project

My understanding is the last group you mentioned above is who you are looking to serve with this proposal, and perhaps new plugin devs as well as you could say something like "this would be a bad example to use if you are looking to make a plugin, look at this other plugin instead" type notes.

If my understanding is correct, then I don't think that having it in a docstring makes sense, in the way that docstrings are generally used, or in the way that we are using them.

Akuli:
[...]

PR #1399, opened 2023-08-01
Mark `menubar.add_filetab_command` as deprecated

benjamin-kirkbride:
This PR:
- Marks `menubar.add_filetab_command` as deprecated

---
Original:

This PR:
- logs warnings
- Adds a Decorator to mark functions as deprecated
- Marks `menubar.add_filetab_command` as deprecated

Decorator design credit: https://shareg.pt/I2pB5Ud

Accompanying deprecation issue: https://github.com/Akuli/porcupine/issues/1398

### DIFF
--- a/porcupine/__init__.py
+++ b/porcupine/__init__.py
@@ -7,7 +7,6 @@ plugin API documentation:
     https://akuli.github.io/porcupine/
 """
 
-import logging
 import os
 import sys
 
@@ -18,8 +17,6 @@ __version__ = "%d.%02d.%02d" % version_info
 __author__ = "Akuli"
 __copyright__ = "Copyright (c) 2017-2023 Akuli"
 __license__ = "MIT"
-
-logging.captureWarnings(True)
 
 if sys.platform in {"win32", "darwin"}:
     # these platforms like path names like "Program Files" or "Application Support"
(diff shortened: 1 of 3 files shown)


Akuli:
It doesn't really make sense to have deprecated functions in Porcupine. If you don't want a function, just remove it. We make no guarantees of API stability.

For this specifically, it would be better to change existing `add_filetab_command()` calls to use actions instead.

benjamin-kirkbride:
I see your perspective. Two counters:

If we want to (eventually) have people be able to make external plugins for porcupine, having some level of API stability would be great, and by extension having a deprecation policy/system would be needed.

The argument of "just fix it" vs make a note that something needs done could be said about things like https://github.com/Akuli/porcupine/issues/1383, or any TODO or FIXME type comments (of which there are many) ;)

I don't have the time or inclination at the present moment to remove `menubar.add_filetab_command`, but in the interest of not leaving a foot gun laying around, seems appropriate to mark it as deprecated.

If I had instead made a PR to add a `FIXME` comment would you have pushed back on that?

Akuli:
You can add a FIXME comment or simply leave the deprecation issue open, whichever you prefer :)

benjamin-kirkbride:
@Akuli done

rdbende (approved):

Issue #1401, opened 2023-08-05
Menubar State Not Resetting Between Tests


benjamin-kirkbride:
Not sure if the issue is with `filetab` fixture, or the `menubar` state, or what.

Reproduction example PR to follow.

Akuli:
Discussion on linked PR (#1402) --> behaves as intended but needs to be documented

Issue #1329, opened 2023-07-04
Parallel Testing with xdist


benjamin-kirkbride:
I tried to use `pytest-xdist` to do parallel testing, but it didn't work. It seems like some state is shared between the running instances somehow. I'm not really sure what though.

Thoughts?

Akuli:
I have tried it too, but I have forgotten if I ever figured out why it didn't work.

Akuli:
I think it has to do with focusing. Some tests assume that you have the Porcupine window focused. But with xdist there are two Porcupine windows, and they can't both simultaneously have the keyboard focus.

benjamin-kirkbride:
How does `xvfb-run` handle this? Is it possible that it's less about "having" focus than "having then losing" focus?

Akuli:
Here is more evidence that it has to do with keyboard focus:

![s](https://github.com/Akuli/porcupine/assets/18505570/5395f669-2d97-44c1-8c25-2e4ad6bb1ebe)

The sequence `.F.F.F.F.FF.` indicates to me that one of the two windows is consistently passing the tests and the other is failing. Also these tests use `event_generate()` which is a bit fragile and I think depends on the focus.

>  xvfb-run

I believe it emulates giving focus to the most recently opened window. That's probably why I get very consistently half of these `event_generate()` dependent tests failing also under xvfb.

Akuli:
I just realized that xvfb doesn't run a window manager, so I think there's not really "multiple windows" to speak of. I'm not really sure what happens instead.

It is apparently possible to run a window manager inside xvfb: https://github.com/TestFX/TestFX/issues/158#issuecomment-62421691

Akuli:
Lol. I think I got a window manager to work inside xvfb.

Without window manager, `.focus_get()` always returns None, maybe it means that focusing isn't really a thing or something. With `xfwm4` running in the background it returns the root window itself because it apparently got focused. (No specific reason to use `xfwm4` of all window managers, it's just what I happen to have)

![t](https://github.com/Akuli/porcupine/assets/18505570/c5c0fefa-3795-484d-9262-952416d8afdf)

Akuli:
I just realized I can take screenshots inside xvfb-run + bash sessions by running `scrot`, a command-line tool that saves a screenshot as image file with no GUI interaction.

The window manager is definitely doing something. Without window manager, `python3 -c 'import tkinter; tkinter.Tk().mainloop()' &` looks like this:

[...]

PR #1402, opened 2023-08-05
Menubar State Not Resetting Between Tests

benjamin-kirkbride:
Fixes https://github.com/Akuli/porcupine/issues/1401

### DIFF
--- a/tests/test_menubar.py
+++ b/tests/test_menubar.py
@@ -134,3 +134,45 @@ def test_add_filetab_action(filetab, tmp_path):
 
     # check unavailable (because Markdown != Python)
     assert menu.entrycget(index=menu_item, option="state") == "disabled"
+
+
+def test_menubar_state_resetting_1(filetab):
+    """we don't want trailing slashes to effect the path at all
+
+    you can accidentally make "empty" sub-menus in tkinter
+    which are confusing to reason about
+    """
+
+    path = "state_resetting_menu/state_resetting_submenu"
+
+    # check that no item exists at path
+    menu_item = menubar._find_item(
+        menubar.get_menu(menubar._split_parent(path)[0]), menubar._split_parent(path)[1]
+    )
+    assert menu_item is None
+
+    menubar.get_menu(path).add_command(label="foo", command=lambda: None)
+    menu = menubar.get_menu(path)
+    menu_item = menubar._find_item(menu, "foo")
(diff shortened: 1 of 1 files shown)


benjamin-kirkbride:
Possibly related to https://github.com/Akuli/porcupine/issues/1329

Akuli:
My first instinct was to say that your tests shouldn't add things to menubar. Usually you add to menubar in plugins and then check the state of the menubar in tests.

But looking at what your test does, I see why you want this. Some ideas:
- Create a fixture that adds a whole new menu for a test, say `get_menu("Testing Temp Menu")`, and then deletes that entire menu once the test completes?
- Test `menubar._split_parent()` and similar functions in their various corner cases, and trust that the rest of the menubar plugin calls them correctly? This gives us confidence that if it works in "normal" cases, it also works in weird corner cases.

benjamin-kirkbride:
I guess I assumed this was unexpected and undesirable behavior.

In previous discussions, particularly revolving around #1329, was that we were not expecting any state to be persisted between tests.

IMO, things like this need to be eliminated wherever they are found if we want to be able to consistently support parallelization of tests (which I do desperately want).

Akuli:
With the way the tests work right now, Porcupine is started once, and then all tests run consecutively in the Porcupine instance. This means that there will always be various ways to influence one test from another if you try.

"Restarting" Porcupine between the tests isn't really an option, if it means that `setup()` functions in all plugins would be called again. I imagine that would make testing a lot slower. It's not very easy to try it now, because there's currently no good way to restart Porcupine without launching a whole new Python process.

benjamin-kirkbride:
Alright, seems like there isn't anything to be done here other than maybe make a fixture and possibly document this as a footgun in the dev docs

benjamin-kirkbride:
What exactly are we envisioning here for documentation? A docstring at the top of the menubar test file? A section in CONTRIBUTING.md?

Akuli:
I'm thinking a new markdown file all about docs. It would explain, among other things:
- what resets between tests and why
- what doesn't reset between tests and using fixtures to work around this

Akuli:
There is now a markdown file about tests that explains why menus behave like they do.

IMO this problem isn't worth fixing, because relatively only a few tests add stuff to menus, and in general, tests shouldn't be confused if there is extra stuff in the menus.

Issue #1224, opened 2022-11-27
Tooltips go off-screen


rdbende:
![image](https://user-images.githubusercontent.com/77941087/204140362-e166f02b-9c9d-42a9-9077-9586cc052a38.png)

Moosems:
You can borrow the placement code from DIP's tooltip, I developed it with Akuli so I hope it lives up to his standards.

Akuli:
yes, I like how simple we eventually got DIP's tooltip logic :)

Though for these tooltips it could be even simpler. We could e.g. always show the tooltip above and to the left, there's always room.

Moosems:
Thank you for your help with that by the way, it was tremendously useful.

lawson89:
I can grab this one - I'll submit a PR by end of day today for review

PR #1412, opened 2023-10-13
Feature/tooltips offscreen

lawson89:
Tooltips go off-screen #1224

show up and to the left by calculating width of tooltip window

### DIFF
--- a/porcupine/utils.py
+++ b/porcupine/utils.py
@@ -175,7 +175,7 @@ class _TooltipManager:
             # the label will have light text on a light background or
             # dark text on a dark background on some systems.
             tkinter.Label(tipwindow, text=self.text, border=3, fg="black", bg="white").pack()
-            tipwindow.update()
+            tipwindow.update_idletasks()
             tipwindow.geometry(f"+{self.mousex - tipwindow.winfo_width()}+{self.mousey - 30}")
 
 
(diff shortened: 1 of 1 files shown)


porcupine/utils.py
```diff
@@ -176,6 +175,8 @@ def show(self) -> None:
             # the label will have light text on a light background or
             # dark text on a dark background on some systems.
             tkinter.Label(tipwindow, text=self.text, border=3, fg="black", bg="white").pack()
+            tipwindow.update()
```
rdbende, line 178:
```suggestion
            tipwindow.update_idletasks()
```

I'm being a little nit-picky here, but usually it's better to use `update_idletasks` instead of `update`, unless you have a good reason not to.

rdbende (changes requested):
Thanks, I like your simple solution! It works well, but I'll wait for @Akuli to have a look at it, I don't dare to merge it just like that 🙃

porcupine/utils.py
lawson89: Thanks, just updated!

rdbende:
btw, the Windows build failure is not related.

rdbende (approved):

Akuli:
This works, but the tooltip looks weird to me when it appears around the middle of the screen.

2023-10-13-225630_1680x1050_scrot

Maybe center it horizontally (`self.mousex - tipwindow.winfo_width()/2`) if it fits, and otherwise do like in this PR now?

Akuli (approved):
Thanks!

PR #1414, opened 2023-10-13
Fix nsis download link (again)

### LINKED ISSUE #1413: Windows build is failing
See e.g. https://github.com/Akuli/porcupine/actions/runs/6511940487/job/17688576360

We need to find a download for nsis (installer creating tool). It doesn't matter if it's on a random untrusted site, because we check the file hash.

Akuli:
Fixes #1413


Akuli:
Windows build passes, though there seems to be other unrelated issues.

PR #1421, opened 2023-11-06
Do not use "font metrics ... -fixed" in monospace detection

### LINKED ISSUE #1368: Debian 12: default font has warning triangle
In a Debian 12 VM, with whatever the default desktop is, if I open the settings, the font family chooser has a warning triangle next to it before I even touch it. I might take a screenshot later.

Once I set the font from "Noto Sans Mono" (default) to something else, the triangle goes away. Editor works even with the triangle though.

Akuli:
Fixes #1368 

Does not noticeably slow down startup on my system, even when the cache file is generated.

### DIFF
--- a/porcupine/settings.py
+++ b/porcupine/settings.py
@@ -895,10 +895,8 @@ def _is_monospace(font_family: str) -> bool:
     if "emoji" in font_family.lower():
         return False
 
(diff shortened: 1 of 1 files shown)


Akuli (commented):
LGTM

Issue #1427, opened 2023-11-30
"Search selected text on Google" can be done for multiline text


Akuli:
1. Open a file with more than 1 line of text
2. Select more than 1 line of text
3. Right-click

Expected result: "Search selected text on Google" is grayed out. It is not reasonable to google multiple lines of text.

Actual result: You can click "Search selected text on Google" and it does nothing.

ChuckM2196:
Thanks for your clear documentation on getting porcupine setup. Mainly done programming on my own so I'm going to try and tackle this issue. I appreciate your willingness to help if I get stuck. I got porcupine up and working.

ChuckM2196:
Going to try and figure this out. Not familiar with Tkinter, but I've used PyQt in the past. I assue you want all the text searched with one google search?

PR #1428, opened 2023-12-01
Edited the google search plugin to include multi line searches

ChuckM2196:
Just starting to learn to contribute to other projects. So I have no idea if I'm doing this correctly.
https://github.com/Akuli/porcupine/issues/1427 - Fixed the issue of now allowing google searches with multiple lines selected
Edited line 19 in plugins/google_search.py to include an additional conditional with multiple lines

### DIFF
--- a/porcupine/plugins/google_search.py
+++ b/porcupine/plugins/google_search.py
@@ -16,7 +16,7 @@ def google_search(tab: tabs.FileTab) -> None:
         return
 
     # Check multi line or text with only spaces
-    if selected_text.strip() and "\n" not in selected_text:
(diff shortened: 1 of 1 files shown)


ChuckM2196:
So we could still have the feature if someone wanted it, but have it optional and set it to disabled by default? I'll look into `rightclick_menu.py` to see what would need to be changed.

Akuli:
I don't think anyone wants to google multiple lines at once. We can change it later if someone actually makes an issue and asks for it.

Moosems:
You can Google with newlines?

rdbende:
Google will replace it with a space, but yes, this is a completely valid search query: `https://www.google.com/search?q=foo%0Abar`

PR #1435, opened 2024-02-07
Get rid of requirements.txt

### LINKED ISSUE #1381: Merging requirements.txt and requirements-dev.txt
Dependencies are specified in 3 places:
- `requirements.txt` (used when setting up dev environment)
- `requirements-dev.txt` (used when setting up dev environment)
- `pyproject.toml` (used by pip)

There is also a test that fails if `requirements.txt` and `requirements-dev.txt` don't match.

I think we should merge `requirements.txt` and `requirements-dev.txt` into a single file that installs everything you need for developing. The test would be changed to ensures that everything in `pyproject.toml` is in the new combined file, but not require the same in reverse, because dev-only dependencies would go in `requirements.txt` instead.

There is also a standard way to specify dev dependencies in `pyproject.toml`, so that you can do `pip install -e ".[dev]"` for instance. We could also use that.

Akuli:
Now `requirements-dev.txt` contains all dependencies. Its purpose is so that you can get started with developing Porcupine easily.

Fixes #1381


Akuli:
Remaining CI failures are unrelated.

Moosems:
Why not turn `requirements-dev.txt` into `requirements.txt`?

Akuli:
I like the `-dev` in the name, because the file is used only for developing Porcupine. For a user, the dependencies come from `pyproject.toml`.

PR #1440, opened 2024-02-07
Fix some glitches in plugins/filetypes.py

taahol:
* Don’t upset dacite when the filetype is None.
* Sync a new file’s type correctly to the menu.


porcupine/plugins/filetypes.py
```diff
@@ -236,7 +236,7 @@ def _sync_filetypes_menu(event: object = None) -> None:
     filetype_name: str = ""
```
Akuli, line 236: This variable can now be None:
```suggestion
    filetype_name: str | None = ""
```

Unfortunately it doesn't seem to be possible to return the correct type from `.get()`, so that mypy would point this out.

Akuli (approved):

PR #1447, opened 2024-03-10
 right-click menu placing

### LINKED ISSUE #1429: right-click menu placing
2023-12-06-134224_1920x1080_scrot

The right-click menu appears exactly where the mouse is. This makes it easy to accidentally click the first menu item. It would be better to place the menu e.g. 10px below the mouse.

Porcupine has several right-click menus, and they should be all fixed at once. They all use the `tk_popup()` menu, so to find all places that need to be changed, you can search for `tk_popup` in the project (with e.g. `git grep tk_popup`).

ethical-haquer:
Added 10 pixels offset to right-click menus
Fixes #1429

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -431,7 +431,7 @@ class DirectoryTree(ttk.Treeview):
         self._populate_contextmenu()
         if self.contextmenu.index("end") is not None:
             # Menu is not empty
-            self.contextmenu.tk_popup(menu_x, menu_y)
+            self.contextmenu.tk_popup(menu_x+15, menu_y)
         return "break"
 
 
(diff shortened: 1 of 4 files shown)


Akuli (commented):
Thanks! A couple things:
- You missed one of the four menus. Run `git grep tk_popup` to find all places where menus are shown.
- These also need y offset.

Akuli:
The y offset is needed in situations where the menu shows near the right edge of the screen. For example, when right-clicking the text widget, this can happen and the menu should be placed below the cursor:

2024-03-10-193051_1680x1050_scrot

ethical-haquer:
> Thanks! A couple things:
> 
>     * You missed one of the four menus. Run `git grep tk_popup` to find all places where menus are shown.
> 
>     * These also need y offset.

Will do!

Akuli:
IMO 10 pixels is better than 1 pixel. You got that part right initially. Sorry if I wasn't clear :)

Akuli (approved):
Thanks! I tried it on my computer and the new behavior is awesome :)

ethical-haquer:
> Thanks! I tried it on my computer and the new behavior is awesome :)

Your welcome! I figured I had better start on something small. 🤣
Thanks for guiding me through the process and answering my numerous questions!

PR #1452, opened 2024-03-16
Update check on startup

### LINKED ISSUE #1141: Update check on startup
It would be nice if Porcupine notified its users **non-disruptively** when a new version is out. For example, it could show a message in the status bar when it starts, or when the first tab is opened if no tabs open on startup (because of the user's settings or because they had no tabs opened when closing Porcupine).

Querying the latest version from GitHub's API is quite straight-forward:

```
>>> import requests
>>> requests.get('https://api.github.com/repos/Akuli/porcupine/releases/latest', headers={'Accept': 'application/vnd.github+json'}).json()['tag_name']
'v2022.07.31'
```

Akuli:
Fixes #1141 

If there is a new version, a message like this appears in the status bar of every tab:

screenshot-1710557960

Once you save the file for the first time, the message goes away. This seems weird, but it's just like the "File was reloaded with unsaved changes." message that shows up in red. In my experience, this gives me enough time to read the message but doesn't leave it sticking around forever (I save frequently).


porcupine/plugins/update_check.py
```diff
@@ -0,0 +1,137 @@
...
+        return f"about {months // 12} years and 1 month ago"
+    return f"about {months // 12} years and {months % 12} months ago"
+
+
+def get_date(version: str) -> datetime.date:
+    year, month, day = map(int, version.lstrip("v").split("."))
+    return datetime.date(year, month, day)
+
+
+def fetch_release_creator(version: str) -> str | None:
```
rdbende, line 52: Personally I don't see why it would be useful to show, who created a release. It just adds a bunch of complexity to the code.

porcupine/plugins/update_check.py
```diff
@@ -0,0 +1,137 @@
...
+    if days == 1:
+        return "yesterday"
+    if days < 1.5 * days_in_month:
+        return f"{days} days ago"
+
+    months = round(days / days_in_month)
+    if months < 12:
+        return f"about {months} months ago"
+    if months == 12:
+        return "about 1 year ago"
```
rdbende, line 35: If we make a distinction here anyway, wouldn’t it be better, to say "about a year ago"?

porcupine/plugins/update_check.py
Moosems: Knowing if it's an automated release vs a an intentional one is nice

Akuli: Porcupine doesn't do automated releases, and probably never will.

rdbende (approved):
Looks good to me!

porcupine/plugins/update_check.py
Moosems: The GitHub action releases?

Akuli: They are triggered by me or rdbende, not fully automated.

PR #1454, opened 2024-03-20
Add dev-doc folder and document some stuff there

### LINKED ISSUE #1426: Documenting the project structure and philosophy
The ti;dr is:
- core of Porcupine is small, and most of the functionality is in plugins
- plugins are loaded dynamically
- each plugin does a specific thing, and can be enabled/disabled individually by the user
- plugins can depend on each other when needed, and they pass messages to each other through virtual events.

This way, the user is not forced to use any particular thing they don't like. For example, I don't like the minimap plugin as it is now.

Akuli:
Starts the work on #1308 and other docs issues. Fixes #1426.


Akuli:
@rdbende Any chance you could have a quick look in the next few days? You don't need to review super carefully. I mostly want to know your thoughts on the concept of markdown docs and the file structure.

rdbende (commented):
Strange that GH doesn't send notifications about requested reviews.

Anyway, as I said previously, IMO it's a good idea to put these "dev-guide" type of docs into Markdown files, and the current content looks good to me.

I would do two more things. First, put the images in a separate directory so they don't clutter the `dev-doc` folder. Second, I would create a "getting-started" file (a README or something similar) so that it is clear where I should start as a new contributor. Currently this could be confusing.

For the user guide, I'd still prefer a website, but it would indeed be a hassle to maintain, and I'm not sure how many people would even read it.

Akuli:
I moved the images to a new `images` folder.

> a "getting-started" file

This already exists, and it is called `CONTRIBUTING.md`. It mentions the dev-doc folder and links to a good starting place inside the folder.

PR #1458, opened 2024-03-29
Document tests

### LINKED ISSUE #1317: Can't `print` in tests
Any idea why stdout (print) isn't being captured by pytest? I'm trying to use it to debug.

### LINKED ISSUE #1382: When to write tests?
From #1371:

>> Porcupine isn't a project where you need to write a test for everything.

> :disappointed:

Let's discuss when and why tests should be written, and when and why they should not be written.

I think we should write tests when:
- The code is hard to get right when writing it initially.
- It is easy to break the code in unintended ways when you try to modify it.
- The test is simple, easy to understand, and easy to get right when writing it initially.
- There either isn't other higher-level code that invokes the code we're testing, or there is but the higher-level code fails in mysterious ways when the lower-level code is broken.
- If the feature was broken, it would take a long time for Porcupine developers to notice it.

A good example of this is the find plugin. It has lots of corner cases that must behave in exactly the right way for it to not annoy me. Some corner cases I would notice immediately when trying to do real work with Porcupine, but others I would not. All corner cases are easy to test though, since you mostly need to insert stuff to entries, press buttons, and check what got tagged in the text widget.

I think we should NOT write tests when:
- The code is trivially simple.
- There are no surprising corner cases that you would need to be particularly careful about, or you can explain them in a short comment. (Nobody reads long comments.)
- Testing the code would be difficult.
- If the feature was broken, Porcupine developers would notice it when trying to use Porcupine for something and immediately create an issue.

A good example of this is UI details, such as padding, font sizes, window sizes, window titles and the like. To me, pull requests like #794 check all the boxes for something that doesn't need tests.

I also think that the number of lines of code shouldn't be used to determine whether tests are a good idea or not. Sometimes you need 50 lines of tests for a one-line regex, sometimes you need 1000 lines of code to implement autocompletions with a langserver even though a 20 lines long test verifies that completions are working in a couple different situations.

Thoughts?

### LINKED ISSUE #1395: CONTRIBUTING.md should recommend python's breakpoint() instead of "import pdb; pdb.set_trace()"
Does the same thing but more convenient. Did not exist when I learned about pdb, but now everyone has a new enough python for using `breakpoint()`.

### LINKED ISSUE #1401: Menubar State Not Resetting Between Tests

Akuli:
Fixes #1317 
Fixes #1382
Fixes #1395  
Fixes #1401

I removed all coverage stuff, because it doesn't align with Porcupine's way of doing testing. 100% code coverage is just not something to aim for. If someone still needs coverage for something, that will work, but I no longer want to mention it to new contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,21 +65,8 @@ environment in, you can go back to the environment by `cd`'ing to the correct
 place and running `source env/bin/activate` again. You can run `deactivate` to undo
 the `source env/bin/activate`.
 
-Other commands you may find useful:
-- `python3 -m pytest` runs tests. You will see lots of weird stuff happening
-  while testing, and that's expected.
-    A good way to debug a test to see what is actually going on is to add traces.
-    It pauses the test to show you the current state of the program.
-    - Use `import pdb` and `pdb.set_trace()` to set the pause points in the test. You can
-      set as many as you like, and it can conveniently be done on one line: `import pdb; pdb.set_trace()`.
-    - When the test pauses, type `cont` in terminal to continue the test.
-    - If you at any time need to interact with the program during the pause,
-      type `interact` in terminal. Exit interactive mode with `ctrl + D`.
-- To see a report of test coverage, add `--cov=porcupine` to the above pytest
-  command and then run `coverage html`. Open `htmlcov/index.html` in your favorite
-  browser to view it.
-- `cd docs` followed by `python3 -m sphinx . build` creates HTML documentation.
-  Open `docs/build/index.html` in your favorite browser to view it.
+Porcupine also has tests, but many changes to Porcupine don't need tests.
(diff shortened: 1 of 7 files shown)


Akuli:
@rdbende I am pinging you, because you said that you don't get notifications for review requests. Let me know if you don't want to be pinged.

rdbende:
I like it!
Sometimes it feels like an entire tutorial on unit testing, but apparently that's needed, so I think it's good to go.


> @rdbende I am pinging you, because you said that you don't get notifications for review requests. Let me know if you don't want to be pinged.

It's okay. But I indeed didn't get a notification for the review request :)

rdbende (approved):

PR #1459, opened 2024-03-29
Add link checker script for markdown files

Akuli:
I am worried that links in documentation will get outdated over time. This script will check them. The script sends HTTP HEAD requests, but it's probably fine, because the whole purpose of HTTP HEAD is to avoid reading long responses when you don't really care about the contents of a page.

I was going to make this a test, as in `python -m pytest`, but that doesn't work very well for a few reasons:
- Error reporting wouldn't be great. It's hard to show reasonable markdown file line numbers in pytest.
- The test would need to be somehow disabled by default. I don't want to send unsolicited HTTP requests all over the internet from other Porcupine developers' computers. (That said, the pastebin tests already work this way.)

I also tried two existing link checker projects, but they didn't seem to understand that `[...](foo.md)` points at a file named `foo.md` in the same directory.

### DIFF
--- a/scripts/check-markdown-links.py
+++ b/scripts/check-markdown-links.py
@@ -1,5 +1,6 @@
 """Check that links in markdown files point to reasonable places."""
 
+import argparse
 import os
 import re
 import subprocess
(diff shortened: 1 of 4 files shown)


scripts/check-markdown-links.py
```diff
@@ -0,0 +1,105 @@
...
+    output = subprocess.check_output(["git", "ls-files", "*.md"], text=True)
+    return [Path(line) for line in output.splitlines()]
+
+
+def find_links(markdown_file_path):
+    content = markdown_file_path.read_text(encoding="utf-8")
+
+    if markdown_file_path.name == "CHANGELOG.md":
+        # Ignore changelogs of old versions. Editing them doesn't make sense.
+        header_matches = list(re.finditer("^## v", content, flags=re.MULTILINE))
```
Akuli, line 28: This must also match `## UNRELEASED`
```suggestion
        header_matches = list(re.finditer("^## ", content, flags=re.MULTILINE))
```

PR #1462, opened 2024-03-30
Port getting started documentation from Porcupine Wiki and update it

Akuli:
- New screenshots
- Deleted long line marker and line numbers explanation, boring/obvious/unimportant
- Explained some of the most important features, including find/replace
- Applied `<details>` to platform-specific instructions in README and linked getting started doc below them. Without the `<details>`, I would need to link getting-started docs *before* install instructions to make it visible, which would be weird.

Creates the `user-doc` folder discussed in #1308. The `dev-doc` folder already exists.


Akuli:
@rdbende Please review when you feel like it :)

PR #1469, opened 2024-03-30
Add easter eggs to about dialog

Akuli:
The code is intentionally somewhat vague. I don't want to spoil the easter eggs to people who read the diff of this pull request :)

### DIFF
--- a/porcupine/plugins/aboutdialog.py
+++ b/porcupine/plugins/aboutdialog.py
@@ -90,8 +90,8 @@ class EasterEggs:
                 return
             self.timeout = self.pic.after(round(1000 / 60), bounce)
 
-            x_min = -(self.window.winfo_width() - 200)
-            y_min = -(self.window.winfo_height() - 200)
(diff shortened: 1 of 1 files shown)


rdbende:
> The code is intentionally somewhat vague. I don't want to spoil the easter eggs to people who read the diff of this pull request :)

Not vague enough :D I don't have a porcupine repo cloned atm, but I already know what it does :)

Akuli:
I don't think I want this to be more vague, because users need to be able to trust Porcupine, after all. If it looks like [this hello world program I wrote many years ago](https://github.com/Akuli/hello-world/blob/master/hello.py), you can't really tell whether it is a cryptominer or a harmless easter egg.

At least you had fun reading the code :)

rdbende:
> If it looks like [this hello world program I wrote many years ago](https://github.com/Akuli/hello-world/blob/master/hello.py), you can't really tell whether it is a cryptominer or a harmless easter egg.

lol, that's a cryptominer. I don't trust it.

> At least you had fun reading the code :)

Yeah, it was definitely amusing :)

PR #1470, opened 2024-03-30
Delete Porcupine Wiki

Akuli:
All content except the key cheatsheet has been updated and moved to `user-doc`. I'm not sure if I want to keep the key cheatsheet, as it is kinda incomplete anyway. I can always add it back later.

This is an important part of fixing #1308.


rdbende:
This looks good to me. Not sure how useful the key cheatsheet actually was, I expect most people use the accelerators shown in the menu anyway.

rdbende:
Although the tabular approach is pretty cool to explain the thinking behind the key combinations you choose.

PR #1479, opened 2024-03-31
Update flake8-tkinter and improve its config

Akuli:


porcupine/utils.py
```diff
@@ -548,7 +548,7 @@ def callback(shifted: bool, event: tkinter.Event[tkinter.Misc]) -> Any:
         shift_tab = "<Shift-Tab>"
 
     widget.bind("<Tab>", functools.partial(callback, False), **bind_kwargs)  # noqa: TK231
-    widget.bind(shift_tab, functools.partial(callback, True), **bind_kwargs)  # noqa: TK231
+    widget.bind(shift_tab, functools.partial(callback, True), **bind_kwargs)  # noqa: TK131
```
rdbende, line 551: Hmm. Is there a bug, or why is a `# noqa: TK131` needed here?

rdbende, line 551:
```suggestion
    widget.bind(shift_tab, functools.partial(callback, True), **bind_kwargs)  # noqa: TK141
```

PR #1494, opened 2024-04-02
Make sure command output is visible

### LINKED ISSUE #1128: Sometimes command output is invisibly small
You can stretch the output area to make it bigger or smaller. Porcupine even remembers its position when you restart, I think. But this can happen:
1. User wants to hide the output. Instead of pressing F4 (or clicking something in the menu), they just stretch it to be zero height, which is quite a reasonable thing to do.
2. Time goes on. User forgets what they did.
3. User wants to run a new command. The output area comes up with zero height.

When showing the output area for running a command, we should ensure it is never e.g. less than 100px tall.

ethical-haquer:
Fixes #1128.


porcupine/settings.py
```diff
@@ -841,7 +841,18 @@ def settings_to_gui(junk: object = None) -> None:
         if panedwindow["orient"] == "horizontal":
             panedwindow.paneconfig(pane, width=global_settings.get(option_name, int))
         else:
-            panedwindow.paneconfig(pane, height=global_settings.get(option_name, int))
```
Akuli, line 844: This file isn't the right place for this change. Because we're doing this for the run plugin, the fix belongs to the run plugin. The correct place is probably the `run` method in `class Executor` (file `no_terminal.py`), as it already does stuff with font size and textwidget height.

porcupine/settings.py
```diff
@@ -841,7 +841,18 @@ def settings_to_gui(junk: object = None) -> None:
...
+            """Use previous height as long as it's not less than two lines high,
+            otherwise set height to two lines high.
+            """
```
Akuli, line 846: Nit: `#` should be used for comments unless it's in the beginning of a function, class, or file.

porcupine/settings.py
```diff
@@ -841,7 +841,18 @@ def settings_to_gui(junk: object = None) -> None:
...
+            previous_height = global_settings.get(option_name, int)
+            linespace = tkinter.font.Font(font="TkFixedFont").metrics("linespace")
+            linespace += 3
```
Akuli, line 849: Maybewe could use the `get_padding()` function defined in `textutils.py` instead of a hard-coded 3?

Akuli (commented):
Thanks! This needs a few adjustments.

(If you want to prevent your PRs from containing changes of previous PRs, you can make a new branch based on my latest main branch for your PRs. But as I say in CONTRIBUTING.md, I don't really mind it.)

porcupine/settings.py
ethical-haquer: 👍 What about when the comment is too long for one line?
Use two `#`s?

Akuli: Yes, use multiple `#`s.

ethical-haquer:
I have this, but It's not quite correct:
```
        font = tkinter.font.Font(name="TkFixedFont", exists=True)

        # Use previous height as long as it's not less than two lines high,
        # otherwise set height to two lines high.
        padding = textutils.get_padding(self._textwidget)[0]
        print(f"padding is: {padding}")

        linespace = font.metrics('linespace')
        linespace += padding
        linespace *= 2

        # Update needed to get width and height, but causes key bindings to execute
        self._textwidget.update()
        width, height = textutils.textwidget_size(self._textwidget)

        if height < linespace:
            print("height is less than linespace")
            print(height)
            print(linespace)
            # For one thing, this doesn't work
            self._textwidget.configure(height=linespace)
```

Akuli:
Now that I look at the code again, maybe the best place is actually the `run_command()` function in `no_terminal.py`. You will need to set a height in the paneconfigure.

The problem with `textwidget.configure(height=...)` is that the panedwindow ignores the text widget's request for size. This makes sense, because the user must be able to resize the text widget by simply dragging the panedwindow.

porcupine/plugins/run/no_terminal.py
```diff
@@ -477,3 +477,16 @@ def run_command(command: str, cwd: Path) -> None:
     assert runner is not None
     get_vertical_panedwindow().paneconfigure(runner.textwidget, hide=False)
     runner.run_command(cwd, command)
+    # Use previous height as long as it's not less than two lines high,
+    # otherwise set height to two lines high.
```
Akuli, line 481: This adds quite a lot of code into an otherwise simple function. Can you move your code into a new function and call it from `run_command()`?

porcupine/plugins/run/no_terminal.py
```diff
@@ -477,3 +477,16 @@ def run_command(command: str, cwd: Path) -> None:
...
+    current_height = textutils.textwidget_size(runner.textwidget)[1]
+    padding = textutils.get_padding(runner.textwidget)[0]
+    linespace = tkinter.font.Font(font="TkFixedFont").metrics("linespace")
+    linespace += padding
+    # Show two lines of output
+    linespace *= 2
```
Akuli, line 487: This calculation does the right thing, but it's misleading.

In pseudo code, it calculates:

```
height = (linespace + padding) * 2
```

The correct calculation would be:

```
height = 2*linespace + 2*padding
```
where `2` in front of linespace means we make it two lines tall, and `2` in front of padding means we take into account the padding on top and the padding on bottom. Mathematically these are the same, but I would prefer a calculation that looks more like `2 * linespace + 2 * padding`.

porcupine/plugins/run/no_terminal.py
```diff
@@ -477,3 +477,16 @@ def run_command(command: str, cwd: Path) -> None:
     assert runner is not None
     get_vertical_panedwindow().paneconfigure(runner.textwidget, hide=False)
     runner.run_command(cwd, command)
+    # Use previous height as long as it's not less than two lines high,
+    # otherwise set height to two lines high.
+    current_height = textutils.textwidget_size(runner.textwidget)[1]
+    padding = textutils.get_padding(runner.textwidget)[0]
```
Akuli, line 483: We want the y padding:
```suggestion
    padding = textutils.get_padding(runner.textwidget)[1]
```

Akuli (commented):
Thanks for working on this, looks better now :) A few more tweaks are needed...

Akuli:
On my system, it sometimes fails to become large enough:

https://github.com/Akuli/porcupine/assets/18505570/236d0ab8-b33d-403f-830f-0586305bb1b2

I will figure out why.

Akuli:
I tried a few things and can't figure out why that is happening, but I added a funny workaround.

Akuli (approved):
Thanks for working on this! I like it, even though I needed to add an ugly hack, but at least the ugliness is hidden away in a function with good comments :)

ethical-haquer:
> Thanks for working on this! I like it, even though I needed to add an ugly hack, but at least the ugliness is hidden away in a function with good comments :)

No problem!
I was a fun challenge looking through the code.
As for the "hack", sometimes tkinter is like that.
Thanks for guiding me through it!

PR #1506, opened 2024-05-16
Make sure that titlebar is black on all platforms

jhi2:
The titlebar is always black when the icon plugin is enabled

### DIFF
--- a/porcupine/plugins/windowicon.py
+++ b/porcupine/plugins/windowicon.py
@@ -1,14 +1,13 @@
 """Set the title and icon of the Porcupine window."""
-from porcupine import __version__ as porcupine_version
-from porcupine import get_main_window, images
 import Titleman as ttle
 
+from porcupine import __version__ as porcupine_version
+from porcupine import get_main_window
+
+
 def setup() -> None:
-    #Bitmap for the icon? Not with windowmanager because that's how titleman works.
+    # Bitmap for the icon? Not with windowmanager because that's how titleman works.
     window = get_main_window()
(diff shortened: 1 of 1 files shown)


Moosems:
Why would want it black across all systems?

rdbende:
> Make sure that titlebar is black on all platforms

Sorry, it looks like the titleman package doesn't even load outside of Windows, so I don't know how this could work on all platforms.

Other than that, I don't think we want a custom titlebar in Porcupine. They are known to have all kinds of weird bugs on different window managers, including Windows.

jhi2:
In my next release of Titleman, I will make sure it works only on Windows. Sorry about that.  I develop on Windows 11.
Also, on manmy Linux distros and on Mac, it is already black.
________________________________

Moosems:
It's not black on Mac unless you set the theme to dark mode. By default (light mode) it's white.

jhi2:
I will look into how to programacticly set that
________________________________

Akuli:
Thanks for contributing, but I'm not convinced that this is a good idea. What problem does it solve, and why do users need it?

jhi2:
style

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

________________________________

Moosems:
That should be a separate thing then.

Akuli:
I'm still not convinced that this is a good idea. I will not accept code that messes with the title bar beyond setting text and icon.

jhi2:
What do you mean a seprate thing

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

________________________________

PR #1507, opened 2024-06-02
Delete toposort dependency

Akuli:
Given that xz just happened, I want to minimize Porcupine's dependencies, and I think `toposort` is the only dependency that can be reasonably removed.

When I added `toposort` dependency in 2017 (commit c713798b30bad0d3d4812246b07092cb4cfc9c4a), it was because I couldn't figure out how to program it myself. The source code is short, but it looked complicated and I didn't want to copy it. Since then, I have worked with graphs more and the algorithm is now quite obvious to me, so I rewrote it from scratch.


porcupine/pluginloader.py
```diff
@@ -276,35 +274,73 @@ def run_setup_argument_parser_functions(parser: argparse.ArgumentParser) -> None
             _run_setup_argument_parser_function(info, parser)
 
 
+def _handle_circular_dependency(cycle: list[PluginInfo]) -> None:
+    error_message = " -> ".join(info.name for info in cycle)
+    log.exception(f"circular dependency: {error_message}")
+    for info in cycle:
+        info.status = Status.CIRCULAR_DEPENDENCY_ERROR
+        info.error = f"Circular dependency error: {error_message}"
```
Akuli, line 282: I need to test this manually before merging.

Moosems:
Why can't platformdirs be removed?

Moosems:
There's a reason I've decided to cut windows support in all my future projects?

PR #1509, opened 2024-06-02
Improve pluginloader docstrings

Akuli:
The docstrings were quite messy, because they were written to satisfy sphinx, not to be human-readable. Now that sphinx is out, the docstrings can be freely improved.


porcupine/pluginloader.py
```diff
@@ -35,95 +35,71 @@
...
+
+    # Importing the plugin raised an error.
     IMPORT_FAILED = enum.auto()
+
+    # The plugin was imported successfully, but its `setup()` function
+    # raised an exception or logged an error.
+    #
+    # In a plugin named `foo`, any message logged with severity `ERROR`
+    # or `CRITICAL` to the logger named `porcupine.plugins.foo` counts as
+    # logging an error. Therefore you can do this::
```
Akuli, line 63:
```suggestion
    # logging an error. Therefore you can do this:
```

porcupine/pluginloader.py
```diff
@@ -35,95 +35,71 @@
...
+    # raised, the full traceback is shown instead.
     SETUP_FAILED = enum.auto()
+
+    # Plugins with this status were imported, but their `setup_before` and
+    # `setup_after` lists make it impossible to determine the correct order
+    # for calling their `setup()` function.
+    #
+    # For example, if plugin *A* should be set up before *B*, *B* should be
+    # set up before *C* and *C* should be set up before *A*, then *A*, *B*
+    # and *C* will all fail to load with this status.
```
Akuli, line 86:
```suggestion
    # For example, if plugin A should be set up before B, B should be
    # set up before C, and C should be set up before A, then A, B
    # and C will all fail to load with this status.
```

porcupine/pluginloader.py
```diff
@@ -35,95 +35,71 @@
...
+    This dataclass represents a plugin.
+
+    The value of `error` depends on `status`:
+    - `LOADING`: `error` is `None`
+    - `ACTIVE`: `error` is `None`
+    - `DISABLED_BY_SETTINGS`: `error` is `None`
+    - `DISABLED_ON_COMMAND_LINE`: `error` is `None`
+    - `IMPORT_FAILED`: `error` is a Python error message
+    - `SETUP_FAILED`: `error` is a Python error message
+    - `CIRCULAR_DEPENDENCY_ERROR`: `error` is a user-readable one-line message
```
Akuli, line 102: This should be moved just above the `error` field, and replaced with something shorter, like this

```
    The value of `error` is:
    - a user-readable one-line message, if `status` is `CIRCULAR_DEPENDENCY_ERROR`
    - a Python error message, if `status` is `IMPORT_FAILED` or `SETUP_FAILED`
    - `None` otherwise.
```

porcupine/pluginloader.py
```diff
@@ -326,6 +302,7 @@ def _decide_loading_order(
 
 # undocumented on purpose, don't use in plugins
 def run_setup_functions(shuffle: bool) -> None:
+    """Called during Porcupine startup. Do not call from plugins."""
     # the toposort will partially work even if there's a circular
     # dependency, the CircularDependencyError is raised after doing
     # everything possible (see source code)
```
Akuli, line 308:
```suggestion
```

porcupine/pluginloader.py
```diff
@@ -388,10 +365,10 @@ def can_setup_while_running(info: PluginInfo) -> bool:
 
 
 def setup_while_running(info: PluginInfo) -> None:
-    """Run the ``setup_argument_parser()`` and ``setup()`` functions now.
+    """Run the `setup_argument_parser()` and `setup()` functions now.
 
-    Before calling this function, make sure that
-    :func:`can_setup_while_running` returns ``True``.
+    Before calling this function, make sure that the
+    `can_setup_while_running()` function returns ``True``.
```
Akuli, line 371:
```suggestion
    `can_setup_while_running()` function returns `True`.
```

porcupine/pluginloader.py
```diff
@@ -326,9 +294,7 @@ def _decide_loading_order(
 
 # undocumented on purpose, don't use in plugins
```
Akuli, line 295: Now explained in docstring, no need to repeat
```suggestion
```

PR #1511, opened 2024-06-04
Delete platformdirs dependency

Akuli:
Discussed in #1507 

@Moosems Thanks for suggesting this to me :)


Moosems:
How was Windows heck?

Moosems:
> my work computer... unfortunately uses Windows.

It *what*?

PR #1512, opened 2024-06-21
Update directory_tree.py:  Fixed double clicking

celltoolz:
Fixed double click to work properly

### DIFF
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -138,7 +138,7 @@ class DirectoryTree(ttk.Treeview):
 
         little_arrow_clicked = self.identify_element(event.x, event.y) == "Treeitem.indicator"
 
-        if little_arrow_clicked:
+        if little_arrow_clicked or not item.startswith("file"):
             self.item(item, open=(not self.item(item, "open")))
             if self.item(item, "open"):
                 self.open_file_or_dir()
--- a/porcupine/plugins/directory_tree.py
+++ b/porcupine/plugins/directory_tree.py
@@ -139,9 +139,7 @@ class DirectoryTree(ttk.Treeview):
 
         self.set_the_selection_correctly(item)
 
-        little_arrow_clicked = self.identify_element(event.x, event.y) == "Treeitem.indicator"
-
-        if little_arrow_clicked or not item.startswith("file"):
+        if item.startswith(('dir', 'project')):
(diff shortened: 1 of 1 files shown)


Akuli:
Thanks for working on this! Your solution is nicer than calculating milliseconds.

I noticed to differences to how this behaved before. Can you look into whether we can keep the old behavior?

1. Clicking an arrow shouldn't require a double-click, so that you can open folders with a single click. Now I need a double-click to open a folder when clicking the arrow. Here's how it worked before:

https://github.com/Akuli/porcupine/assets/18505570/9240bf15-2025-4d74-8157-03ceba84bcb1

2. Double-clicking a folder should open it, so that if you don't mind clicking twice, you don't have to aim the mouse accurately on top of a tiny arrow. Like this:

https://github.com/Akuli/porcupine/assets/18505570/b326580a-90c7-42d5-9948-cfdec2f6a23b

These probably have something to do with `return "break"`. Let me know if you need help :)

celltoolz:
It's working as you described now, was as simple as adding `or not item.startswith('file')` to line 142.  Now when you single click anywhere on a folder name it'll expand showing the contents

Akuli:
I pushed a couple small changes. You will need to run `git pull` locally. I'm now happy with how it behaves.

porcupine/plugins/directory_tree.py
```diff
@@ -126,27 +137,15 @@ def _on_click(self, event: tkinter.Event[DirectoryTree]) -> str | None:
...
-        # measure max time between double clicks. It's 500ms on my system.
-        double_click = event.time - self._last_click_time < 500 and self._last_click_item == item
-
         self.set_the_selection_correctly(item)
-        if item.startswith("file:"):
-            if double_click:
+
+        little_arrow_clicked = self.identify_element(event.x, event.y) == "Treeitem.indicator"
+
+        if little_arrow_clicked or not item.startswith("file"):
```
Akuli, line 144: I don't think we need `little_arrow_clicked` anymore, because directories open regardless of where you click, and files don't have the arrow.

celltoolz:
Was thinking of maybe adding a symbols table to this this, something that looks like:

Let me know what you think, i've already got a class that i wrote for another project that'd work wonderfully for just this thing. I'd only need to modify it to work with the other languages as it only works for python at the moment.

Akuli (approved):

PR #1514, opened 2024-06-23
Fix simple terminal run command in external window

### LINKED ISSUE #1513: Cannot run a command in an external terminal window with st
st (simple terminal) immediately exits when running a command with: child exited with status 1

To reproduce:

1. Install the latest version of simple terminal
2. Assure it is set as your TERMINAL env variable
3. Run a command in an external terminal window in porcupine

Expected result: runs the command in an external simple terminal window
Actual result: immediately exits with status 1

8manz:
This works on my own build with various patches for v0.9.1 and the default for v0.9.2.

Not sure if this is the best implementation, so please tell me how I can improve.

Fixes  #1513


Akuli (approved):
Looks good! Thanks!

